PR #1, opened 2018-03-02
Add basic .travis.yml

spbnick:


spbnick:
@agustinhenze, do you think this .travis.yml is alright, or should we change/add something?

spbnick:
I see we don't have python2-mock dependency in there. Perhaps we should have it explicitly?

spbnick:
Alright, added the explicit "mock" dependency, even though it's already installed on Travis CI, just in case.

spbnick:
Merged.

agustinhenze:
Sorry, I couldn't answer before. Yes, totally agree about adding mock explicitly. We have CI! Awesome!

Issue #12, opened 2018-03-13
Run pylint on CI


agustinhenze:
The main idea is that the CI pipeline takes care the new code being lint compliant. For now, it can be run only for the `tests` folder but judging for the issue #3 all the code are going to be lint compliant :smile:

PR #13, opened 2018-03-13
Add lint check on ci

agustinhenze:
* Add pylint + pep8 on .travis.yml
* Make tests comply with pep8 requirements

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,8 @@ python:
     - "2.7"
 install:
     - pip install requests junit_xml beaker-client mock
-script: python -m unittest discover tests
+    - pip install pylint pycodestyle
+script:
+    - python -m unittest discover tests
+    - pylint tests
+    - pycodestyle tests
--- a/tests/misc.py
+++ b/tests/misc.py
@@ -1,18 +1,19 @@
 """
 Miscellaneous for tests.
 """
-# Copyright (c) 2018 Red Hat, Inc. All rights reserved. This copyrighted material
-# is made available to anyone wishing to use, modify, copy, or
-# redistribute it subject to the terms and conditions of the GNU General
-# Public License v.2 or later.
+# Copyright (c) 2018 Red Hat, Inc. All rights reserved. This copyrighted
+# material is made available to anyone wishing to use, modify, copy, or
+# redistribute it subject to the terms and conditions of the GNU General Public
+# License v.2 or later.
 #
-# This program is distributed in the hope that it will be useful, but WITHOUT ANY
-# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-# PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
 #
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 import os
 
 
--- a/tests/test_reporter.py
+++ b/tests/test_reporter.py
@@ -1,18 +1,19 @@
 """
 Test cases for reporter module.
 """
-# Copyright (c) 2018 Red Hat, Inc. All rights reserved. This copyrighted material
-# is made available to anyone wishing to use, modify, copy, or
-# redistribute it subject to the terms and conditions of the GNU General
-# Public License v.2 or later.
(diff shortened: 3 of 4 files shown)


agustinhenze:
It closes #13

.pylintrc
```diff
@@ -0,0 +1,3 @@
+[FORMAT]
+# PEP8 compliant
+max-line-length=79
```
spbnick: Ah, so PEP8 is stricter than Google Python Style Guide. Alright. However, we'll need to then fork the guidelines and adjust the column number in there. We should probably put the guidelines into https://github.com/RH-FMK/meta

tests/misc.py
```diff
@@ -1,18 +1,20 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
 #
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
-# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301,
+# USA.
```
spbnick: Could you please rewrap this paragraph as a whole, here and everywhere else, so we don't have an extra line with the hanging "USA"? It should fit.

.travis.yml
```diff
@@ -3,4 +3,7 @@ python:
     - "2.7"
 install:
     - pip install requests junit_xml beaker-client mock
-script: python -m unittest discover tests
+    - pip install pylint
+script:
+    - python -m unittest discover tests
+    - pylint tests
```
spbnick: Could you please put the pylint invocation first, as it's a more basic, lower level test?

spbnick:
Oh, and please add a line saying "Fixes #12" to the message of the commit adding pylint to .travis.yml.

tests/misc.py
agustinhenze: Ok, I have tried again and I get `C: 16, 0: Line too long (80/79) (line-too-long)` from pylint.

.travis.yml
agustinhenze: My reasoning was that I'd prefer to see what is the result of the test suite when a PR is opened and if pylint doesn't pass is not something useful for the review. It has to pass if you want to see the PR accepted but it does not contribute too much in the discussions.

.pylintrc
agustinhenze: Maybe we can mention that we follow the Google Python Style Guide but pylint is on top of it. If you get a pylint error it is something that you have to fix. With the idea in mind to maintain less dev doc :)

spbnick: Yeah, we can have a "contributing" document refer to the style guide and have exceptions on top. That will work as long as exceptions are not numerous. We'll have to spell them out, though, not just point at the CI code.

tests/misc.py
spbnick: I mean something like this:

    # You should have received a copy of the GNU General Public License along with
    # this program; if not, write to the Free Software Foundation, Inc., 51
    # Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.

.travis.yml
spbnick: I would just like to have some logical order, it doesn't matter which. I agree, having it "most important first" or "biggest impact first" makes sense, and will work. Could you perhaps put a comment above, specifying which order the tests should be, so we don't have to discuss this again? You can use `echo` to print it, or just `true` with comment as the argument, which will be ignored.

tests/misc.py
agustinhenze: Ha! You're right, I misunderstood you before. Pushed

spbnick:
I mean put "Fixes #12" line into the commit message *body* (at the end), not *subject*. Sorry if I was unclear. GitHub will then see it and close the issue automatically when I push the commit.

.travis.yml
spbnick: Ah, on the second thought, adding the comment about the order matters little at this moment, please feel free to ignore.

.pylintrc
agustinhenze: Ok, agreed

.travis.yml
agustinhenze: Ok

tests/test_reporter.py
```diff
@@ -34,6 +35,7 @@ def request_get_mocked(filename):
             filename: Asset's filename.
         """
         get_mocked = mock.Mock()
+
```
spbnick: This is a needed change, but it's not about complying with the line length requirements as commit subject says.
Please either make it a separate commit, or change the commit subject to something like "Make tests comply with pep8 requirements". While this is a small matter, it's easier to simply always keeping commit descriptions true, than deciding every time if it's important or not.

agustinhenze: You are so right. I have completely forgot that I have been running pep8 locally... Please take a look to the last push, I have added pep8 and remove the .pylintrc file to keep it simple as possible.

agustinhenze: Well I've changed it by pycodestyle because pep8 is deprecated as you can see [here](https://travis-ci.org/RH-FMK/skt/builds/353475444?utm_source=github_status&utm_medium=notification)

spbnick:
Thanks a lot, Agustin, merged!

PR #18, opened 2018-03-19
[pycodestyle] Fix violations

veruu:
This patch fixes the gigantic number of syntax (mainly pep8) violations,
namely:
- multiple imports on the same line and their order
- whitespace issues (whitespaces present around keyword arguments or
  before ':', number of blank lines, missing space after '#' in comment,
  mixed tabs and spaces...)
- conditionals: comparisons with None, True
- too long lines (>79 chars)
- indentation
- bare excepts (left the one in skt/\_\_init\_\_.py since since fixing it
  requires understanding of the code and what exceptions it may throw)

The patch doesn't fix any 'non-pythonic' code, it just makes the first
step towards it with correcting the syntax issues.

### DIFF
--- a/skt.py
+++ b/skt.py
@@ -1,17 +1,18 @@
 #!/usr/bin/python2
 
-# Copyright (c) 2017 Red Hat, Inc. All rights reserved. This copyrighted material
-# is made available to anyone wishing to use, modify, copy, or
+# Copyright (c) 2017 Red Hat, Inc. All rights reserved. This copyrighted
+# material is made available to anyone wishing to use, modify, copy, or
 # redistribute it subject to the terms and conditions of the GNU General
 # Public License v.2 or later.
 #
-# This program is distributed in the hope that it will be useful, but WITHOUT ANY
-# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-# PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
 #
 # You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 import ConfigParser
 import argparse
@@ -24,11 +25,15 @@ import os
 import shutil
 import sys
 import time
-import skt, skt.runner, skt.publisher, skt.reporter
+import skt.publisher
+import skt.reporter
+import skt.runner
+import skt
 
 DEFAULTRC = "~/.sktrc"
 logger = logging.getLogger()
 retcode = 0
+
 
 def save_state(cfg, state):
     for (key, val) in state.iteritems():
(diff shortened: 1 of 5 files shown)


spbnick:
Thanks a lot, Veronika! This looks great, but as the change is immense and we barely have any tests (nevertheless they worked :)) we need more eyes on this before we merge it.

@agustinhenze, can I ask you take a look at this within 24 hours? If you won't, that's OK, I'll just assume you didn't have time.

agustinhenze:
Wow @veruu outstanding PR! I wanted to do this but you arrived first which is great! @spbnick of course, I can take a look tomorrow.

However I have a preliminary comment :). I think we should add the pep8 (pycodestyle) check on .travis.yml for the whole source code now it is possible.

agustinhenze:
I've executed `pycodestyle $(find . -name "*.py")` and I got `./skt/__init__.py:372:9: E722 do not use bare except'`. I know it would be a behavior change but you can write a test case first and after that be sure that you catch the specific exception is meant to be caught.

veruu:
> I've executed pycodestyle $(find . -name "*.py") and I got ./skt/__init__.py:372:9: E722 do not use bare except'. I know it would be a behavior change but you can write a test case first and after that be sure that you catch the specific exception is meant to be caught.

I mentioned in the commit message that I left out this exception because I currently lack knowledge of skt (as the focus of my work is elsewhere). I simply went through all the errors that had obvious solutions, as the moment I opened the code my vim exploded. Writing a proper test case would require digging around the code to find out what it's supposed to do and what are the corner cases.

Feel free to fix the exception in addition to this patch if you are familiar with the codebase though!

spbnick:
Regarding the bare `except` in `skt/__init__.py`, yeah, we need to figure out what `git_cmd` does, document its exceptions, and then fix it. I'll see if I can figure it out.

spbnick:
OK, I think [that bare `except`](https://github.com/veruu/skt/blob/74bd757e0e0c10d9003ff1f6e77a8662cae8aaea/skt/__init__.py#L372) should be for `subprocess.CalledProcessError`, the function is trying to catch `git am` failing to apply the patches. @veruu, could you please amend your patch to do that? Then @agustinhenze can take his look and if everything else looks fine we'll merge it, and enable the CI style checks if possible.

veruu:
> OK, I think that bare except should be for subprocess.CalledProcessError

Added the exception and merged commits together. Let's get the CI checks!

agustinhenze:
If you want to add the CI check it'd be:
```
diff --git a/.travis.yml b/.travis.yml
index f68d1cb..be02fc3 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,4 @@ install:
 script:
     - python -m unittest discover tests
     - pylint tests
-    - pycodestyle tests
+    - pycodestyle .
```

agustinhenze:
LGTM Everything seems to be OK. I think all the comparisons like `if x is not None` and `if x is None` should be `if x` and `if not x` correspondingly but It's better if it's made in a separate commit or maybe in another PR because it is a behavior change.

spbnick:
Thanks a lot, everyone, merged! This is awesome :) Let's enable style checks in a separate PR. If the comparison is not strictly mandated let's not change it now, but wait until we have more tests.

veruu:
> I think all the comparisons like if x is not None and if x is None should be if x and if not x correspondingly but It's better if it's made in a separate commit or maybe in another PR because it is a behavior change.

That really depends on what are the possible values of the data in comparison and desired outcomes. Things like empty lists `[]` pass the `if not x` condition, but not `if x is not None`.

agustinhenze:
Oh yes, you're right but I have took a look closely and most of them I am pretty sure that wasn't the intention to catch `[]` or `""` in those comparisons. But yes, I agree with @spbnick let's no change it for now.

PR #17, opened 2018-03-19
WIP: README.md: Expand Usage section

spbnick:

PR #19, opened 2018-03-20
WIP: PoC First smoke test

agustinhenze:
Here is a PoC of the smoke test following the steps described on #17. The minimal config is not so minimal yet as it could be and it does the ci pipeline too slow. I guess more than 25 minutes (at the moment that I'm writing this, it didn't finish yet...). I have rewritten the .travis.yml using stages + docker, so you can see if the lint check pass before wait the whole pipeline ends. I'm sure we can improve the speed a lot from here. The most important for me now, is to discuss if the approach is good or not.

"Easy" to speed up
* Creating a custom repository where it can pull the kernel source squashed in a single commit
* Using a minimal config that allow us to be sure it is useful of course

I would like to know your thoughts about it before following in that direction


agustinhenze:
You can see the output pipeline here https://travis-ci.org/agustinhenze/skt/builds/355851651. It uses secret variables therefore it won't pass here.

agustinhenze:
I forgot to ping you, please take a look at it :point_up: @spbnick @stefwalter

spbnick:
Wow, impressive work, Agustin!

Here's my perspective.

First thing, I would like to keep CI fast, at least until development slows down. For that reason, I wonder if we can perhaps create a small repo in this "organization", which would look like a kernel to skt, but would "build" very fast. Or rather generate that repo as part of the integration tests, so they don't depend on GitHub being available, and so that they don't break independently of this repo.

I would also like to have any non-trivial tests added to Travis CI runs available for easy execution without it. I.e., first of all, they should be available to developers for running on their machines easily, and only then used by Travis CI.

I'm on the fence regarding Docker. On one hand, sure, it's easy to create a test environment with it, but on the other hand it's hard to debug. It's not helped by my minimal experience with it, so at the very least I would like more explanations of what's going on there :)

Lastly, I am very reluctant to bring such an explosion of complexity to our CI.

Do you think you can remake this as a shell or Python script running skt without Docker in a temporary or, perhaps, specified directory, and using a dummy git repo generated on the fly, and then add that script to .travis.yml?

@stefwalter, what's your opinion?

spbnick:
By "generating a git repo" I meant something like this (from an old project of mine): https://github.com/spbnick/carton/blob/master/tests/test_git.sh

agustinhenze:
> First thing, I would like to keep CI fast, at least until development slows
> down. For that reason, I wonder if we can perhaps create a small repo in this
> "organization", which would look like a kernel to skt, but would "build" very
> fast. Or rather generate that repo as part of the integration tests, so they
> don't depend on GitHub being available, and so that they don't break
> independently of this repo.

Great idea! Create a bare git repository with the minimal things needed, so skt
can see it like a kernel as you said above :). But we still depending on a
patchwork server running :thinking:. We could use
https://github.com/getpatchwork/patchwork/blob/master/docker-compose.yml to
solve this. It means much more code to write and maintain but I agree it's
worth it, to keep a CI stable.

> I would also like to have any non-trivial tests, added to Travis CI runs,
> available for easy execution without it. I.e., first of all, they should be
> available to developers for running on their machines easily, and only then
> used by Travis CI.

Yes, I totally agree with you about both the testing and development
environments must be reproducible by any developer. I think using containers is
a good way to achieve that, all you need here is docker and docker-compose.
Then everyone can do exactly the same thing that travis does on this branch.
Conversely, if you need to reproduce what we have on master, we don't have any
chance because we can't control the travis' environment. I understand this is a
little more complicated but it solves the reproducibility of the testing and
development environments.

> I'm on the fence regarding Docker. On one hand, sure, it's easy to create a
> test environment with it, but on the other hand it's hard to debug. It's not
> helped by my minimal experience with it, so at the very least I would like more
> explanations of what's going on there :)

Actually it's the opposite, it's easy to debug because if something fails on CI
you can reproduce it easily on your machine. At least in my experience, the
benefits obtained, pay for the complexity added. I'll write a summary trying to
make the curve softer.

> Lastly, I am very reluctant to bring such an explosion of complexity to our CI.

Ups, my idea is to write an integration test using docker-compose. In this way
everyone could have a complete setup with just a single command. I have been
playing a little bit with beaker server here
https://github.com/agustinhenze/beaker-lab

> Do you think you can remake this as a shell or Python script running skt
> without Docker in a temporary or, perhaps, specified directory, and using a
> dummy git repo generated on the fly, and then add that script to .travis.yml?

Ok, I could do that, but I'm not so convinced it's a good idea. IMO, having a
decent way to reproduce the environment both for development and for testing
it's really important to allow enter new contributors easily. And assuming you
already have all dependencies installed and running on your machine is not an
easy way for someone who doesn't understand anything about the project and
want to contribute.

PR #21, opened 2018-03-22
CI: Enable style checks on all files

spbnick:


spbnick:
@agustinhenze, @veruu, style checks are in!

PR #22, opened 2018-03-22
Convert to installable python module

### LINKED ISSUE #9: Rename "skt.py" to "skt"
Rename the `skt.py` program to just `skt`. We don't need the `.py` on the end, and we don't need it to be confused with a module.

major:
This patch converts the repo to an installation python module that can
be easily installed via pip. The main executable, `skt`, will be placed
in the appropriate executable directory.

This patch includes documentation updates to reflect the new executable
name.

Fixes #9 


setup.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python
```
agustinhenze: pylint output says

```************* Module setup
C:  1, 0: Missing module docstring (missing-docstring)

-----------------------------------
Your code has been rated at 5.00/10```

File: setup.cfg [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+author = Red Hat, Inc.
+license = GPL
+
+[options]
+packages = find:
+include_package_data = True
+
+[options.entry_points]
+console_scripts =
+    skt = skt.executable:main
```
agustinhenze: Maybe the path `bin/skt` is a more intuitive place to put it and if in the future we have more than one executable we already have a canonical folder to put it.

setup.py
major: Oops, I'll get that fixed.

setup.cfg
major: This should cause the `skt` executable to land in `~/.local/bin` or `/usr/bin` on the installer's system. We could add extra executables by adding more `console_scripts` entries later if we need to.

major:
@agustinhenze Good comments. I've pushed a new patch.

setup.cfg
agustinhenze: You're right, `console_scripts` put the binary in the right place when you install the project. But I was talking about inside the repository not in the delivery stage.

major: Oh, I see what you're saying now. Let me see what I can do.

major: @agustinhenze To be fair, you could use `pip install -e .` in the skt directory to install the package in develop mode. You'll be able to make changes within the local repository without needing to fully reinstall the package. Would that be sufficient?

setup.cfg
```diff
@@ -0,0 +1,19 @@
+[metadata]
+name = skt
+description = "Sonic Kernel Testing"
+long_description = file: README.md
+version = 0.0.1
+author = Red Hat, Inc.
+license = GPL
```
stefwalter: Should this be GPLv2+?

stefwalter:
I've tested this and used it in  building the container in #28 Can confirm that it works.

setup.cfg
agustinhenze: @major I agree with you that using `pip install -e .` is the recommended way to development. Maybe we should mention it on the README.

I'm not explaining properly, sorry. The name that you picked `executable.py` is not the happiest choice for me. What I'm proposing you is, you move the skt.py to bin/skt. I think that's a more intuitive path for the file and also fixes the issue #9.

setup.cfg
agustinhenze: Yes, GPLv2+ is the right license. Quoting one of the headers `GNU General Public License v.2 or later.`

major: Ah, good catch. I'll fix that up.

README.md
```diff
@@ -31,6 +31,18 @@ For running some specific tests you can do this as following:
 
     $ python -m unittest tests.test_publisher
 
+
+Installation
+------------
+
+Install `skt` directly from git:
+
+    $ pip install git+https://github.com/RH-FMK/skt
```
spbnick: Could you also add a paragraph about "installation" for development, from already checked-out repo?

setup.cfg
```diff
@@ -0,0 +1,19 @@
...
+name = skt
+description = "Sonic Kernel Testing"
+long_description = file: README.md
+version = 1
+author = Red Hat, Inc.
+license = GPLv2+
+
+[options]
+packages = find:
+include_package_data = True
```
spbnick: Could you please add comments for each of these options, explaining what they do? I failed to find documentation for these quickly, and comments might be useful for others too. Unless this is obvious for all other Python developers, of course. In that case a pointer to the documentation here would suffice :) Thanks!

README.md
major: Can do.

setup.cfg
major: Yes, I'll do that now.

spbnick:
Thanks a lot for the comments in `setup.cfg`, @major! We only need a paragraph on development setup (from pre-existing checkout) in README.md and we can merge this :)

major:
Okay, I'll write up that paragraph right now.

major:
@stefwalter & @spbnick -- let me know if this paragraph helps.

stefwalter:
WFM. Ship it.

agustinhenze:
We have so many PR nested with this one, we need to merge it before the list keeps growing

spbnick:
Merged, thanks a lot, Major, everyone!
@major, could you please submit a little PR clarifying where the executable is installed, in the new "development" section?

spbnick:
@agustinhenze, I also miss the directly-accessible executable in the sources. From my discussion with @major it seemed like it's unavoidable if we want to package this with pip. If you two could figure out how to get the executable back, that would be great :)

PR #23, opened 2018-03-22
Fix linting errors and add docstrings

major:
This PR fixes some pylint errors and adds docstrings to some
class modules.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -30,6 +30,7 @@ class runner(object):
 
 
 class beakerrunner(runner):
+    """Run tests on Beaker"""
     TYPE = 'beaker'
 
     def __init__(self, jobtemplate, jobowner=None):
@@ -66,6 +67,18 @@ class beakerrunner(runner):
         logging.info("beaker template: %s", self.template)
 
     def getxml(self, replacements):
+        """
+        Generate job XML with template replacements applied. Search the
+        template for words surrounded by "##" strings and replace them with
+        strings from the supplied dictionary.
+
+        Args:
+            replacements:   A dictionary of placeholder strings with "##"
+                            around them, and their replacements.
+
+        Returns:
+            The job XML text with template replacements applied.
+        """
         xml = ''
         with open(self.template, 'r') as f:
             for line in f:
@@ -79,6 +92,18 @@ class beakerrunner(runner):
         return xml
 
     def getresultstree(self, jobid, logs=False):
+        """
+        Retrieve Beaker job results in Beaker's native XML format.
+
+        Args:
+            jobid: The Beaker jobid
+            logs:  Set to 'True' to retrieve logs from Beaker as part of the
+                   job results. The default is 'False', which excludes logs
+                   from the Beaker job results.
+
+        Returns:
+            The job results XML text with job logs (if logs==True).
+        """
         args = ["bkr", "job-results"]
         if not logs:
             args.append("--no-logs")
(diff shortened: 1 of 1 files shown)


skt/runner.py
```diff
@@ -98,6 +102,7 @@ def dumpjunitresults(self, jobid, junit):
             (stdout, stderr) = bkr.communicate()
 
     def getconsolelog(self, jobid=None):
+        """Retrieves console log from a running beaker test."""
```
spbnick: Could you please keep the descriptions in imperative mood? I.e. write "Retrieve", not "Retrieves". This way, I think, it's easier to reason about functions and quicker to grasp the essence of what they do.

skt/runner.py
```diff
@@ -79,6 +92,17 @@ def getxml(self, replacements):
         return xml
 
     def getresultstree(self, jobid, logs=False):
+        """
+        Retrieve Beaker job results in Beaker's native XML format.
+
+        Args:
+            jobid: The beaker jobid
+            logs:  Set to 'True' to retrieve logs from beaker. The default is
+                   'False'.
```
spbnick: Could you please clarify a little what happens with `False`? I read this like the logs would then be retrieved from somewhere else, and was confused. I was probably thinking "logs" are the "results". Perhaps mentioning in the function description, that log inclusion into results is optional would help.

skt/runner.py
```diff
@@ -98,6 +122,15 @@ def dumpjunitresults(self, jobid, junit):
             (stdout, stderr) = bkr.communicate()
 
     def getconsolelog(self, jobid=None):
+        """
+        Retrieve console log from a Beaker job.
```
spbnick: I think the description should say "log URL". Otherwise this is confusing, as e.g. I assumed this would return the logs themselves. We also need to fix the function name in a separate commit (not necessarily now).

skt/runner.py
```diff
@@ -98,6 +122,15 @@ def dumpjunitresults(self, jobid, junit):
...
+
+        Args:
+            jobid: The Beaker jobid
```
spbnick: What is called "jobid" in the code, in [official, modern Jenkins terminology](https://jenkins.io/doc/book/glossary/) is "build ID". So, we need to describe this properly, and have a separate commit fixing the names (might be done later).

skt/runner.py
major: Fixed.

skt/runner.py
major: Fixed.

skt/runner.py
major: Agreed.

major:
@spbnick I think https://github.com/RH-FMK/skt/pull/23/commits/17d8b16a40d64db325aeb7a720727f76c25b080c should resolve your comments. Let me know if there's anything else remaining.

skt/runner.py
spbnick: Could you please make this say "The Beaker build ID"? Sorry, if I wasn't clear before. We can change the variable name to match later. This is needed to avoid confusion between "projects" and "builds", as earlier they were both often called "job".

skt/runner.py
```diff
@@ -96,9 +96,10 @@ def getresultstree(self, jobid, logs=False):
         Retrieve Beaker job results in Beaker's native XML format.
 
         Args:
-            jobid: The beaker jobid
-            logs:  Set to 'True' to retrieve logs from beaker. The default is
-                   'False'.
+            jobid: The Beaker jobid
+            logs:  Set to 'True' to retrieve logs from Beaker as part of the
+                   job results. The default is 'False', which excludes logs
+                   from the Beaker job results.
```
spbnick: I think this should go into the previous commit.

major: Fixed.

skt/runner.py
spbnick: Heey, I was properly tired here :) I confused Beaker and Jenkins, niiice :) This is good to go, and I just pushed this to master. Thank you, Major! I fixed it up a little bit, though: spelled out "Beaker job ID" instead of "Beaker jobid", and terminated the argument description with a full stop.

PR #25, opened 2018-03-22
Add flake8 to Travis

### LINKED ISSUE #24: Add flake8 checks
We should run `flake8` against the code in Travis to ensure we're following good coding guidelines. Linting should also happen before the code is tested to save time.

major:
This patch set adds flake8 testing in the Travis job and also runs
the linting checks prior to the python tests.

Fixes for flake8 issues are included.

Depends on #22 to merge first.
Fixes #24.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ python:
     - "2.7"
 install:
     - pip install requests junit_xml beaker-client mock
-    - pip install pylint pycodestyle
+    - pip install pylint pycodestyle flake8
 script:
+    - flake8 --show-source .
     - python -m unittest discover tests
     - pylint tests
-    - pycodestyle .
     - pip install .
     - pip uninstall -y skt
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -37,6 +37,10 @@ def stringify(v):
         return str(v)
 
 
+class RpcProtocolMismatch(Exception):
+    """Exception for Patchwork API mismatch errors"""
+
+
 # PatchWork adds a magic API version with each call
 # this class just magically adds/removes it
 class RpcWrapper:
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -125,7 +125,7 @@ def cmd_merge(cfg):
         idx = 0
         for mb in cfg.get('merge_ref'):
             save_state(cfg, {'mergerepo_%02d' % idx: mb[0],
-                             'mergehead_%02d' % idx: head})
+                             'mergehead_%02d' % idx: bhead})
             (retcode, head) = ktree.merge_git_ref(*mb)
 
             utypes.append("[git]")
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -124,7 +124,7 @@ def cmd_merge(cfg):
     try:
         idx = 0
         for mb in cfg.get('merge_ref'):
-            save_state(cfg, {'meregerepo_%02d' % idx: mb[0],
+            save_state(cfg, {'mergerepo_%02d' % idx: mb[0],
                              'mergehead_%02d' % idx: head})
             (retcode, head) = ktree.merge_git_ref(*mb)
 
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -19,16 +19,18 @@ import argparse
 import ast
 import datetime
 import json
-import junit_xml
 import logging
 import os
 import shutil
 import sys
 import time
+
+import junit_xml
+
+import skt
 import skt.publisher
 import skt.reporter
 import skt.runner
-import skt
 
 DEFAULTRC = "~/.sktrc"
 logger = logging.getLogger()
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -11,21 +11,19 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-
-import bkr.client
-from email.mime.application import MIMEApplication
-from email.mime.multipart import MIMEMultipart
-from email.mime.text import MIMEText
+import StringIO
 import gzip
 import logging
 import os
 import re
-import requests
 import smtplib
-import StringIO
-import tempfile
-import xml.etree.ElementTree as etree
-import xmlrpclib
+
+from email.mime.application import MIMEApplication
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+import requests
+
 import skt
 import skt.runner
 
@@ -354,7 +352,6 @@ class reporter(object):
                     continue
 
                 (res, system, clogurl, slshwurl, llshwurl) = rdata
-                ssys = system.split(".")[0]
 
                 result.append("Test Run: #%d" % jidx)
                 result.append("result: %s" % res)
--- a/tests/test_publisher.py
+++ b/tests/test_publisher.py
@@ -15,6 +15,7 @@ Test cases for publisher module.
 # this program; if not, write to the Free Software Foundation, Inc., 51
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 import unittest
+
 from skt import publisher
 
 
--- a/tests/test_reporter.py
+++ b/tests/test_reporter.py
@@ -14,11 +14,15 @@ Test cases for reporter module.
 # You should have received a copy of the GNU General Public License along with
 # this program; if not, write to the Free Software Foundation, Inc., 51
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-import unittest
-from contextlib import contextmanager
 import re
+import unittest
+
+from contextlib import contextmanager
+
 import mock
+
 from skt import reporter
+
 from tests import misc
 
 


README.md
```diff
@@ -31,6 +31,34 @@ For running some specific tests you can do this as following:
...
+    -h, --help            show this help message and exit
+    -d WORKDIR, --workdir WORKDIR
+                            Path to work dir
+    -w, --wipe            Clean build (make mrproper before building), remove
+                            workdir when finished
+    --junit JUNIT         Path to dir to store junit results in
+    -v, --verbose         Increase verbosity level
+    --rc RC               Path to rc file
+    --state               Save/read state from 'state' section of rc file
+
```
spbnick: Could we perhaps not include the `skt -h` output here, and e.g. simply say that `skt -h` should produce command-line help, or such? This way README.md would be shorter and we won't have to update it every time we change the `skt -h` output.

setup.cfg
```diff
@@ -0,0 +1,19 @@
+[metadata]
+name = skt
+description = "Sonic Kernel Testing"
+long_description = file: README.md
+version = 0.0.1
```
spbnick: Would you be open to just having `version = 1` here? I'm aware of semantic versioning, but, I find that so many numbers in version are not needed until at least half a dozen releases. Meanwhile we'll be having major changes in the code and version number changes won't reflect that. We can always add dots and number when we need them.

setup.py
```diff
@@ -0,0 +1,20 @@
...
+# details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Install skt using setuptools."""
+from setuptools import setup
+
+
+setup()
```
spbnick: I don't know how Python packaging works, sorry, but does this file have to have `.py` in the name?

skt/executable.py
```diff
@@ -19,16 +19,18 @@
...
 import sys
 import time
+
+import junit_xml
+
+import skt
 import skt.publisher
 import skt.reporter
 import skt.runner
-import skt
```
spbnick: I'm curious, what's the logic behind the import reordering you're doing here and other files? Could you perhaps put those into a separate commit and elaborate there?

skt/executable.py
```diff
@@ -123,7 +125,7 @@ def cmd_merge(cfg):
         idx = 0
         for mb in cfg.get('merge_ref'):
             save_state(cfg, {'meregerepo_%02d' % idx: mb[0],
-                             'mergehead_%02d' % idx: head})
+                             'mergehead_%02d' % idx: bhead})
```
spbnick: Hmm, could you please make this a separate commit and elaborate a little bit on what's going on? Thanks!

skt/__init__.py
```diff
@@ -37,6 +37,11 @@ def stringify(v):
         return str(v)
 
 
+class RpcProtocolMismatch(Exception):
+    """Exception handler for Patchwork API mismatch errors."""
+    pass
+
+
```
spbnick: Hmm, did you intend to use this somewhere and forget?

README.md
```diff
@@ -82,7 +110,7 @@ default.
 
 E.g. to build with the current system's config file run:
 
-    $ skt.py --workdir skt-workdir -vv \
+    $ skt --workdir skt-workdir -vv \
              build -c /boot/config-`uname -r`
```
spbnick: Ah, those feel nice :relieved:.

.travis.yml
```diff
@@ -4,6 +4,7 @@ python:
 install:
     - pip install requests junit_xml beaker-client mock
     - pip install pylint pycodestyle
+    - pip install .
```
spbnick: Shouldn't this be a part of tests instead, i.e. be in the `script` list? And should we perhaps add an uninstall test as well?

.travis.yml
```diff
@@ -3,9 +3,11 @@ python:
     - "2.7"
 install:
     - pip install requests junit_xml beaker-client mock
-    - pip install pylint pycodestyle
+    - pip install pylint pycodestyle flake8
     - pip install .
+before_script:
+    - flake8 --show-source .
+    - pycodestyle .
```
spbnick: Travis docs [say](https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build) that any failure in `before_script` aborts the whole "build", i.e. none of the `script` commands would run. Was that your intention?

spbnick:
Ah, one thing I missed: what would be the procedure to run skt from the source tree after this change?

.travis.yml
major: Sounds reasonable. I'll make those changes.

README.md
major: That works for me.

setup.cfg
major: Sure. ;)

setup.py
major: Yes, the `.py` is required because pip/setuptools look for `setup.py` and `setup.cfg` (among other files) when installing a new python module.

major:
Waiting for #22 to merge first...

.travis.yml
```diff
@@ -3,8 +3,12 @@ python:
     - "2.7"
 install:
     - pip install requests junit_xml beaker-client mock
-    - pip install pylint pycodestyle
+    - pip install pylint pycodestyle flake8
+before_script:
+    - flake8 --show-source .
+    - pycodestyle .
```
agustinhenze: Quoting the flake8's doc:
```
Flake8 is a wrapper around PyFlakes, pycodestyle, and Ned's McCabe script.
Flake8 runs all tools by launching the single 'flake8' script, but ignores pycodestyle and PyFlakes extended options and just uses defaults.
```
So I think that you should remove `pycodestyle .` item.

major: Ah, good catch, @agustinhenze. I'll fix that now.

skt/__init__.py
major: This exception was referenced in the code already, but it didn't exist. That caused a linting error.

skt/executable.py
major: That is being done to please `flake8`. There are ordering and spacing requirements for modules that come from core python, extra modules installed later, and the skt project itself.

spbnick: Wow, that's heavy. And I thought *I* was a perfectionist :smiley:.

skt/__init__.py
spbnick: Riight. Then we need the same change in sktm, because it has the same code there (we'll also probably need a library shared by skt and sktm).

Could you fix the description, though? This is not an "Exception handler", just an "Exception". Also, please don't put a full stop after a single-sentence docstring.

skt/executable.py
```diff
@@ -124,7 +124,7 @@ def cmd_merge(cfg):
     try:
         idx = 0
         for mb in cfg.get('merge_ref'):
-            save_state(cfg, {'meregerepo_%02d' % idx: mb[0],
+            save_state(cfg, {'mergerepo_%02d' % idx: mb[0],
```
spbnick: Hey, good catch!

.travis.yml
```diff
@@ -3,10 +3,11 @@ python:
...
 install:
     - pip install requests junit_xml beaker-client mock
-    - pip install pylint pycodestyle
+    - pip install pylint pycodestyle flake8
+before_script:
+    - flake8 --show-source .
 script:
     - python -m unittest discover tests
     - pylint tests
-    - pycodestyle .
```
spbnick: Does flake8 include pycodestyle tests, all of them?

.travis.yml
```diff
@@ -3,10 +3,11 @@ python:
     - "2.7"
 install:
     - pip install requests junit_xml beaker-client mock
-    - pip install pylint pycodestyle
+    - pip install pylint pycodestyle flake8
+before_script:
+    - flake8 --show-source .
```
spbnick: Hmm, this means that nothing from `script` would run, if this fails. Was that your intention?

.travis.yml
major: Yes. @agustinhenze pointed that out to me in another PR. Hooray for less code! :smile:

.travis.yml
major: Is `script` a directory somewhere? I'm not sure where that might be.

Running `flake8` like this is recommended in their docs and it's a common practice when using `tox` for python testing, too.

major:
@spbnick Just pushed a fix for the exception issue you noted.

skt/__init__.py
```diff
@@ -37,6 +37,11 @@ def stringify(v):
         return str(v)
 
 
+class RpcProtocolMismatch(Exception):
+    """Exception for Patchwork API mismatch errors"""
+    pass
```
agustinhenze: `pass` is not needed here

major: Fixed in latest push. ;)

skt/__init__.py
spbnick: I think this should be in the first patch of the PR.

.travis.yml
spbnick: Right, then we'll need to remove `pycodestyle` from [CODING.md](https://github.com/RH-FMK/meta/blob/master/CODING.md), once we switch sktm to flake8.

.travis.yml
spbnick: No, I meant `script` array next to `before_script` in `.travis.yml`. Travis CI doc [says](https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build) if any of the `before_script` commands fails, nothing under `script` runs, or so I understood. I'm not sure why flake8 recommends that, but perhaps because they think it's very important. I'd say being able to also get test results in a single CI run is a good thing, so I would move that to `script`.

major:
@spbnick Sorry for putting that `pass` removal in the last commit. I guess I was in a hurry and squashed it into that last one. ;)

spbnick:
Thanks a lot, Major, merged!

PR #26, opened 2018-03-23
README: Add dependency on dnf-plugins-core

stefwalter:
This is required by the 'dnf builddep' tooling.


stefwalter:
Otherwise

```
No such command: builddep. Please use /usr/bin/dnf --help
It could be a DNF plugin command, try: "dnf install 'dnf-command(builddep)'"
```

PR #28, opened 2018-03-23
WIP: Initial Docker container

stefwalter:
Here's an initial container definition. It's pretty general purpose at this point and just allows running the various commands in a containerized fashion.

 * [x] This depends on pull request #22 from @major which needs to be merged first. 
 * [ ] #31 should be merged first, required for volumes to work right
 * [x] Around the time this is merged, we should push an image to docker.io hub.
 * [x] Add @major to kerneltest docker hub team.
 * [ ] Add @spbnick to kerneltest docker hub team.
 * [ ] Add @veruu to kerneltest docker hub team.


agustinhenze:
Awesome! :). I wrote a Dockerfile too on #19 but with a different purpose. Also I've configured travis to allow push the image to docker.io hub. But those variables are secret using my docker login, we need to setup an account to allow push the images from travis.

Also I've rewritten the .travis.yml using stages and the docker image generated, please take a look at it and steal whatever you want from there.

skt/executable.py
```diff
@@ -385,7 +385,8 @@ def setup_parser():
     """
     parser = argparse.ArgumentParser()
 
-    parser.add_argument("-d", "--workdir", type=str, help="Path to work dir")
+    parser.add_argument("-d", "--workdir", type=str, help="Path to work dir",
+                        default=os.environ.get("WORKDIR"))
```
spbnick: Could you use name "SKT_WORKDIR" instead, so it plays better with other environment variables?

skt/executable.py
```diff
@@ -420,7 +420,8 @@ def setup_parser():
 
     parser_build = subparsers.add_parser("build", add_help=False)
     parser_build.add_argument("-c", "--baseconfig", type=str,
-                              help="Path to kernel config to use")
+                              help="Path to kernel config to use",
+                              default=os.environ.get("BASECONFIG"))
```
spbnick: Same here, could you use name "SKT_BASECONFIG" instead?

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN git --work-tree /var/work config user.email "skt@example.com" && \
+	git --work-tree /var/work config user.name "Sonic Kernel Testing"
+
+# We don't pollute the container file system  tree with
+# massive intermediate build output
+ENV KBUILD_OUTPUT=/var/build
+VOLUME [ "/var/build" ]
+
+COPY . /var/src/skt
+RUN cd /var/src/skt && ls && pip install -v .
```
spbnick: Do you think you can put everything under the same directory, so people don't have to hunt down these non-standard locations? Perhaps under "/root", where shell starts? Or, if that doesn't work, "/home/skt", or "/var/lib/skt"?

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+	bison flex python2-mock dnf-plugins-core
+RUN dnf builddep -y kernel
+
+# Use a default checkout of the kernel as our workdir Also make it the default directory
+# and setup an environment variable to tell the skt executable to use this workdir
+#
+# Notice that we give 777 and 666 permissions on this directory. This is because Openshift
+# likes to run containers as random UIDs, and we cannot rely on SKT being root
+RUN git clone git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git /var/work && \
+	chmod -R ugo+w /var/work
```
spbnick: If you want to use this clone as `skt` workdir, then I think the proper way to do it would be using `skt` itself.
I.e. `skt --state merge --baserepo git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git` (provided we have the environment set and working directory created).

Otherwise we can have a bare clone as a local mirror which all the further `skt` commands would use as `--baserepo`, with `--workdir` someplace else. That would make creating new "workdirs" faster, but would need a bit more instructions, which might not be what we want.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+WORKDIR /var/work
+
+# Pull a default kernel config down for use in the container
+# It's expected that this will be overriden either by mounting a
+# new config file, or by adding one in a layered container
+RUN mkdir -p /var/config && \
+        git clone --depth=1 https://src.fedoraproject.org/git/rpms/kernel.git /tmp/kernel && \
+        mv /tmp/kernel/*.config /var/config/ && \
+        rm -rf /tmp/kernel
+ENV BASECONFIG=/var/config/kernel-x86_64.config
```
spbnick: Do you think you can pull the config directly from `https://src.fedoraproject.org/rpms/kernel/raw/master/f/kernel-x86_64.config` with wget/curl?

README.md
```diff
@@ -121,6 +121,26 @@ You can make skt output junit-compatible results by adding a `--junit
...
+purposes, or looking at intermediate files:
+
+    $ sudo docker run -ti --entrypoint=/bin/bash kerneltest/skt
+
+If you change the Dockerfile that defines the container, you can rebuild
+the container:
+
+    $ sudo docker build -t docker.io/kerneltest/skt:latest .
+    $ sudo docker push docker.io/kerneltest/skt:latest
+
```
spbnick: Wait, does that push to the same image for everyone? How does that work without authentication?

skt/executable.py
stefwalter: Done in #31

skt/executable.py
stefwalter: Done.

README.md
stefwalter: I can add accounts.

spbnick: I mean it wouldn't work for just somebody outside, would it? If so, then let's put the second line in a separate paragraph, saying this is for developers with appropriate access only.

Dockerfile
stefwalter: In Openshift, the containers won't start as a consistent UID. It will be random. So /root and /home/skt will be awkward.

I'd like to /var. 
 * I wanted to use ```/var/src``` and sub-directories like ```/var/src/kernel``` for source and git repositories.
 * And ```/var/build``` and sub-directories for built output.

See #31

Dockerfile
stefwalter: Good point. Will do.

stefwalter:
@spbnick So there's a problem ... If we use the SKT command to set up the workdir ... there's a massive overhead to creating a new container. Here's why:

 * Each command like ```RUN``` in a ```Dockerfile``` creates a layer in the resulting image.
 * ```docker build``` caches every layer, and does not regenerate them unless the inputs have changed.
 * If a layer has changed, then all following layers will change.
 * Only layers that change are pushed with ```docker push```
 * The skt command ```pip install``` is near the end for a reason: It will change the most often when iterating on the container.
 * The kernel checkout is pretty early in the list for a reason: We want it to rarely change, since it represents 3G of download and upload (when pushing the container). 

Therefore putting the kernel checkout *before* bringing SKT into the container is a necessary optimization for making rebuilding, pushing, pulling and iterating on the container manageable.

In addition, using a kernel checkout as the initial bootstrap of a workdir seems to work just fine. One of the first commands issued to skt in the container can and should be ```merge```.

Dockerfile
stefwalter: This ended up being problematic, as seen below. I don't think we should block the initial version of the container on this. If it is possible to do, we should experiment with it in a follow up.

Dockerfile
stefwalter: Done.

README.md
stefwalter: Good point. Clarified in the README.

PR #31, opened 2018-03-27
Allow building out of tree

stefwalter:
This is just basic good practice, not to mix your build artifacts with your source code.
    
However in this case it has a concrete benefit, it allows a container to work properly. A container can carry around the basic kernel source code staged in its image, and still allow building to happen in a throw away empty volume.
    
Later this can grow into work to have a possible prebuilt tree in an additional volume or directory, thus likely speeding up the build by only building the diff. We don't implement that yet with this patch.

 * [x] #22 needs to be merged
 * [x] #55 needs to be merged


stefwalter:
Rebased now that #22 is done.

stefwalter:
Look at #28 for an example of how this is used with ```/work``` as a srcdir and ```/work/build``` as a builddir. Although this idea works in different hierarchies anyway.

README.md
```diff
@@ -126,11 +126,11 @@ written to the `<JUNIT_DIR>` directory.
 
 To checkout a kernel tree run:
 
-    $ skt --rc <SKTRC> --state --workdir <WORKDIR> -vv \
+    $ skt --rc <SKTRC> --state --workdir <SRCDIR> -vv \
```
spbnick: I think we should keep calling this "workdir". Partly because the option is still named "workdir", and mostly because I cannot vouch at the moment that it is not used for anything else beside just keeping the source.

skt/__init__.py
```diff
@@ -464,15 +470,15 @@ def prepare(self, clean=True):
             logging.info("cleaning up tree: %s", args)
             subprocess.check_call(args)
 
-        shutil.copyfile(self.basecfg, "%s/.config" % self.path)
+        shutil.copyfile(self.basecfg, self.get_cfgpath())
```
spbnick: This doesn't seem to be related, could you please do it in a separate commit?

stefwalter: It's related, but separated.

README.md
stefwalter: Done. Moved all back to ```workdir``` ```wdir``` and ```WORKDIR```.

skt/executable.py
```diff
@@ -423,7 +430,8 @@ def setup_parser():
     parser_build.add_argument("--cfgtype", type=str, help="How to process "
                               "default config (default: olddefconfig)")
     parser_build.add_argument("--makeopts", type=str,
-                              help="Additional options to pass to make")
+                              help="Additional options to pass to make",
+                              default=os.environ.get("MAKEFLAGS"))
```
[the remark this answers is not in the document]
stefwalter: Whoops, this was just in there by accident. No need for it. Removed ... at least for now.

stefwalter:
Updated.

skt/executable.py
```diff
@@ -385,10 +387,15 @@ def setup_parser():
     """
     parser = argparse.ArgumentParser()
 
-    parser.add_argument("-d", "--workdir", type=str, help="Path to work dir")
+    parser.add_argument("-d", "--workdir", type=str,
+                        help="Path to work dir with kernel source git tree",
```
spbnick: I'm not sure we can write this kind of description for this option. Since workdir can also contain the build, and possibly other data. At least it should say "contains source as well as build, unless -b points somewhere else", or something like that.

skt/executable.py
```diff
@@ -385,10 +387,15 @@ def setup_parser():
...
+                        default=os.environ.get("SKT_WORKDIR"))
```
spbnick: Could you please make adding the environment variable retrieval a separate commit as well?

skt/__init__.py
```diff
@@ -136,9 +136,8 @@ def __init__(self, uri, ref=None, wdir=None):
         # FIXME Move expansion up the call stack, as this limits the class
         # usefulness, because tilde is a valid path character.
         # The git "working directory" (the "checkout")
-        self.wdir = (os.path.expanduser(wdir)
-                     if wdir is not None
-                     else tempfile.mkdtemp())
+        self.wdir = (os.path.expanduser(wdir) if wdir is not None else
+                     tempfile.mkdtemp())
```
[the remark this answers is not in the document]
stefwalter: Sorry, this was just from changing things back after review.

skt/executable.py
stefwalter: Just changed this back. Less relevant now that we no longer refer to SRCDIR

skt/executable.py
stefwalter: Done.

stefwalter:
An easy way to test this is:

```
$ sudo docker pull docker.io/kerneltest/skt:latest
$ sudo docker run --rm docker.io/kerneltest/skt -vvv build
```

stefwalter:
Rebased on master. Had to update for disable debug_info patch.

spbnick:
Stef, could you please rebase your PR? I had to fix building with disabled debuginfo in master and that intersects with your changes. Thank you.

stefwalter:
Rebased.

spbnick:
Thank you, @stefwalter. I'm still having problem making it work, investigating.

stefwalter:
> Thank you, @stefwalter. I'm still having problem making it work, investigating.

Could you describe your workflow here in GitHub?

spbnick:
@stefwalter, at the moment I'm running these commands with an already-existing skt-rc and skt-workdir:

    skt --rc skt-rc --state --workdir skt-workdir -vv          cleanup
    skt --rc skt-rc --state --workdir skt-workdir -vv          merge --baserepo git://git.kernel.org/pub/scm/linux/kernel/git/davem/net-next.git                --ref a870a02cc963de35452bbed932560ed69725c4f2 --pw https://patchwork.ozlabs.org/patch/886637
    skt --rc skt-rc --state --builddir skt-builddir --workdir skt-workdir -vv          build -c /boot/config-`uname -r`

I get the following output for the last command:

      File "/home/nkondras/skt/skt/executable.py", line 631, in main
        args.func(cfg)
      File "/home/nkondras/skt/skt/executable.py", line 108, in wrapper
        func(cfg)
      File "/home/nkondras/skt/skt/executable.py", line 189, in cmd_build
        raise e
    Exception: Failed to find tgz path in stdout

spbnick:
And here's the relevant piece of the build.log:

      HOSTCC  arch/x86/tools/relocs_32.o
      Using .. as source for kernel
      .. is not clean, please run 'make mrproper'
      in the '..' directory.

spbnick:
Updated the comment with the output to include the full output.

spbnick:
Nope, apparently not the relative path, but previous builds in the workdir? After running "make mrproper" manually in the workdir, it works.

stefwalter:
> Nope, apparently not the relative path, but previous builds in the workdir? After running "make mrproper" manually in the workdir, it works.

Yes, that's unfortunately how the kernel build process works. See: https://www.linux.com/learn/kernel-newbie-corner-building-and-running-new-kernel Section "
Configuring in a Remote Directory"

```
Finally (and even though it's still not clear why this is necessary), the source tree being 
used for a remote build must be clean. That is, you can't be using the same source tree
for both a local build and a remote build. Why this is true is not entirely clear, but that it's 
true is obvious if you check the top-level Makefile where, down around line 953, the 
prepare3 target does the following checking:

 prepare3: include/config/kernel.release
 ifneq ($(KBUILD_SRC),)         [doing a remote build?]
   @$(kecho) '  Using $(srctree) as source for kernel'
   $(Q)if [ -f $(srctree)/.config -o -d $(srctree)/include/config ]; then \
     echo "  $(srctree) is not clean, please run 'make mrproper'";\
     echo "  in the '$(srctree)' directory.";\
     /bin/false; \
   fi;
```
Should I put a failure check if someone tries to change the builddir once a build exists?

In addition, you ran headlong into the problem that both my pull request from two weeks ago #30 and @agustinhenze open pull request

skt/__init__.py
```diff
@@ -484,13 +491,18 @@ def prepare(self, clean=True):
 
         shutil.copyfile(self.basecfg, self.get_cfgpath())
 
+        args = self.defmakeargs + [self.cfgtype]
+        logging.info("prepare config: %s", args)
+        subprocess.check_call(args)
+        self._ready = 1
+
```
spbnick: I think you might have forgotten to remove this from below, when moving it, so now we would have two of these. Or am I missing something?

stefwalter: Whoops, that got introduced in all the rebases. Removed.

spbnick:
Unfortunately, this doesn't work. Executing these command:

    mkdir skt-builddir2
    skt --rc skt-rc --state --workdir skt-workdir2 --builddir skt-builddir2 -vv merge --baserepo git://git.kernel.org/pub/scm/linux/kernel/git/davem/net-next.git --ref a870a02cc963de35452bbed932560ed69725c4f2 --pw https://patchwork.ozlabs.org/patch/886637
    skt --wipe --rc skt-rc --state --builddir skt-builddir2 --workdir skt-workdir2 -vv build -c /boot/config-`uname -r`

Produces the following exception:

      File "/home/nkondras/skt/skt/executable.py", line 631, in main
        args.func(cfg)
      File "/home/nkondras/skt/skt/executable.py", line 108, in wrapper
        func(cfg)
      File "/home/nkondras/skt/skt/executable.py", line 189, in cmd_build
        raise e
    Exception: Failed to find tgz path in stdout

And `tail skt-builddir2/build.log` outputs:

	  INSTALL sound/x86/snd-hdmi-lpe-audio.ko
	  INSTALL virt/lib/irqbypass.ko
	  DEPMOD  4.16.0-rc4+
	'./System.map' -> './tar-install/boot/System.map-4.16.0-rc4+'
	'.config' -> './tar-install/boot/config-4.16.0-rc4+'
	'./vmlinux' -> './tar-install/boot/vmlinux-4.16.0-rc4+'
	'./arch/x86/boot/bzImage' -> './tar-install/boot/vmlinuz-4.16.0-rc4+'
	Tarball successfully created in ./linux-4.16.0-rc4+-x86.tar.gz
	make[1]: Leaving directory '/home/nkondras/skt-workdir2/skt-builddir2'
	make: Leaving directory '/home/nkondras/skt-workdir2'

Apart from fixing this, I think it would be good to create the builddir
automatically, if it doesn't exist.

PR #33, opened 2018-03-27
Make building with debuginfo optional

### LINKED ISSUE #32: Need the ability to disable debuginfo
The debuginfo option is enabled by default and we need an option to disable it. For merge/build/boot tests, those debug symbols aren't critical.

A kernel built with debuginfo is ~ 250MB (tar.gz) and one without is ~ 70MB.

major:
By default, kernels are built with debug symbols. This increases the
build time and the size of the resulting tarball.

    tar.gz with debug symbols: ~ 250MB
    tar.gz without debug symbols: ~ 75MB

If debug symbols are really needed for diagnosing a tough problem,
the user can specify `--enable-debuginfo` when calling `skt build`.

Fixes #32.

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -431,7 +431,8 @@ class ktree(object):
 
 
 class kbuilder(object):
-    def __init__(self, path, basecfg, cfgtype=None, makeopts=None):
+    def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
+                 enable_debuginfo=False):
         # FIXME Move expansion up the call stack, as this limits the class
         # usefulness, because tilde is a valid path character.
         self.path = os.path.expanduser(path)
@@ -465,6 +466,15 @@ class kbuilder(object):
             subprocess.check_call(args)
 
         shutil.copyfile(self.basecfg, "%s/.config" % self.path)
+
+        # NOTE(mhayden): Building kernels with debuginfo can increase the
+        # final kernel tarball size by 3-4x and can increase build time
+        # slightly. Debug symbols are really only needed for deep diagnosis
+        # of kernel issues on a specific system. This is why debuginfo is
+        # disabled by default.
+        if not self.enable_debuginfo:
+            logging.info("disabling debuginfo:")
+            subprocess.check_call(["script/config", "--disable debug_info"])
 
         args = self.defmakeargs + [self.cfgtype]
         logging.info("prepare config: %s", args)
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -169,8 +169,13 @@ def cmd_build(cfg):
     tstamp = datetime.datetime.strftime(datetime.datetime.now(),
                                         "%Y%m%d%H%M%S")
 
-    builder = skt.kbuilder(cfg.get('workdir'), cfg.get('baseconfig'),
-                           cfg.get('cfgtype'), cfg.get('makeopts'))
+    builder = skt.kbuilder(
+        cfg.get('workdir'),
+        cfg.get('baseconfig'),
+        cfg.get('cfgtype'),
+        cfg.get('makeopts'),
+        cfg.get('enable_debuginfo')
+    )
 
     try:
         tgz = builder.mktgz(cfg.get('wipe'))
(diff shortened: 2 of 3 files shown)


stefwalter:
Needs a rebase.

major:
@stefwalter Done!

stefwalter:
Just a cross check on the default here.

So I'm interested in the "why" ... do you expect building a kernel without debuginfo to be routine and normal case for SKT runs? If not, why is it the default? 

Does a non-debuginfo kernel significantly increase the difficulty of diagnosing a test failure? @dzickusrh @veruu Any comments?

skt/__init__.py
```diff
@@ -466,6 +468,12 @@ def prepare(self, clean=True):
 
         shutil.copyfile(self.basecfg, "%s/.config" % self.path)
 
+        # Disable debuginfo by default unless we have been asked to enable it.
+        if not self.enable_debuginfo:
+            logging.info("disabling debuginfo:")
+            args = shlex.split('script/config --disable debug_info')
+            subprocess.check_call(args)
```
stefwalter: This should probably just be:

```
subprocess.check_call(['script/config', '--disable', 'debug_info'])
```

And drop the ```shlex``` usage? What do you think?

veruu:
> Does a non-debuginfo kernel significantly increase the difficulty of diagnosing a test failure? @dzickusrh @veruu Any comments?

Based on my experience, debug info is only useful if you want to run the resulting image through `gdb` / `objdump` etc but it doesn't add anything otherwise. Will we keep kernel images that failed testing somewhere for future debugging? If not, we don't need debug info, IMHO (unless there is another use case that I'm currently overlooking).

stefwalter:
Sounds like defaulting to without debuginfo makes sense.

skt/__init__.py
major: I'm not opposed, but `shlex` is usually my go-to for increasing readability. ;)

I'll change it.

major:
Okay, the patch is updated. I also added a quick note in the `README.md` about the change.

dzickusrh:
This comment is a little late, but just to add to it.  I agree with Veronika, for the most part debuginfo isn't useful.  The only piece in there that folks use is 'vmlinux' to feed into crash.  Most local builds we strip out the debug symbols to massively reduce the size.
I usually accomplish that with a for-loop and a 'strip -g' on all kernel modules.  I didn't realize a simple DEBUG_INFO option could be disabled.  Learn something new everyday..

skt/__init__.py
```diff
@@ -466,6 +467,15 @@ def prepare(self, clean=True):
 
         shutil.copyfile(self.basecfg, "%s/.config" % self.path)
 
+        # NOTE(mhayden): Building kernels with debuginfo can increase the
+        # final kernel tarball size by 3-4x and can increase build time
+        # slightly. Debug symbols are really only needed for deep diagnosis
+        # of kernel issues on a specific system. This is why debuginfo is
+        # disabled by default.
```
spbnick: Hey, nice touch with the comment :)

skt/executable.py
```diff
@@ -169,8 +169,13 @@ def cmd_build(cfg):
...
 
-    builder = skt.kbuilder(cfg.get('workdir'), cfg.get('baseconfig'),
-                           cfg.get('cfgtype'), cfg.get('makeopts'))
+    builder = skt.kbuilder(
+        cfg.get('workdir'),
+        cfg.get('baseconfig'),
+        cfg.get('cfgtype'),
+        cfg.get('makeopts'),
+        cfg.get('enable_debuginfo')
+    )
```
spbnick: This is getting a little out of hand. We'll need to convert this to named arguments next.

skt/executable.py
```diff
@@ -422,6 +427,12 @@ def setup_parser():
                               help="Path to kernel config to use")
     parser_build.add_argument("--cfgtype", type=str, help="How to process "
                               "default config (default: olddefconfig)")
+    parser_build.add_argument(
+        "--enable-debuginfo",
+        type=bool,
+        default=False,
+        help="build kernel with debuginfo (default: disabled)"
```
spbnick: Could you please start the help text with a capital letter to match the rest?

skt/__init__.py
major: Thanks!

skt/executable.py
major: Agreed. I can work on that.

skt/executable.py
major: Sure.

spbnick:
Thank you, Major :)!

PR #39, opened 2018-03-27
Add test coverage for __init__.py

major:
Signed-off-by: Major Hayden <major@mhtx.net>

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -1,0 +1,69 @@
+"""
+Test cases for __init__.py.
+"""
+# Copyright (c) 2018 Red Hat, Inc. All rights reserved. This copyrighted
+# material is made available to anyone wishing to use, modify, copy, or
+# redistribute it subject to the terms and conditions of the GNU General Public
+# License v.2 or later.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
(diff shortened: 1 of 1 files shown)


tests/test_init.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_parse_bad_patchwork_url(self):
+        """Ensure parse_patchwork_url() handles a parsing exception."""
+        patchwork_url = "garbage"
+        with self.assertRaises(Exception) as context:
+            skt.parse_patchwork_url(patchwork_url)
+
+            self.assertTrue(
+                "Can't parse patchwork url: '{}'".format(patchwork_url)
+                in context.exception
+            )
```
spbnick: Hmm, I'm not quite getting how this works. Should this really be inside `with`?

major: That's the only spot were the `context` variable is available to me.

major:
@spbnick In my last software project(s), the full stop was required. I will do my best to remember it here! :)

spbnick:
@major, full stops are a matter of aesthetics mostly, but I already started documenting the code this way :)

tests/test_init.py
spbnick: Hmm, then what would happen if the assertion checking throws the same exception, and why does [this code in the documentation](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises) has the test outside? Sorry, my context manager-fu is weak.

major: Weird, it does work outside the `with:`. I removed the indention in the latest push.

major:
Should be fixed in 887a8ab, @spbnick.

spbnick:
Thank you, Major, merged!

PR #40, opened 2018-03-27
Show progress on kernel building

agustinhenze:
Try to follow PR #30 but with a different approach plus test cases.

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -20,6 +20,10 @@ import shutil
 import subprocess
 import tempfile
 import xmlrpclib
+import sys
+import io
+import time
+from threading import Timer
 
 
 def stringify(v):
@@ -538,8 +542,25 @@ class kbuilder(object):
 
         return krelease
 
-    def mktgz(self, clean=True):
-        tgzpath = None
+    def mktgz(self, clean=True, timeout=60 * 60 * 12):
+        """
+        Build kernel and modules, after that, pack everything into a tarball.
+
+        Args:
+            clean:      If it is True, run the `mrproper` target before build.
+            timeout:    Max time in seconds will wait for build.
+        Returns:
+            The full path of the tarball generated.
+        Raises:
+            CommandTimeoutError: When building kernel takes longer than the
+                                 specified timeout.
+            CalledProcessError:  When a command returns an exit code different
+                                 than zero.
+            ParsingError:        When can not find the tarball path in stdout.
+            IOError:             When tarball file doesn't exist.
+        """
+        fpath = None
+        stdout_list = []
         self.prepare(clean)
 
         args = self.defmakeargs + ["INSTALL_MOD_STRIP=1",
@@ -548,24 +569,75 @@ class kbuilder(object):
 
         logging.info("building kernel: %s", args)
 
-        mk = subprocess.Popen(args,
-                              stdout=subprocess.PIPE,
-                              stderr=subprocess.STDOUT)
-        (stdout, stderr) = mk.communicate()
-        for line in stdout.split("\n"):
-            m = re.match("^Tarball successfully created in (.*)$", line)
-            if m:
-                tgzpath = m.group(1)
-                break
+        with io.open(self.buildlog, 'wb') as writer, \
+                io.open(self.buildlog, 'rb') as reader:
+            make = subprocess.Popen(args,
+                                    stdout=writer,
+                                    stderr=subprocess.STDOUT)
+            make_timedout = []
 
-        fpath = None
-        if tgzpath is not None:
-            fpath = "/".join([self.path, tgzpath])
+            def stop_process(proc):
+                """
+                Terminate the process with SIGTERM and flag it as timed out.
+                """
+                if proc.poll() is None:
+                    proc.terminate()
+                    make_timedout.append(True)
+            timer = Timer(timeout, stop_process, [make])
+            timer.setDaemon(True)
+            timer.start()
+            try:
+                while make.poll() is None:
+                    self.append_and_log2stdout(reader.readlines(), stdout_list)
+                    time.sleep(1)
+                self.append_and_log2stdout(reader.readlines(), stdout_list)
+            finally:
+                timer.cancel()
+            if make_timedout:
+                raise CommandTimeoutError(
+                    "'{}' was taking too long".format(' '.join(args))
+                )
+            if make.returncode != 0:
+                raise subprocess.CalledProcessError(make.returncode,
+                                                    ' '.join(args))
 
-        if fpath is None or not os.path.isfile(fpath):
-            with open(self.buildlog, "w") as fp:
-                fp.write(stdout)
+        match = re.search("^Tarball successfully created in (.*)$",
+                          ''.join(stdout_list), re.MULTILINE)
+        if match:
+            fpath = os.path.realpath(os.path.join(self.path, match.group(1)))
+        else:
+            raise ParsingError('Failed to find tgz path in stdout')
 
-            raise Exception("Failed to find tgz path in stdout")
+        if not os.path.isfile(fpath):
+            raise IOError("Built kernel tarball {} not found".format(fpath))
 
         return fpath
+
+    @staticmethod
+    def append_and_log2stdout(lines, full_log):
+        """
+        Append `lines` into `full_log` and show `lines` on stdout.
+
+        Args:
+            lines:      list of strings.
+            full_log:   list where `lines` members are appended.
+        """
+        full_log.extend(lines)
+        sys.stdout.write(''.join(lines))
+        sys.stdout.flush()
+
+
+class CommandTimeoutError(Exception):
+    """
+    Exception raised when a timeout occurs on a process which has had timeouts
+    enabled. The accompanying value is a string whose value is the command
+    launched plus a small explanation.
+    """
+
+
+class ParsingError(Exception):
+    """
+    Exception raised when a regex does not match and it is impossible to
+    continue. The accompanying value is a string which explains what it can not
+    find.
+    """
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -14,7 +14,14 @@ Test cases for __init__.py.
 # You should have received a copy of the GNU General Public License along with
 # this program; if not, write to the Free Software Foundation, Inc., 51
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+from __future__ import division
 import unittest
+import tempfile
+import shutil
+import os
+import subprocess
+import mock
+from mock import Mock
 
 import skt
 
@@ -53,3 +60,105 @@ class TestInit(unittest.TestCase):
             "Can't parse patchwork url: '{}'".format(patchwork_url)
             in context.exception
         )
+
+
+class KBuilderTest(unittest.TestCase):
+    # (Too many instance attributes) pylint: disable=R0902
+    """Test cases for skt.kbuilder class"""
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp()
+        self.tmpconfig = tempfile.NamedTemporaryFile()
+        self.kbuilder = skt.kbuilder(self.tmpdir, self.tmpconfig.name)
+        self.m_popen = Mock()
+        self.m_popen.returncode = 0
+        self.ctx_popen = mock.patch('subprocess.Popen',
+                                    Mock(return_value=self.m_popen))
+        self.ctx_check_call = mock.patch('subprocess.check_call', Mock())
+        self.m_io_open = Mock()
+        self.m_io_open.__enter__ = lambda *args: self.m_io_open
+        self.m_io_open.__exit__ = lambda *args: None
+        self.ctx_io_open = mock.patch('io.open',
+                                      Mock(return_value=self.m_io_open))
+        self.kernel_tarball = 'linux-4.16.0.tar.gz'
+        self.success_str = 'Tarball successfully created in ./{}\n'
+        self.success_str = self.success_str.format(self.kernel_tarball)
+
+    def tearDown(self):
+        shutil.rmtree(self.tmpdir)
+
+    def test_mktgz_clean(self):
+        """
+        Check if `mrproper` target is called when `clean` is True and not
+        called when `clean` is False.
+        """
+        with self.ctx_popen, self.ctx_check_call as m_check_call:
+            self.assertRaises(Exception, self.kbuilder_mktgz, clean=True)
+            self.assertEqual(
+                m_check_call.mock_calls[0],
+                mock.call(['make', '-C', self.tmpdir, 'mrproper'])
+            )
+            m_check_call.reset_mock()
+            self.assertRaises(Exception, self.kbuilder_mktgz, clean=False)
+            self.assertNotEqual(
+                m_check_call.mock_calls[0],
+                mock.call(['make', '-C', self.tmpdir, 'mrproper']),
+            )
+
+    def test_mktgz_timeout(self):
+        """
+        Check if timeout error is raised when kernel building takes longer than
+        specified timeout.
+        """
+        self.m_popen.poll = Mock(side_effect=[None, None, -15])
+        self.m_popen.returncode = -15
+        from time import sleep as real_sleep
+
+        def m_sleep(seconds):
+            """sleep function but acelerated 100x"""
+            real_sleep(seconds/100)
+        ctx_sleep = mock.patch('time.sleep', m_sleep)
+        with ctx_sleep, self.ctx_check_call, self.ctx_popen:
+            self.assertRaises(skt.CommandTimeoutError,
+                              self.kbuilder_mktgz, timeout=0.001)
+
+    def test_mktgz_parsing_error(self):
+        """Check if ParsingError is raised when no kernel is found in stdout"""
+        self.m_io_open.readlines = Mock(return_value=['foo\n', 'bar\n'])
+        with self.ctx_popen, self.ctx_check_call, self.ctx_io_open:
+            self.assertRaises(skt.ParsingError, self.kbuilder_mktgz)
+
+    def test_mktgz_ioerror(self):
+        """Check if IOError is raised when tarball path does not exist"""
+        self.m_io_open.readlines = Mock(
+            return_value=['foo\n', self.success_str]
+        )
+        with self.ctx_io_open, self.ctx_popen, self.ctx_check_call:
+            self.assertRaises(IOError, self.kbuilder_mktgz)
+
+    def test_mktgz_make_fail(self):
+        """
+        Check if subprocess.CalledProcessError is raised when make command
+        fails to spawn.
+        """
+        self.m_popen.returncode = 1
+        with self.ctx_popen:
+            self.assertRaises(subprocess.CalledProcessError,
+                              self.kbuilder_mktgz)
+
+    def test_mktgz_success(self):
+        """Check if mktgz can finish successfully"""
+        self.m_io_open.readlines = Mock(
+            return_value=['foo\n', self.success_str, 'bar']
+        )
+        self.m_popen.returncode = 0
+        with self.ctx_io_open, self.ctx_popen, self.ctx_check_call:
+            with open(os.path.join(self.tmpdir, self.kernel_tarball), 'w'):
+                pass
+            full_path = self.kbuilder_mktgz()
+            self.assertEqual(os.path.join(self.tmpdir, self.kernel_tarball),
+                             full_path)
+
+    def kbuilder_mktgz(self, *args, **kwargs):
+        """Wrapper of kbuilder.mktgz to avoid stdout pollution"""
+        with mock.patch('sys.stdout'):
+            return self.kbuilder.mktgz(*args, **kwargs)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: python
 python:
     - "2.7"
-script:
+install:
     # Install skt using pip to ensure dependencies are downloaded correctly.
     - pip install .
+script:
     # Run tests and linters.
     - flake8 --show-source .
     - pylint tests


stefwalter:
From experience with reliable automated infrastructure, having a 60 minute timeout is risky. Could you bump it to 10 or 12 hours? We really want to catch the pathological case here.

skt/__init__.py
```diff
@@ -503,24 +508,53 @@ def mktgz(self, clean=True):
...
+            timer.setDaemon(True)
+            timer.start()
+            while make.poll() is None:
+                self.append_and_log2stdout(reader.readlines(), stdout_list)
+                time.sleep(1)
+            self.append_and_log2stdout(reader.readlines(), stdout_list)
+            if not timer.isAlive():
+                raise CommandTimeoutError(
+                    "'%s' was taking too long" % ' '.join(args)
+                )
```
spbnick: It seems to me this can race with the timer handler (`make.terminate`) and produce an error even if `make` has finished on time. For example, if it finished during that `time.sleep(1)` or during blocked output to stdout. Could you perhaps write a timer handle which runs `make.poll()` and sets a variable if `make` was still alive on timeout, and *then* calls `make.terminate()`? Then you can check that variable and throw an error if it's set.

BTW, shouldn't we also be checking make's exit status?

skt/__init__.py
```diff
@@ -498,6 +498,17 @@ def getrelease(self):
...
+        """
+        Build kernel and modules, after that, pack everything into a tarball.
+
+        Args:
+            clean:      If it's True, run the `mrproper` target before build.
+            timeout:    Max time in seconds will wait for build, 1h by default.
+        Returns:
+            The full path of the tarball generated.
+        Raises:
+            CommandTimeoutError: When building kernel takes too long.
```
spbnick: I think this needs to say "takes longer than the specified timeout".

skt/__init__.py
```diff
@@ -503,24 +508,53 @@ def mktgz(self, clean=True):
...
-            fpath = "/".join([self.path, tgzpath])
+        with io.open(self.buildlog, 'wb') as writer, \
+                io.open(self.buildlog, 'rb') as reader:
+            make = subprocess.Popen(args,
+                                    stdout=writer,
+                                    stderr=subprocess.STDOUT)
+
+            timer = Timer(timeout, make.terminate)
+            # If the process is cancelled, it doesn't need to wait for anything
+            timer.setDaemon(True)
```
spbnick: Thanks for the comment above, but what does this line actually do? I'm unfamiliar with this API, and wasn't able to find documentation on this quickly.

agustinhenze:
> From experience with reliable automated infrastructure, having a 60 minute timeout is risky. Could you bump it to 10 or 12 hours? We really want to catch the pathological case here.

@stefwalter ok, bumped to 12 hours. Please take a look at it and it'd be great if you test it a little bit as well.

agustinhenze:
> Also, we can have documentation addition in the same commit, as this technically changes the function interface, and documentation should be updated together with that.

Squashed

skt/__init__.py
```diff
@@ -503,24 +508,53 @@ def mktgz(self, clean=True):
...
+            full_log:   list where `lines` members are appended.
+        """
+        full_log.extend(lines)
+        sys.stdout.write(''.join(lines))
+        sys.stdout.flush()
+
+
+class CommandTimeoutError(Exception):
+    """
+    This exception is raised when a timeout occurs on a process which has had
```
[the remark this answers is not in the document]
agustinhenze: Done

skt/__init__.py
agustinhenze: Yes, it wasn't checking the returncode. I have added some test cases that should validate the correct behavior. Please take a look at it again.

skt/__init__.py
agustinhenze: Done

skt/__init__.py
agustinhenze: I have rewritten the comment, I hope it's understandable now

spbnick: Sorry, it's probably obvious to regular Python developers (and if that's true, perhaps no comment is necessary), but not to me. What's the difference between `setDaemon(False)` and `setDaemon(True)` in timer behavior? A link to documentation would also be fine. Thanks :)

skt/__init__.py
spbnick: Thanks a lot :smiley:! However, timer could still have expired right after make failed, but before we checked, and then we would report the wrong error. Do you think you can fix that?

skt/__init__.py
```diff
@@ -503,24 +524,68 @@ def mktgz(self, clean=True):
...
+    Exception raised when a timeout occurs on a process which has had timeouts
+    enabled. The accompanying value is a string whose value is the command
+    launched plus a small explanation.
+    """
+
+
+class RegexNotFound(Exception):
+    """
+    Exception raised when a regex does not match and it is impossible to
+    continue. The accompanying value is a string whose explains what it can not
```
spbnick: I think this should say "which", not "whose".

skt/__init__.py
```diff
@@ -503,24 +524,68 @@ def mktgz(self, clean=True):
...
+        if match:
+            fpath = os.path.realpath(os.path.join(self.path, match.group(1)))
+        else:
+            raise RegexNotFound('Failed to find tgz path in stdout')
 
-        fpath = None
-        if tgzpath is not None:
-            fpath = "/".join([self.path, tgzpath])
+        if not os.path.isfile(fpath):
+            raise IOError("No such file {}".format(fpath))
```
spbnick: Can you elaborate here a little? E.g. say "Built kernel tarball {} not found"?

skt/__init__.py
```diff
@@ -493,8 +497,25 @@ def getrelease(self):
...
+
+        Args:
+            clean:      If it is True, run the `mrproper` target before build.
+            timeout:    Max time in seconds will wait for build.
+        Returns:
+            The full path of the tarball generated.
+        Raises:
+            CommandTimeoutError: When building kernel takes longer than the
+                                 specified timeout.
+            CalledProcessError:  When a command returns an exit code different
```
spbnick: Shouldn't this say "When "make" returns an exit code..."? Or do you want to keep that abstracted away, or plan adding other commands to this?

agustinhenze: Yes, I'd like to keep it abstracted because it calls to `.prepare` which uses check_call and the latter raises this exception is something goes wrong.

skt/__init__.py
agustinhenze: The difference in terms of behavior between a daemon timer and a non-daemon timer is none. Timer is a subclass of Thread, that means it is necessary to read the thread section https://docs.python.org/2/library/threading.html#thread-objects, I have to say that the doc is not clear here... The long story short for our case is, if we don't set as daemon this timer, the main process cannot be stopped (e.g. Ctrl+C or SIGTERM) because it will keep waiting until the timer/thread finish. We can solve this too, calling timer.cancel() on SIGTERM or Ctrl+C but I think this way is much simpler.

Now that you mention it, I think it'd be better remove the comment here...

skt/__init__.py
agustinhenze: oh, I didn't get it before. Good catch! Please take a look at it now.

agustinhenze:
I think, actually I'm pretty sure, that this commit 0d3c31f626a0bcbc1f2d1dfd759c1669d5dfb2ff broke master.
```
2018-03-28 19:18:12,803     INFO   basecfg: tests/smoke/kernel-minimal-config
2018-03-28 19:18:12,803     INFO   cfgtype: olddefconfig
2018-03-28 19:18:12,803    DEBUG   state: buildlog -> ../skt-workdir/build.log
  File "skt/executable.py", line 627, in <module>
    main()
  File "skt/executable.py", line 616, in main
    args.func(cfg)
  File "skt/executable.py", line 105, in wrapper
    func(cfg)
  File "skt/executable.py", line 184, in cmd_build
    raise e
AttributeError: 'kbuilder' object has no attribute 'enable_debuginfo'
```

veruu:
> I think, actually I'm pretty sure, that this commit 0d3c31f broke master.

Thanks for reporting, I submitted a pull request to add the missing attribute.

skt/__init__.py
spbnick: I see, thank you for the explanation :)

skt/__init__.py
```diff
@@ -518,24 +539,73 @@ def mktgz(self, clean=True):
...
-            m = re.match("^Tarball successfully created in (.*)$", line)
-            if m:
-                tgzpath = m.group(1)
-                break
+        with io.open(self.buildlog, 'wb') as writer, \
+                io.open(self.buildlog, 'rb') as reader:
+            make = subprocess.Popen(args,
+                                    stdout=writer,
+                                    stderr=subprocess.STDOUT)
+            make.timedout = False
```
spbnick: Hmm, is it normal to just go and add a field to an instance of some 3rd-party class? What if a later version of subprocess adds a field named the same, but used for something else?

skt/__init__.py
```diff
@@ -518,24 +539,73 @@ def mktgz(self, clean=True):
...
+    launched plus a small explanation.
+    """
+
+
+class RegexNotFound(Exception):
+    """
+    Exception raised when a regex does not match and it is impossible to
+    continue. The accompanying value is a string which explains what it can not
+    find.
+    """
```
spbnick: Do you think you can name this something like `ParsingError`, or similar? The `RegexNotFound` is a bit too specific, while at the same time not that helpful.

skt/__init__.py
agustinhenze: You're right! It is not a good idea. Anyway, I was looking at the code again and... The race still there, because if the process ends after the poll call, it will report a timeout error when the process finished correctly... So I think I'm going back to the first implementation which it is simpler. What do you think?

spbnick: Well, if the process was still running at the moment of the poll call, it means it timed out. I.e. it took longer than allowed. So, I think that part is OK.

skt/__init__.py
agustinhenze: OK, I'll rename it to `ParsingError` :+1:

skt/__init__.py
agustinhenze: Fair enough, please take a look at it again.

spbnick: OTOH, previously you simply checked if the timer has expired or not, and that might have happened after make has finished. E.g. because of the sleep in the loop, or because your output to stdout was blocked.

skt/__init__.py
```diff
@@ -518,24 +539,73 @@ def mktgz(self, clean=True):
...
+                    proc.terminate()
+                    make_timedout.append(True)
+            timer = Timer(timeout, stop_process, [make])
+            timer.setDaemon(True)
+            timer.start()
+            while make.poll() is None:
+                self.append_and_log2stdout(reader.readlines(), stdout_list)
+                time.sleep(1)
+            self.append_and_log2stdout(reader.readlines(), stdout_list)
+            timer.cancel()
```
spbnick: Could you wrap the code executing while the timer is running, above, into a try-finally, and put `timer.cancel` into the `finally`, so we're sure the timer gets killed?

tests/test_init.py
```diff
@@ -53,3 +60,99 @@ def test_parse_bad_patchwork_url(self):
...
+        self.m_popen = Mock()
+        self.m_popen.returncode = 0
+        self.ctx_popen = mock.patch('subprocess.Popen',
+                                    Mock(return_value=self.m_popen))
+        self.ctx_check_call = mock.patch('subprocess.check_call', Mock())
+
+    def tearDown(self):
+        shutil.rmtree(self.tmpdir)
+
+    def test_clean_mktgz(self):
```
spbnick: Could you name those so that they all start with `test_mktgz_` and then have specifics on the end? E.g. `test_mktgz_clean`?

tests/test_init.py
```diff
@@ -53,3 +60,99 @@ def test_parse_bad_patchwork_url(self):
...
+        """Check if `mrproper` target is called when clean is True"""
```
spbnick: This checks much more than just the `mrproper` target being "made". Please remove the rest of the tests from here, and please also only check that there is a command which has `make` as the first argument and `mrproper` as one of the following arguments. Otherwise this test is going to break often (including when you rebase now).

tests/test_init.py
```diff
@@ -53,3 +60,99 @@ def test_parse_bad_patchwork_url(self):
...
+            real_sleep(seconds/100)
+        ctx_sleep = mock.patch('time.sleep', m_sleep)
+        with ctx_sleep, self.ctx_check_call, self.ctx_popen:
+            self.assertRaises(skt.CommandTimeoutError,
+                              self.kbuilder.mktgz, timeout=0.001)
+
+    def test_mktgz(self):
+        """
+        Check if ParsingError, IOError, and subprocess.CalledProcessError are
+        raised, as well also check if mktgz can finish successfully.
```
spbnick: Do you think you can split this into separate test functions?

skt/__init__.py
agustinhenze: Great idea! :+1:

tests/test_init.py
agustinhenze: Ok, done

agustinhenze:
> Thank you, Agustin, this is a needed change, and sorry for the delay with the review. Please take a look at my comments, and also please rebase on the latest master. Thank you!

@spbnick Please take a look at it again. I have addressed all your comments and rebased to master.

tests/test_init.py
```diff
@@ -53,3 +60,102 @@ def test_parse_bad_patchwork_url(self):
...
+                m_check_call.mock_calls[0],
+                mock.call(['make', '-C', self.tmpdir, 'mrproper'])
+            )
+            m_check_call.reset_mock()
+            self.assertRaises(Exception, self.kbuilder.mktgz, clean=False)
+            self.assertEqual(m_check_call.call_count, 2)
+            self.assertNotEqual(
+                m_check_call.mock_calls[0],
+                mock.call(['make', '-C', self.tmpdir, 'mrproper']),
+            )
```
spbnick: This still tests the exact arguments, which is making the test fragile. Do you think you can test that argv[0] == "make" and argv[something above zero] == "mrproper", and the reverse?

tests/test_init.py
```diff
@@ -53,3 +60,102 @@ def test_parse_bad_patchwork_url(self):
...
+        self.m_popen = Mock()
+        self.m_popen.returncode = 0
+        self.ctx_popen = mock.patch('subprocess.Popen',
+                                    Mock(return_value=self.m_popen))
+        self.ctx_check_call = mock.patch('subprocess.check_call', Mock())
+        self.m_io_open = Mock()
+        self.m_io_open.__enter__ = lambda *args: self.m_io_open
+        self.m_io_open.__exit__ = lambda *args: None
+        self.ctx_io_open = mock.patch('io.open',
+                                      Mock(return_value=self.m_io_open))
```
spbnick: I'm concerned about the approach of having instance variables shared and updated by tests. This is similar to having global variables. I think this creates an opportunity for a test to use one of those without updating it, and then breaking when tests are reordered or omitted. Do you think you can instead, e.g., write functions creating those values as needed and then use them in the tests, or just directly creating them every time?

tests/test_init.py
```diff
@@ -53,3 +60,99 @@ def test_parse_bad_patchwork_url(self):
...
+        Check if `mrproper` target is called when `clean` is True and not
+        called when `clean` is False.
+        """
+        with self.ctx_popen, self.ctx_check_call as m_check_call:
+            self.assertRaises(Exception, self.kbuilder.mktgz, clean=True)
+            self.assertIn('make', m_check_call.mock_calls[0][1][0])
+            self.assertIn('mrproper', m_check_call.mock_calls[0][1][0])
+            m_check_call.reset_mock()
+            self.assertRaises(Exception, self.kbuilder.mktgz, clean=False)
+            self.assertNotIn('make', m_check_call.mock_calls[0][1][0])
```
spbnick: Strictly-speaking, this should check that `make` is not in the first argument only, not in any position. And the reverse above.

agustinhenze: I think it's **not** going to be better than the previous version in terms of the test fragility, after all, it's checking the exact arguments using hardcoded indexing. And the output when it fails is less intuitive. So I propose two things here, or we remove this test case or back to the previous version.

spbnick: Hmm, perhaps I misunderstood what's being done here, I'm not familiar with Python's `Mock`. I read this specific line (100) as it's checking that there was a command executed, which had no exact string `make` as any element in the `argv` array. E.g. command "echo hello" would pass, command "make something" would fail, but command "echo make" would fail as well and the latter was my concern.

However, I see that this is not as trivial as I thought. Could you perhaps explain what you're doing here, and why you think changing it to what I suggest would be worse? Thank you.

agustinhenze: Yeah, np. In the beginning, you mentioned that the test case was fragile because it checked all the arguments and I got your point. But if you take a look of what we'll have now (after addressing your last comment) and what we had before, it doesn't seem to be better at least for me in the term of fragility.

I didn't push the following code yet. It retrieves the first call to subprocess.check_call with the anti-intuitive [0][1][0] indexing and then compares index [0] == 'make' and so on.
```python
            self.assertEqual('make', m_check_call.mock_calls[0][1][0][0])
            self.assertIn('mrproper', m_check_call.mock_calls[0][1][0][3])
..
            self.assertNotEqual(make', m_check_call.mock_calls[0][1][0][0])
            self.assertNotEqual('mrproper', m_check_call.mock_calls[0][1][0][3])
```
vs
```python
            self.assertEqual(
                m_check_call.mock_calls[0],
                mock.call(['make', '-C', self.tmpdir, 'mrproper'])
            )
...
            self.assertNotEqual(
                m_check_call.mock_calls[0],
                mock.call(['make', '-C', self.tmpdir, 'mrproper']),
            )
```
Both have hardcoded arguments and for me the previous one has a better output when it fails.
```
======================================================================
FAIL: test_mktgz_clean_2 (tests.test_init.KBuilderTest)
Check if `mrproper` target is called when clean is True
----------------------------------------------------------------------
  File "tests/test_init.py", line 109, in test_mktgz_clean_2
    mock.call(['mke', '-C', self.tmpdir, 'mrproper'])
AssertionError: call(['make', '-C', '/tmp/tmpQ1LiQC', 'mrproper']) != call(['mke', '-C', '/tmp/tmpQ1LiQC', 'mrproper'])
```
vs
```
======================================================================
FAIL: test_mktgz_clean (tests.test_init.KBuilderTest)
----------------------------------------------------------------------
  File "tests/test_init.py", line 96, in test_mktgz_clean
    self.assertEqual('mke', m_check_call.mock_calls[0][1][0][0])
AssertionError: 'mke' != 'make'
```
I don't think it is worse, but I am not convinced it is better either hehe. I'm used to seeing comparisons using mock.call, maybe that's why it sounds more intuitive to me. Anyway I want to see this PR merged soon so I'll take the way you prefer and if in the future we find a better way to solve this, we can change it :):

spbnick: Ah, I see, thank you, Agustin. OK, minor correction to what I actually
suggested about `mrproper`: it should be searched for among all arguments. So,
it should rather be:

    self.assertEqual('make', m_check_call.mock_calls[0][1][0][0])
    self.assertIn('mrproper', m_check_call.mock_calls[0][1][0])

However, I agree that readability of both assert and the failure output is bad
in this case, so I think it's OK for us to match the exact command line, and
change this test from time to time as we update it, if that means we can
actually understand the test code and its failures.

Take your pick on how you want this done :)

agustinhenze: Thank you Nikolai! I really like your reviews :smile:. Please take a look at it again to see if it is what you expected.

tests/test_init.py
```diff
@@ -53,3 +60,103 @@ def test_parse_bad_patchwork_url(self):
...
+            self.assertRaises(IOError, self.kbuilder.mktgz)
+
+    def test_mktgz_make_fail(self):
+        """
+        Check if subprocess.CalledProcessError is raised when make command
+        fails to spawn.
+        """
+        self.m_io_open.readlines = Mock(
+            return_value=['foo\n', self.success_str, 'bar']
+        )
```
spbnick: Hmm, we shouldn't get to actually reading the "make" output if it fails to spawn, therefore this is unnecessary and is confusing. Can we remove this, and the `ctx_io_open` use? Also, `ctx_check_call` shouldn't be used either, right?

agustinhenze: Yeah! Done!

agustinhenze:
It seems that pypi is broken.
screenshot_20180416_124903
I don't understand why travis doesn't stop there and keep trying next commands despite pip install dropped an exit code != 0.

spbnick:
@agustinhenze, yeah, something is amiss here. Regarding Travis CI continuing, I think we should put `pip install .` into [`before_script`](https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build). Can you check and submit a PR if that helps?

spbnick:
Or perhaps we should use `install` instead: https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle

agustinhenze:
> Or perhaps we should use install instead: https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle

Yeah! That worked! Well pypi still broken but our CI is improved! :smile:

spbnick:
Thank you, now we have proper dependency checking, and pip is working again, great :) I tried testing your tests, and there's one thing, which I think needs fixing: the tests produce output on stdout when running, which is not at all helpful and breaks normal unittest output:

    .foo
    Tarball successfully created in ./linux-4.16.0.tar.gz
    ..foo
    bar
    .foo
    Tarball successfully created in ./linux-4.16.0.tar.gz
    bar..........
    ----------------------------------------------------------------------
    Ran 14 tests in 0.044s

    OK

Do you think you can make the tests silent?

agustinhenze:
> Do you think you can make the tests silent?

Done

spbnick:
Merged, thank you, Agustin!

PR #41, opened 2018-03-28
Initial test coverage for runner

major:
This patch provides initial test coverage for the `runner` module. More is left to be done, but this is a start! :)

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -117,6 +117,13 @@ class BeakerRunner(Runner):
         return etree.fromstring(stdout)
 
     def dumpjunitresults(self, jobid, junit):
+        """
+        Retrieve junit XML from beaker and write it to a file
+
+        Args:
+            jobid: The beaker jobid
+            junit: The directory where the junit XML will be written
+        """
         args = ["bkr", "job-results", "--format=junit-xml"]
         args.append(jobid)
 
@@ -148,6 +155,12 @@ class BeakerRunner(Runner):
         return url
 
     def _forget_cid(self, cid):
+        """
+        Remove a job or recipe from the current list
+
+        Args:
+            cid: The job (J:xxxxx) or recipe (R:xxxxx) id
+        """
         if cid.startswith("J:"):
             self.jobs.remove(cid)
             for rid in self.j2r[cid]:
@@ -167,6 +180,15 @@ class BeakerRunner(Runner):
             raise ValueError("Unknown cid type: %s" % cid)
 
     def getverboseresults(self, joblist):
+        """
+        Retrieve verbose results for a list of beaker jobs
+
+        Args:
+            joblist: A list of beaker jobs
+
+        Returns:
+            A dictionary with the jobid as a key and job results as the value
+        """
         result = dict()
         for jobid in joblist:
             result[jobid] = dict()
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -1,0 +1,235 @@
+"""
+Test cases for runner module.
+"""
+# Copyright (c) 2018 Red Hat, Inc. All rights reserved. This copyrighted
+# material is made available to anyone wishing to use, modify, copy, or
+# redistribute it subject to the terms and conditions of the GNU General Public
+# License v.2 or later.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+import os
+import tempfile
+import unittest
+
+import xml.etree.ElementTree as etree
+import mock
+
+from skt import runner
+
+from tests import misc
+
+SCRIPT_PATH = os.path.dirname(__file__)
+
+DEFAULT_ARGS = {
+    'jobtemplate': '{}/assets/test.xml'.format(SCRIPT_PATH)
+}
+
+
+class TestRunner(unittest.TestCase):
+    """Test cases for runner module"""
+
+    def setUp(self):
+        """Set up test fixtures"""
+        self.myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        with open('{}/assets/test.xml'.format(SCRIPT_PATH), 'r') as fileh:
+            self.test_xml = fileh.read()
+
+    def test_getrunner(self):
+        """Ensure getrunner() can create a runner subclass"""
+        result = runner.getrunner('beaker', {'jobtemplate': 'test'})
+        self.assertIsInstance(result, runner.BeakerRunner)
+
+    def test_invalid_getrunner(self):
+        """Ensure getrunner() throws an exception for an invalid runner"""
+        bad_runner = 'pizza'
+        with self.assertRaises(Exception) as context:
+            runner.getrunner(bad_runner, {'jobtemplate': 'test'})
+
+        self.assertIn(
+            "Unknown runner type: {}".format(bad_runner),
+            context.exception
+        )
+
+    def test_getxml(self):
+        """Ensure BeakerRunner.getxml() returns xml"""
+        result = self.myrunner.getxml({})
+        self.assertEqual(result, self.test_xml)
+
+    def test_getxml_replace(self):
+        """Ensure BeakerRunner.getxml() returns xml with replacements"""
+        result = self.myrunner.getxml({'TEST': 'replaced text'})
+        expected_xml = self.test_xml.replace("##TEST##", "replaced text")
+        self.assertEqual(result, expected_xml)
+
+    @mock.patch('subprocess.Popen')
+    def test_getresultstree(self, mock_popen):
+        """Ensure getresultstree() works"""
+        test_xml = "<xml><test>TEST</test></xml>"
+        mock_popen.return_value.returncode = 0
+        mock_popen.return_value.communicate.return_value = (test_xml, '')
+        result = self.myrunner.getresultstree(jobid=0)
+        self.assertEqual(next(x.text for x in result.iter('test')), 'TEST')
+
+    @mock.patch('subprocess.Popen')
+    def test_dumpjunitresults(self, mock_popen):
+        """Ensure dumpjunitresults() works"""
+        mock_popen.return_value.returncode = 0
+        mock_popen.return_value.communicate.return_value = (
+            'stdout',
+            'stderr'
+        )
+
+        junit_dir = tempfile.mkdtemp()
+        self.myrunner.dumpjunitresults("J:00001", junit_dir)
+        expected_file = "{}/j_00001.xml".format(junit_dir)
+        self.assertTrue(os.path.exists(expected_file))
+
+        # Clean up
+        os.unlink(expected_file)
+        os.rmdir(junit_dir)
+
+    @mock.patch('skt.runner.BeakerRunner.getresultstree')
+    def test_getconsolelog(self, mock_getresultstree):
+        """Ensure getconsolelog() works"""
+        # Mock up a beaker XML reply
+        mocked_xml = misc.get_asset_content('beaker_results.xml')
+        mock_getresultstree.return_value = etree.fromstring(mocked_xml)
+
+        result = self.myrunner.getconsolelog()
+        self.assertEqual(result, "http://example.com/")
+
+    def test_forget_cid_withj(self):
+        """Ensure _forget_cid() works with jobs"""
+        # pylint: disable=protected-access
+        self.myrunner.jobs = ["J:00001"]
+        self.myrunner.j2r = {"J:00001": ["R:00001"]}
+        self.myrunner.recipes = ["R:00001"]
+        result = self.myrunner._forget_cid("J:00001")
+        self.assertIsNone(result)
+        self.assertEqual(self.myrunner.jobs, [])
+        self.assertEqual(self.myrunner.recipes, [])
+
+    def test_forget_cid_withr(self):
+        """Ensure _forget_cid() works with recipes"""
+        # pylint: disable=protected-access
+        self.myrunner.jobs = ["J:00001"]
+        self.myrunner.j2r = {"J:00001": ["R:00001"]}
+        self.myrunner.recipes = ["R:00001"]
+        result = self.myrunner._forget_cid("R:00001")
+        self.assertIsNone(result)
+        self.assertEqual(self.myrunner.recipes, [])
+
+    def test_forget_cid_bad_job(self):
+        """Ensure _forget_cid() fails with an invalid taskspec"""
+        # pylint: disable=protected-access
+        with self.assertRaises(Exception) as context:
+            self.myrunner._forget_cid("OHCOMEON:00001")
+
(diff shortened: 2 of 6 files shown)


tests/test_runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_invalid_getrunner(self):
+        """Ensure getrunner() throws an exception for an invalid runner"""
+        bad_runner = 'pizza'
+        with self.assertRaises(Exception) as context:
+            runner.getrunner(bad_runner, {'jobtemplate': 'test'})
+            self.assertTrue(
+                "Unknown runner type: {}".format(bad_runner)
+                in context.exception
+            )
```
spbnick: Could you fix this exception assert as well?

tests/test_runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_getxml(self):
+        """Ensure beakerrunner.getxml() returns xml"""
+        args = {
+            'jobtemplate': '{}/assets/test.xml'.format(SCRIPT_PATH)
+        }
+        myrunner = runner.beakerrunner(**args)
+        result = myrunner.getxml({'replace_this_text': 'with_this_text'})
+        self.assertTrue('<xml>' in result)
+        self.assertTrue('This is test xml!' in result)
```
spbnick: The description and the test don't quite match. You seem to also test that a non-existent replacement doesn't affect the retrieval. If that's so, please describe the test appropriately, and also check that the whole text is intact, not just a couple sub-strings. If you simply want to check if XML is returned, then I think you need to pass just an empty dictionary to `getxml`. Otherwise, please explain what I'm not getting here :)

tests/test_runner.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_getxml_replace(self):
+        """Ensure beakerrunner.getxml() returns xml with replacements"""
+        args = {
+            'jobtemplate': '{}/assets/test.xml'.format(SCRIPT_PATH)
+        }
+        myrunner = runner.beakerrunner(**args)
+        result = myrunner.getxml({'TEST': 'replaced text'})
+        self.assertTrue('<xml>' in result)
+        self.assertTrue('This is test xml!' in result)
```
spbnick: You don't seem to test that replacement was made here. Do you think you could check that the whole resulting text matches the expectations?

tests/test_runner.py
major: Fixed.

tests/test_runner.py
major: Good point. I'll use an empty dictionary instead.

tests/test_runner.py
major: Good catch. Fixed.

spbnick: I mean, could you check that the *whole* returned text matches the expectations? I.e. match against the whole returned text, so that we catch any possible text mangling besides the replacement itself, instead of testing just the addition of "replaced text", and the two sub-strings being intact, which you do now.

tests/test_runner.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def test_getxml(self):
+        """Ensure beakerrunner.getxml() returns xml"""
+        args = {
+            'jobtemplate': '{}/assets/test.xml'.format(SCRIPT_PATH)
+        }
+        myrunner = runner.beakerrunner(**args)
+        result = myrunner.getxml({})
+        self.assertTrue('<xml>' in result)
+        self.assertTrue('This is test xml!' in result)
```
spbnick: Can you check that the result matches the file contents exactly? Perhaps load the file in `__init__` beforehand, so it's easier for the tests?

major: Fixed!

major:
@spbnick Should be better now.

tests/test_runner.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @mock.patch('subprocess.Popen')
+    def test_getresultstree(self, mock_popen):
+        """Ensure getresultstree() works"""
+        test_xml = "<xml><test>TEST</test></xml>"
+        mock_popen.return_value.returncode = 0
+        mock_popen.return_value.communicate.return_value = (test_xml, '')
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        result = myrunner.getresultstree(jobid=0)
+        self.assertEqual(next(x.text for x in result.iter('test')), 'TEST')
```
spbnick: I wish there was a less obscure way of doing this, but OK.

tests/test_runner.py
```diff
@@ -0,0 +1,166 @@
...
+    @mock.patch('subprocess.Popen')
+    def test_dumpjunitresults(self, mock_popen):
+        """Ensure dumpjunitresults() works"""
+        mock_popen.return_value.returncode = 0
+        mock_popen.return_value.communicate.return_value = (
+            'stdout',
+            'stderr'
+        )
+
+        junit_dir = tempfile.mkdtemp()
```
spbnick: Could you please take care of removing this directory?

tests/test_runner.py
```diff
@@ -0,0 +1,166 @@
...
+        with open(beaker_xml, 'r') as fileh:
+            mocked_xml = fileh.read()
+        mock_getresultstree.return_value = etree.fromstring(mocked_xml)
+
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        result = myrunner.getconsolelog()
+        self.assertEqual(result, "http://example.com/")
+
+    def test_forget_cid_withj(self):
+        """Ensure _forget_cid() works with J type jobs"""
```
spbnick: The `bkr` manpage [describes the terminology](https://www.mankier.com/1/bkr#Description-Specifying_tasks) applicable here. I'm not sure what Artem meant by "cid", perhaps "check ID" (would be good to find out). However, outside of those we need to use Beaker terminology (i.e. "taskspec", "job", "recipe set", "recipe", etc.), not "J type jobs" or "R type jobs". We need to stay consistent, so that the code is easier to link to Beaker documentation and Web UI and thus easier to understand. Please fix. Later we might need to rename `cid` as well.

Also, please document the `_forget_cid` function before testing it.

tests/test_runner.py
```diff
@@ -0,0 +1,166 @@
...
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        myrunner.jobs = ["J:00001"]
+        myrunner.j2r = {"J:00001": ["R:00001"]}
+        myrunner.recipes = ["R:00001"]
+        result = myrunner._forget_cid("R:00001")
+        self.assertIsNone(result)
+        self.assertEqual(myrunner.recipes, [])
+
+    def test_forget_cid_bad_job(self):
+        """Ensure _forget_cid() fails with unknown jobs"""
```
spbnick: I expect this should be "fails with invalid taskspec".

tests/test_runner.py
```diff
@@ -0,0 +1,166 @@
...
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        with self.assertRaises(Exception) as context:
+            myrunner._forget_cid("OHCOMEON:00001")
+
+        self.assertTrue(
+            "Unknown cid type: OHCOMEON:00001" in context.exception
+        )
+
+    @mock.patch('skt.runner.BeakerRunner.getresultstree')
+    def test_getverboseresults(self, mock_getresultstree):
```
spbnick: Please document `getverboseresults` function before testing it.

tests/test_runner.py
```diff
@@ -0,0 +1,166 @@
...
+        """Ensure getresultstree() works"""
+        test_xml = "<xml><test>TEST</test></xml>"
+        mock_popen.return_value.returncode = 0
+        mock_popen.return_value.communicate.return_value = (test_xml, '')
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        result = myrunner.getresultstree(jobid=0)
+        self.assertEqual(next(x.text for x in result.iter('test')), 'TEST')
+
+    @mock.patch('subprocess.Popen')
+    def test_dumpjunitresults(self, mock_popen):
```
spbnick: Please document `dumpjunitresults` before testing it. A commit in the same PR would be fine.

tests/test_runner.py
major: Done

tests/test_runner.py
major: Done

tests/test_runner.py
major: Done

tests/test_runner.py
```diff
@@ -0,0 +1,260 @@
...
+        os.unlink(expected_file)
+        os.rmdir(junit_dir)
+
+    @mock.patch('skt.runner.BeakerRunner.getresultstree')
+    def test_getconsolelog(self, mock_getresultstree):
+        """Ensure getconsolelog() works"""
+        # Mock up a beaker XML reply
+        beaker_xml = '{}/assets/beaker_results.xml'.format(SCRIPT_PATH)
+        with open(beaker_xml, 'r') as fileh:
+            mocked_xml = fileh.read()
```
agustinhenze: You can use misc.get_asset_content to avoid writing repeated code

tests/test_runner.py
```diff
@@ -0,0 +1,260 @@
...
+        result = runner.getrunner('beaker', {'jobtemplate': 'test'})
+        self.assertIsInstance(result, runner.BeakerRunner)
+
+    def test_invalid_getrunner(self):
+        """Ensure getrunner() throws an exception for an invalid runner"""
+        bad_runner = 'pizza'
+        with self.assertRaises(Exception) as context:
+            runner.getrunner(bad_runner, {'jobtemplate': 'test'})
+
+        self.assertTrue(
```
agustinhenze: Please use assert *In* instead *True*, it shows the proper info when it fails.

tests/test_runner.py
```diff
@@ -0,0 +1,260 @@
...
+            "Unknown runner type: {}".format(bad_runner)
+            in context.exception
+        )
+
+    def test_getxml(self):
+        """Ensure BeakerRunner.getxml() returns xml"""
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
```
agustinhenze: You can put myrunner as attribute class on setUp

tests/test_runner.py
```diff
@@ -0,0 +1,260 @@
...
+                    None,
+                    None,
+                    'http://example.com/',
+                    'http://example.com/machinedesc.log',
+                    'http://example.com/lshw.log'
+                ),
+                'result': 'Pass'
+            }
+        }
+        self.assertEqual(result, expected_result)
```
agustinhenze: s/assertEqual/assertDictEqual/

tests/test_runner.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def test_get_mfhost(self):
+        """Ensure get_mfhost() works"""
```
agustinhenze: What is the purpose of mfhost?

tests/test_runner.py
```diff
@@ -0,0 +1,260 @@
...
+        """Ensure jobresult() works"""
+        # Mock up a beaker XML reply
+        beaker_xml = '{}/assets/beaker_results.xml'.format(SCRIPT_PATH)
+        with open(beaker_xml, 'r') as fileh:
+            mocked_xml = fileh.read()
+        mock_getresultstree.return_value = etree.fromstring(mocked_xml)
+
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        result = myrunner.jobresult("J:00001")
+        self.assertEqual(result, (0, 'Pass'))
```
agustinhenze: s/assertEqual/assertTupleEqual/

tests/test_runner.py
```diff
@@ -0,0 +1,260 @@
...
+        """Ensure jobresult() handles a job failure"""
+        # Mock up a beaker XML reply
+        beaker_xml = '{}/assets/beaker_fail_results.xml'.format(SCRIPT_PATH)
+        with open(beaker_xml, 'r') as fileh:
+            mocked_xml = fileh.read()
+        mock_getresultstree.return_value = etree.fromstring(mocked_xml)
+
+        myrunner = runner.BeakerRunner(**DEFAULT_ARGS)
+        result = myrunner.jobresult("J:00001")
+        self.assertEqual(result, (1, 'Fail'))
```
agustinhenze: s/assertEqual/assertTupleEqual/

spbnick:
@major, could you please address Agustin's comments before I review this again? Thank you.

agustinhenze:
Ok, I'll keep it in mind for the next reviews! Thank you!

spbnick:
@major, I tried running the tests with your change and got this output:

    ...........WARNING:root:FAILED 1/0 on a single host: A
    WARNING:root:FAILED 2/0 on multiple hosts
    ..............
    ----------------------------------------------------------------------
    Ran 25 tests in 0.021s

    OK

Is that a failure, is it a test artifact, or some unittest quirk?

tests/test_runner.py
major: Fixed.

tests/test_runner.py
major: Fixed.

tests/test_runner.py
major: Fixed.

tests/test_runner.py
major: Fixed.

tests/test_runner.py
major: I have no idea. :)

tests/test_runner.py
major: Fixed.

tests/test_runner.py
major: Fixed.

skt/runner.py
```diff
@@ -148,6 +155,12 @@ def getconsolelog(self, jobid=None):
         return url
 
     def _forget_cid(self, cid):
+        """
+        Remove a job or recipe from the current list
+
+        Args:
+            cid: The job (J:xxxxx) or recipe (R:xxxxx) id
```
spbnick: This will have to do for now, but I think we should use Beaker's [official term](https://www.mankier.com/1/bkr#Description-Specifying_tasks) "taskspec" instead? E.g. "cid: The job or recipe taskspec (universal ID)".

skt/runner.py
```diff
@@ -167,6 +180,15 @@ def _forget_cid(self, cid):
             raise ValueError("Unknown cid type: %s" % cid)
 
     def getverboseresults(self, joblist):
+        """
+        Retrieve verbose results for a list of beaker jobs
+
+        Args:
+            joblist: A list of beaker jobs
```
spbnick: I think this needs to be described more specifically and using official terms. E.g. "joblist: A list of Beaker job taskspecs (universal IDs)." This will have to do for now, though.

skt/runner.py
```diff
@@ -167,6 +180,15 @@ def _forget_cid(self, cid):
...
+
+        Returns:
+            A dictionary with the jobid as a key and job results as the value
```
spbnick: We need to use "taskspec" here as well, I think, but we'll have to merge this as is.

spbnick:
@veruu or @major, could you please merge this after a rebase?

spbnick:
Oh, @major, one thing I missed: it would be really good, if we could avoid tests spewing out garbage while executing, such as that `WARNING:root:FAILED 1/0 on a single host: A` line. If you don't have time for that now, could you please submit an issue to silence that later? Thank you :)

major:
@spbnick Done. I opened #124.

PR #47, opened 2018-03-29
Remove flaky test_parse_valid_patchwork_url

spbnick:
The test_parse_valid_patchwork_url attempts to contact the server in the
URL, which is not guaranteed to always be available, resulting in broken
tests. Removing it for now.

See https://travis-ci.org/RH-FMK/skt/builds/359797306


spbnick:
@major, I'm removing `test_parse_valid_patchwork_url` to fix CI. We'll likely need to refactor the code being tested to be able to write a test for it.

agustinhenze:
You can mock the request to the xmlrpcserver and return the expected result. It'll be fast and deterministic.

spbnick:
@agustinhenze, yeah, that can work.

PR #48, opened 2018-03-29
[executable.py] Code enhancements

veruu:
- make code more pythonic:
  - use `if x` or `if not x` instead of comparisons with 0 or None
  - `dict.get(key)` returns None in case key is not found so we can
    simplify the comparisons there as well
- add initialization of cfgurl - it is created conditionally so we'd get
  error in case the condition doesn't go through.
- get rid of `continue` statements

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -86,7 +86,7 @@ def junit(func):
     """
     def wrapper(cfg):
         global retcode
-        if cfg.get('junit') is not None:
+        if cfg.get('junit'):
             tstart = time.time()
             tc = junit_xml.TestCase(func.__name__, classname="skt")
 
@@ -131,10 +131,10 @@ def cmd_merge(cfg):
 
             utypes.append("[git]")
             idx += 1
-            if retcode != 0:
+            if retcode:
                 return
 
-        if cfg.get('patchlist') is not None:
+        if cfg.get('patchlist'):
             utypes.append("[local patch]")
             idx = 0
             for patch in cfg.get('patchlist'):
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -259,18 +261,18 @@ def cmd_run(cfg):
         save_state(cfg, {'baseretcode': baseres})
 
         # If baseline also fails - assume pass
-        if baseres != 0:
+        if not baseres:
```
spbnick: Should this be inverted?

skt/executable.py
```diff
@@ -558,39 +560,35 @@ def load_config(args):
 
     if config.has_section('config'):
         for (name, value) in config.items('config'):
-            if name not in cfg or cfg.get(name) is None:
+            if not cfg.get(name):
                 cfg[name] = value
 
-    if config.has_section('publisher') and ('publisher' not in cfg or
-                                            cfg.get('publisher') is None):
+    if config.has_section('publisher') and not cfg.get('publisher'):
```
spbnick: Hmm, we'll need better name differentiation than `cfg` vs `config` here.

skt/executable.py
veruu: Crap, good catch!

skt/executable.py
veruu: Agreed, a lot of naming is confusing.

spbnick:
Thanks you, @veruu, this looks good. Could you please rebase this before I merge it?

veruu:
> Could you please rebase this before I merge it?

Done

spbnick:
Thank you, Veronika, merged!

PR #52, opened 2018-04-03
[reporter.py] Code enhancements

veruu:
* reorder imports to make pylint happy
* remove `continue` lines
* use `enumerate()` instead of iterating over `range(len())` (pylint)
* don't use `len()` in comparisons (pylint)
* indentation
* merge together declaration and initialization with result lists
* simplify `endswith()` calls


veruu:
Rebased on current master and changed `[]` for `:` for @spbnick

skt/reporter.py
```diff
@@ -442,11 +430,9 @@ def report(self):
         print self.getreport()
 
         for (name, att) in self.attach:
```
spbnick: Should we also remove parentheses here?

veruu: If I checked the code correctly, in the both cases you mention we use lists of tuples. While in these cases the parens are not necessary for correct value unpacking, I left them there for the clarity that the iteration happens on tuples. So you can automatically see what the variable contains to work with, instead of needing to check the initialization, or having to add a comment there (I'm not a fan of comments where the functionality can be made clear form the code).

I don't feel strong about either way (and `pylint` should be fine with both too, AFAIK), it's just a readability thing since the structures used are a bit complicated.

veruu:
So rebasing on master dismisses reviews, good to know

skt/reporter.py
spbnick: Ah, I see, that's fine by me :)

veruu:
> Yes, I enabled clearing reviews as soon as any new commits are added to the PR (which includes rebases). I think it makes sense, but we can change that if it doesn't work for everyone.

Makes sense, I just didn't know clearing was triggered by this and thought you need to do it manually

PR #55, opened 2018-04-04
Fix debuginfo disabling

spbnick:
* Log the command being executed.
* Refer to the "config" script by its full pathname, otherwise it cannot
  be found when building outside the source directory.
* Pass the "--file" option, pointing to the configuration, otherwise the
  script cannot find it when building outside the source directory.
* Pass "--disable" and "debug_info" as separate arguments, otherwise the
  option is not recognized by the script.

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -490,8 +490,11 @@ class kbuilder(object):
         # of kernel issues on a specific system. This is why debuginfo is
         # disabled by default.
         if not self.enable_debuginfo:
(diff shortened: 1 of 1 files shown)


spbnick:
@major, could you please check this matches what you wanted it to do? I failed to notice the issues, but could you please test the code more thoroughly before submitting? Thanks :)!

spbnick:
Merged, thanks, Major!

PR #56, opened 2018-04-04
README.md: Cleanup examples

spbnick:


spbnick:
@veruu, could you take a quick look at this PR, and see if it's OK? Thanks!

veruu:
Can we fix the placeholders with the reporting as well? Not the generic examples but these:

```$ skt.py --rc <SKTRC> --state --workdir <WORKDIR> -vv report --reporter stdio '{}'```

```$ skt.py --rc <SKTRC> --state --workdir <WORKDIR> -vv report --reporter mail '{"mailfrom": "'$USER'@localhost", "mailto": "'$USER'@localhost, root@localhost"}'```

since they also use specific examples for email addresses and so.

/me has no idea how to comment on the part of file that's not in the diff

**Comment by spbnick** [comment]
Done. I missed these, thank you! I also added a commit fixing newline escaping.

There doesn't seem to be a way to comment on unchanged lines, unfortunately. You can link to lines in specific commits, though. Click on the line number and copy the address. Like this:

https://github.com/RH-FMK/skt/pull/56/files#diff-04c6e90faac2675aa89e2176d2eec7d8R335

File: README.md [hunk 1]
```diff
@@ -164,7 +164,7 @@ E.g. to apply a particular patch to a particular, known-good commit from the
 
 And to build the kernel run:
 
-    $ skt --rc <SKTRC> --state --workdir skt-workdir -vv \
+    $ skt --rc <SKTRC> --state --workdir <WORKDIR> -vv \
```
veruu: This doesn't look like the previous version.

veruu: Ah, right! I'm still getting confused by GitHub's interface (versus very distinct color scheme I use for diffs) so I missed the dividing light blue line and thought this belongs to the above specific example. Ignore this comment then.

spbnick:
Thank you, Veronika, merged!

PR #59, opened 2018-04-05
Functional tests in Circle-CI

### LINKED ISSUE #60: CentOS 7's git is too old for `git am` outside of the tree
When using `git --work-tree=... --git-dir=... am -` on CentOS 7, you get an error:

    fatal: /usr/libexec/git-core/git-am cannot be used without a working tree.

The git version in CentOS 7 is too old for this to work properly and it's been fixed in later versions. We have a few options:

* Use GitPython in skt (probably a good idea)
* `chdir` into the work directory to apply patches (would make paths confusing)
* Upgrade git whenever we use CentOS 7 (sounds awful)

### LINKED ISSUE #62: skt running in a virtualenv cannot run bkr
When skt is installed with pip inside a virtualenv, its execution of `bkr` command fails with something like this:

      File "/home/worker/runner/workspace/skt_test@tmp/virtualenv/bin/bkr", line 7, in <module>
        from bkr.client.main import main
      File "/home/worker/runner/workspace/skt_test@tmp/virtualenv/lib/python2.7/site-packages/bkr/client/main.py", line 16, in <module>
        import krbV

major:
This PR does three things:

* Makes `beaker-client` an optional installation (includes docs) (fixes #62)
* Fixes a CentOS 7 bug when using `git-am` outside the workdir (fixes #60)
* Adds Circle-CI functional testing for CentOS 7/Fedora 27

Includes three commits:

    0ee9041 Functional tests in Circle-CI
    ba24f8f Switch to workdir before `git am`
    55cc4ba Make beaker-client optional

Signed-off-by: Major Hayden <major@mhtx.net>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,63 @@
+defaults: &defaults
+  working_directory: /tmp/skt
+  steps:
+    - checkout
+    - run:
+        name: Install skt using pip
+        command: pip install .
+    - run:
+        name: Run Python linters and tests
+        command: |
+          flake8 --show-source .
+          pylint tests
+          python -m unittest discover tests
+    # TODO(mhayden): Add --depth support to skt and then let skt to the clone
+    # rather than using the git command below.
+    - run:
+        name: Clone and prepare repo
+        command: |
+          mkdir -p /tmp/skt_test/workdir
+          git config --global user.name "SKT"
+          git config --global user.email "noreply@redhat.com"
+          git clone --branch v4.16 --depth 1 git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git /tmp/test/workdir
+    - run:
+        name: Test `skt merge`
+        command: |
+          skt -vv --state -d workdir --rc sktrc merge \
+            --pw https://patchwork.ozlabs.org/patch/895383/ \
+            -b git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git \
+            --ref master
+        working_directory: /tmp/skt_test
+    - run:
+        name: Test `skt build`
+        command: |
+            make -C workdir tinyconfig
+            mv workdir/.config config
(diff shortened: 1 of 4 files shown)


major:
Looks like the average functional test takes 2-4 minutes.

setup.cfg
```diff
@@ -12,8 +12,7 @@ packages = find:
 # Parse the MANIFEST.in file and include those files, too.
 include_package_data = True
 # Let pip install dependencies automatically.
-install_requires = beaker-client
-                   flake8
+install_requires = flake8
```
spbnick: This is a bummer. Do we have to break the pip package dependencies for Circle CI?

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+        name: Run Python linters and tests
+        command: |
+          flake8 --show-source .
+          pylint tests
+          python -m unittest discover tests
+    - run:
+        name: Clone and prepare repo
+        command: |
+          mkdir -p /tmp/skt_test/workdir
+          git clone --depth 1 git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git /tmp/skt_test/workdir
```
spbnick: I agree with @dzickusrh that we need to fix the kernel version with which we are testing. We're testing skt here, not the kernel, after all. And then we don't want patching to break once it stops applying.

Another thing is I think we should be cloning the kernel using skt itself (with the `merge` command). What do you think?

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+          git config --global user.name "SKT"
+          git config --global user.email "noreply@redhat.com"
```
spbnick: This perhaps better be the first two things done with git, before cloning.

major:
@spbnick Also, made an issue to figure out how to clone a stable branch in #64. Hopefully I can get to that today.

PR #44, opened 2018-03-28
Clean up argument/option parsers

major:
This patch does a few things:

* Indents parser functions to make them more readable
* Groups each parser function along with its subcommand (with comments)
* Organizes subcommands alphabetically

Fixes #43.

Signed-off-by: Major Hayden <major@mhtx.net>

PR #65, opened 2018-04-06
Add extra args for `git fetch`

### LINKED ISSUE #63: Need the ability to do a shallow clone
Cloning a repository with `--depth <somenumber>` is much faster than cloning the entire repo, especially when we want to quickly check if a patch merges. `skt` needs support for the `--depth` argument.

Once this is done, the Circle-CI jobs should be adjusted to use `skt` for the cloning instead (see #59).

major:
Full clones of upstream kernel repositories take a long time. Using
shallow clones (--depth 1) can save time when a lengthy history is
not needed.

This patch adds an option for users to specify additional arguments
to `git fetch`.

Documentation is included. Fixes #63.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,20 @@ E.g. to apply a particular patch to a particular, known-good commit from the
                    --ref a870a02cc963de35452bbed932560ed69725c4f2 \
                    --pw https://patchwork.ozlabs.org/patch/886637
 
+#### Faster clones
+
+In some instances, a full git history is not needed. Shallow clones are git
+repositories that have a truncated history and they can be cloned much faster
+than a full copy of the repository. Use the `--fetch-depth` option with
+`skt merge`:
+
+    $ skt ... merge ... --fetch-depth 5
+
+In the example above, the repository will be cloned with a git history of five
+commits. This speeds up the cloning process, but it also prevents the use of
+any git references (tags, branches, etc) that were made before the last five
+commits.
+
 ### Build
 
 And to build the kernel run:
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -128,9 +128,12 @@ def cmd_merge(cfg):
     """
     global retcode
     utypes = []
-    ktree = skt.ktree(cfg.get('baserepo'),
-                      ref=cfg.get('ref'),
-                      wdir=cfg.get('workdir'))
+    ktree = skt.ktree(
+        cfg.get('baserepo'),
+        ref=cfg.get('ref'),
+        wdir=cfg.get('workdir'),
+        fetch_depth=cfg.get('fetch_depth')
+    )
     bhead = ktree.checkout()
     commitdate = ktree.get_commit_date(bhead)
     save_state(cfg, {'baserepo': cfg.get('baserepo'),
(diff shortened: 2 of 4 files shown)


skt/executable.py
```diff
@@ -540,6 +540,15 @@ def setup_parser():
         type=str,
         help="Base repo URL"
     )
+    parser_merge.add_argument(
+        "--extra-git-args",
+        type=int,
+        help=(
+            "extra arguments to pass to `git fetch` "
+            "(ex: --depth 1 for shallow clone)")
+        ,
```
veruu: The last ending parenthesis should be on this line, before the `,`(pycodestyle)

skt/executable.py
```diff
@@ -540,6 +540,15 @@ def setup_parser():
         type=str,
         help="Base repo URL"
     )
+    parser_merge.add_argument(
+        "--extra-git-args",
+        type=int,
```
veruu: The docs suggest passing a string to the option, but here it says `type=int`. Based on what the option should do, I believe the docs are right ;)

veruu:
I'm unable to run most skt commands to test it out, I'm not sure yet if it's a problem in my environment or with the parser patches. Investigating, if I find a culprit in those I'll add a comment on #44

skt/executable.py
major: Ah, good catch. I'll fix that.

skt/executable.py
major: Oops! That was from a previous revision. I'll test that, too.

major:
@spbnick Hah, okay. I was trying to allow for more git arguments without code changes, but I see your point.

spbnick:
Yeah, I understand, and I agree that in some cases this would make perfect
sense, I would reach for this approach too, then.

However, a big part of skt value is wrapping and narrowing git interface to
a few basic operations. Letting users pass options freely there is working
against that, especially since most of what they could pass would break skt.

It would be like adding a second gas pedal to a passenger seat in a taxi
instead of letting passengers tell the driver they're in a hurry.

Hope that makes sense :)

skt/executable.py
```diff
@@ -507,6 +510,15 @@ def setup_parser():
     parser_merge.add_argument("-m", "--merge-ref", nargs="+",
                               help="Merge ref format: 'url [ref]'",
                               action="append")
+    parser_merge.add_argument(
+        "--fetch-depth",
+        type=str,
+        help=(
+            "Create a shallow clone with a history truncated to the specified"
+            "number of commits."
```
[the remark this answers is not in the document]
major: Very good catch. Fixed.

README.md
```diff
@@ -165,6 +165,20 @@ E.g. to apply a particular patch to a particular, known-good commit from the
...
+#### Faster clones
+
+In some instances, a full git history is not needed. Shallow clones are git
+repositories that have a truncated history and they can be cloned much faster
+than a full copy of the repository. Use the `--fetch-depth` option with
+`skt merge`:
+
+    $ skt ... merge ... --fetch-depth 1
+
+In the example above, the repository will be cloned with a git history of five
```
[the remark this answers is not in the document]
major: Fixed.

spbnick:
Thanks a lot, Major, merged!

PR #67, opened 2018-04-06
Simplify ktree.__init__ and add tests

major:
This patch adds class modules for some parts of `ktree.__init__` to make it easier to test. It also includes additional tests for `ktree.__init__` and some other `ktree` methods.


skt/__init__.py
```diff
@@ -137,41 +137,62 @@ def __init__(self, uri, ref=None, wdir=None):
...
 
-        logging.info("base repo url: %s", self.uri)
-        logging.info("base ref: %s", self.ref)
-        logging.info("work dir: %s", self.wdir)
+    def get_workdir(workdir):
+        """Determine the full path to the work directory"""
+        if workdir:
+            return os.path.expanduser(workdir)
+        else:
+            return tempfile.mkdtemp()
```
spbnick: This makes a "get" function non-idempotent, i.e. its result changes with every call. Please rename or refactor the function.

skt/__init__.py
```diff
@@ -137,41 +137,62 @@ def __init__(self, uri, ref=None, wdir=None):
...
+
+    def get_gitdir(workdir):
+        """Get the path to the .git directory within the work directory"""
+        return "%s/.git" % workdir
+
+    def create_workdir():
+        """Create the work directory if it does not exist"""
+        if not os.path.is_dir(self.wdir):
+            os.mkdir(self.wdir)
```
spbnick: I'm not sure we need a public function for creating a workdir. I doubt the class user would want to call it.

Also, please just run `os.mkdir` and then handle the "already exists" exception, instead of checking, then creating. This way we would avoid a race condition.

skt/__init__.py
```diff
@@ -137,41 +137,62 @@ def __init__(self, uri, ref=None, wdir=None):
...
+
+    def remove_mergelog():
+        """Remove the old merge.log if it exists"""
+        if os.path.isfile(self.mergelog):
+            os.unlink(mergelog)
```
spbnick: I'm also not sure this needs to be public, or be a function at all.

Also, same here, just call `os.unlink` and then handle "doesn't exist".

tests/test_init.py
```diff
@@ -53,3 +57,60 @@ def test_parse_bad_patchwork_url(self):
...
+        workdir = tempfile.mkdtemp()
+        merge_log = "{}/merge.log".format(workdir)
+        with open(merge_log, 'w') as fileh:
+            fileh.write("merge log sample content")
+
+        result = skt.ktree(GIT_URI, wdir=workdir)
+        self.assertIsInstance(result, skt.ktree)
+        self.assertEqual(result.uri, GIT_URI)
+        self.assertEqual(result.wdir, workdir)
+        self.assertFalse(os.path.isfile(merge_log))
```
spbnick: I think, first of all, these two functions shouldn't be a part of the public interface (or exist at all). Then they're so trivial in purpose, they're not likely to fail and so should have tests for them written last, if at all.

tests/test_init.py
```diff
@@ -53,3 +57,60 @@ def test_parse_bad_patchwork_url(self):
...
+        self.assertEqual(result.wdir, workdir)
+
+    def test_ktree_create_workdir(self):
+        """Ensure create_workdir() creates a work dir if it does not exist"""
+        workdir = tempfile.mkdtemp()
+        os.rmdir(workdir)
+        result = skt.ktree(GIT_URI, wdir=workdir)
+        self.assertIsInstance(result, skt.ktree)
+        self.assertEqual(result.uri, GIT_URI)
+        self.assertEqual(result.wdir, workdir)
```
spbnick: Please do not test everything again and again in every test. If you have to test something, then have that as an explicit separate test. Make them independent and orthogonal. Otherwise you will have all these tests fail when one thing they all repeat fails, and that's not helpful.

skt/__init__.py
```diff
@@ -154,45 +154,46 @@ def __init__(self, uri, ref=None, wdir=None):
...
-    def set_repo_remote():
+    def set_repo_remote(self):
         """Set the origin remote for the repository"""
         try:
             self.git_cmd("remote", "set-url", "origin", self.uri)
         except subprocess.CalledProcessError:
             self.git_cmd("remote", "add", "origin", self.uri)
 
-    def get_workdir(workdir):
+    def get_workdir(self, workdir):
```
spbnick: These changes warrant a separate commit.

skt/__init__.py
```diff
@@ -152,27 +151,35 @@ def __init__(self, uri, ref=None, wdir=None):
...
+        logging.info("work dir: %s", self.wdir)
+
+    def initialize_repo(self):
+        """Initialize the git repository"""
+        git_directory = "{}/.git".format(self.wdir)
+        if not os.path.isdir(git_directory):
+            self.git_cmd("init")
+
+    def set_repo_remote(self):
+        """Set the origin remote for the repository"""
```
spbnick: I appreciate that you're trying to make the `__init__` function easier to read. However, I think that benefit here is marginal, if not negative. Each function (`initialize_repo` and `set_repo_remote`) is just a few lines long and line-count overhead is bigger than the functions amount to. The code in functions depends on certain variables being initialized already, and that's not obvious at the place of their invocation, which can lead to issues after code is reordered. That could be fixed by making them accept the values in question as arguments, but then they're so short, that would add significant syntax overhead. Better effect can be achieved by just adding comments to the original code, I think. Lastly, I don't know what exact conventions in Python are, but I would make the functions private, if I were adding them.

Do you think you can leave the code in place and just add comments explaining what it does?

tests/test_init.py
```diff
@@ -53,3 +57,54 @@ def test_parse_bad_patchwork_url(self):
...
+
+    def test_ktree_create_workdir(self):
+        """Ensure create_workdir() creates a work dir if it does not exist"""
+        workdir = tempfile.mkdtemp()
+        os.rmdir(workdir)
+        skt.ktree(GIT_URI, wdir=workdir)
+        self.assertTrue(os.path.isdir(workdir))
+
+    def test_ktree_remove_mergelog(self):
+        """Ensure remove_mergelog really removes the merge.log"""
```
spbnick: This and the above docstrings need update as the functions don't exist anymore.

tests/test_init.py
```diff
@@ -53,3 +57,54 @@ def test_parse_bad_patchwork_url(self):
             "Can't parse patchwork url: '{}'".format(patchwork_url)
             in context.exception
         )
+
+    def test_ktree_init(self):
+        """Test ktree.__init__ without specifying a work directory"""
```
spbnick: This doesn't say what exactly is being tested. Did you mean "Test ktree can be created without specifying a work directory"?

tests/test_init.py
```diff
@@ -53,3 +57,54 @@ def test_parse_bad_patchwork_url(self):
...
+        result = skt.ktree(GIT_URI)
+        self.assertIsInstance(result, skt.ktree)
+        self.assertEqual(result.uri, GIT_URI)
+
+    def test_ktree_init_workdir(self):
+        """Test ktree.__init__ with specifying a work directory"""
```
spbnick: Same here.

tests/test_init.py
```diff
@@ -53,3 +57,54 @@ def test_parse_bad_patchwork_url(self):
...
+        workdir = tempfile.mkdtemp()
+        ktree_obj = skt.ktree(GIT_URI, wdir=workdir)
+        ktree_obj.info = [["test", "result"]]
+        ktree_obj.dumpinfo("test.csv")
+        self.assertTrue(os.path.isfile("{}/test.csv".format(workdir)))
+
+        with open("{}/test.csv".format(workdir), 'r') as fileh:
+            csv_content = fileh.read()
+
+        self.assertEqual(csv_content, "test,result\n")
```
spbnick: I would say since the contents of the CSV file is highly fluid at the moment (depends on the order of method calls), and is not documented, we shouldn't really check it. Check for non-zero size perhaps?

PR #70, opened 2018-04-06
Refactor patchwork patch retrieval

### LINKED ISSUE #69: Avoid using XMLRPC
XMLRPC leads to a lot of headaches and requires lots of wrapper code. All of the data we need about a particular patch is held within the mbox file of the patch itself. We should just download the mbox file and parse it.

major:
XMLRPC is a tricky protocol and it requires a lot of wrapper code
to avoid various problems. However, all of the information we need
for the patch itself is included in the mbox file that is downloaded
from Patchwork.

This patch removes the XMLRPC code and simply downloads the mbox file.
That file can then be parsed to get various information about the
patch itself.

This allows us to delete some code! Fixes #69.


spbnick:
@major, before I review this, could you please consult https://github.com/RH-FMK/sktm/blob/master/sktm/patchwork.py and see if we can do something similar there. Please also consider RH-FMK/meta#1 (we can discuss there as well), and how it would affect the code here. Please write what you think!

major:
@spbnick So the functionality that `skt` needs it pretty limited (get an mbox file and apply it). `sktm` needs a lot more functionality, like parsing through series and cover letters. A shared library would make some sense.

I looked at https://github.com/getpatchwork/git-pw but it only supports patchwork 2.0+. I guess that's a question for us -- do we need support for patchwork 1 and 2?

major:
Also, siesta could be a good option to reduce the code we need to write: http://scastillo.sh/siesta/

spbnick:
Yes, we'll need support for both Patchwork v1 and v2 for the time being. Siesta looks nice, but we can't really use it as it's not in RHEL.

Now, thinking a little more about this, it would be great if skt didn't have to know anything about Patchwork at all and operated purely on patches (plus a little extra info) instead. Therefore, this can be a good change. Maybe we'll even avoid creating a library shared between sktm and skt.

A further change can be removing the `--pw` option and simply accepting mbox URLs, each containing a single patch. Then it would be the caller's job to make sure the URLs point to mailboxes (and not to patches) and they have the right amount of slashes for Patchwork to respond properly.

I'll go ahead and review this.

veruu:
> Yes, we'll need support for both Patchwork v1 and v2 for the time being. Siesta looks nice, but we can't really use it as it's not in RHEL.

Is this really a problem since it's available via pip?

spbnick:
@veruu, I'm not really sure. Can you check with Marcela? That would be good to know.

Otherwise, I would like to keep our dependencies to a minimum in general, if possible.

spbnick:
I will be OK with Siesta, if it really saves us a good deal of code, is reasonable, reliable, and we can have it in RHEL.

spbnick:
@veruu, yes, that would be nice, thanks :) The question is if there's any policy preventing this.

tests/test_init.py
```diff
@@ -16,40 +16,33 @@
...
-        """Ensure parse_patchwork_url() handles a parsing exception"""
-        patchwork_url = "garbage"
-        with self.assertRaises(Exception) as context:
-            skt.parse_patchwork_url(patchwork_url)
-
-        self.assertTrue(
-            "Can't parse patchwork url: '{}'".format(patchwork_url)
-            in context.exception
-        )
+    @mock.patch('requests.get', side_effect=mocked_requests_get)
```
agustinhenze: I don't know if you are thinking on writing more complicated test cases using `mocked_requests_get` but here with the following code will be enough:
```pythonI don't think it is worse, but I am not convinced it is better either hehe. I'm used to seeing comparisons using mock.call, maybe that's why it sounds more intuitive to me. Anyway I want to see this PR merged soon so I'll take the way you prefer and if in the future we find a better way to solve this, we can change it :):
    @mock.patch('requests.get', return_value=mock.Mock) 
    def test_get_patchwork_mbox(self, mock_get):
        mock_get().text = 'mbox text goes here'
```
I've tested it locally and it works ok
```diff
@@ -39,10 +39,11 @@ def mocked_requests_get(*args):
 class TestInit(unittest.TestCase):
     """Test cases for skt's __init__.py"""
 
-    @mock.patch('requests.get', side_effect=mocked_requests_get)
+    @mock.patch('requests.get', return_value=mock.Mock)
     def test_get_patchwork_mbox(self, mock_get):
-        # pylint: disable=unused-argument
         """Ensure get_patchwork_mbox() can get an mbox file"""
+        expected_text = 'mbox text goes here'
+        mock_get().text = expected_text
         patchwork_uri = 'https://patchwork.kernel.org/patch/10326957/'
         result = skt.get_patchwork_mbox(patchwork_uri)
-        self.assertEqual(result, 'mbox text goes here')
+        self.assertEqual(result, expected_text)
```

spbnick:
@veruu, right, the problem with pip is that we need to have an RPM package of this eventually, when RPM package installs it should have all of its dependencies installed already, and we can't install anything with pip as part of RPM install. So, we might need to replace junit_xml with our own code in the long term, which shouldn't be that hard for our purposes.

PR #73, opened 2018-04-10
Add a restraint for pycodestyle

### LINKED ISSUE #72: pycodestyle 2.4.0 is broken
We need a constraint that causes CI to use a working version of pycodestyle until the issue is fixed.

major:
pycodestyle 2.4.0 has a bug that causes flake8 to fail. This patch
adds a constraint to avoid this version.

Fixes #72.


spbnick:
I assume we won't be able to remove the constraint as the pycodestyle will have to have a new release with the fix?

PR #80, opened 2018-04-19
tests: Clarify called kbuilder.mktgz is silent

spbnick:
The single-character difference between self.kbuilder_mktgz and
self.kbuilder.mktgz is easy to miss. Plus the function name doesn't say
what the difference is. Rename self.kbuilder_mktgz to
self.kbuilder_mktgz_silent to make that clear. Switch the function
description to imperative mood while at it.


spbnick:
@veruu, can you give me an approval, as @agustinhenze doesn't have the rights yet. I wouldn't mind a review too, if you have the time ;) Thanks!

PR #83, opened 2018-04-21
Add pytest to travis and README

danghai:
Hello, 
I think Pytest is a good framewor to support complex functional testing for applications and libraries. I add the Pytest to travis, and README in this PR. Could you review it?
Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,6 @@ script:
     - flake8 --show-source .
     - pylint tests
     - python -m unittest discover tests
+    - python -m pytest tests
     # Test the uninstallation of skt,
     - pip uninstall -y skt
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ For running some specific tests you can do this as following:
 
     $ python -m unittest tests.test_publisher
 
+For running all tests by pytest:
+
+    $ python -m pytest tests
+
 
 Installation
 ------------


veruu:
Hi, thanks for the pull request! However, as it currently stands, the only thing it does is re-running the tests we already have which doesn't really add any value. I currently see two options to improve it:

a) remove the `unittest` run from travis (and fix it in `circleci`, remove from `README`, etc too): I do agree that `pytest` is more advanced than `unittest`, has a more useful output.... Since it can handle `unittest` tests we can keep the ones we already have and continue working on them as needed, while utilizing the nice things from `pytest`.

b) rewrite the current test suite to `pytest`: We currently don't have any plans to do it but since you submitted this pull, maybe you do?

Let me know which route makes more sense to you. Also, calling in @RH-FMK/mantainers in case somebody sees a flaw with the proposals.

Thanks again!

spbnick:
I would prefer to keep things as simple as possible (but no simpler). If we can do well with unittest, and pytest doesn't provide any considerable benefits, *and* if unittest is simpler, I would stick with unittest.

danghai:
@veruu : I am willing to rewrite the current test suite to `pytest`. However, pytest does not provide any benefits in this case. I closed this PR, and will reopen if the team consider to switch `pytest` for sure

PR #85, opened 2018-04-23
Add build output directory support

### LINKED ISSUE #84: Need capability to build kernel outside source directory
When running `skt` in a container, it would be helpful to build kernel artifacts outside the workdir where the kernel source is cloned. This would allow for artifacts to be sent to other systems.

major:
This patch allows skt to build kernel artifacts outside the kernel
source directory. This is helpful for containerized environments
where disk space is limited and artifacts must be shipped to other
systems.

Fixes #84.


spbnick:
Thank you, Major. I appreciate your effort for making skt transparently pick up the build output directory, it has its perks. However, I think in this case Stef's approach is more robust.

Parsing the make command line would be very difficult to do reliably in all cases. Starting with the fact that relative paths passed to `O=` are resolved relative to the kernel source directory, continuing to the cases where there can be multiple assignments on the command line, and finally the fact that make supports full assignment syntax on the command line, such as `O+=`, `O?=`, etc., and can also refer to other variables (including those defined in the Makefile) from them. We don't want to handle all that.

So do you think you can pick up Stef's PR or simply do your version with a dedicated command-line option, which would add `O=` to the end of the make's command line?

major:
@spbnick Your recommendations make sense. I'll try to pick this up today.

skt/__init__.py
```diff
@@ -476,16 +477,36 @@ def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
...
+            # Ensure our output directory is an absolute path
+            if not os.path.isabs(self.kbuild_output_dir):
+                self.kbuild_output_dir = "{}/{}".format(
+                    os.getcwd(),
+                    self.kbuild_output_dir
+                )
+
+            # Make the directory if it does not exist already
+            if not os.path.isdir(self.kbuild_output_dir):
+                os.makedirs(self.kbuild_output_dir)
```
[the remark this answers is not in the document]
major: Fixed.

skt/__init__.py
```diff
@@ -476,16 +477,36 @@ def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
...
+            logging.info(
+                "Found kernel output directory specifier in make options: "
+                "{}".format(self.kbuild_output_dir)
+            )
+            # Ensure our output directory is an absolute path
+            if not os.path.isabs(self.kbuild_output_dir):
+                self.kbuild_output_dir = "{}/{}".format(
+                    os.getcwd(),
+                    self.kbuild_output_dir
+                )
```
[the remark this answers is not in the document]
major: Well, if a user runs `skt build --kbuild-output output_directory`, wouldn't `output_directory` be expected to be in the current working directory? 

Putting it in the `workdir` somewhere defeats the point of building outside the tree. Within containers, we need the workdir to be on /tmp but have the build output be somewhere else, like `/opt/nfs/build-name`.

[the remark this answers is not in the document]
spbnick: Sorry, this comment is only about paths supplied via `O=` and I should've discarded it before submitting the review. Sure, you're right, the command-line option value should be relative to the current directory.

major:
Just rebased the patch.

skt/__init__.py
```diff
@@ -513,22 +544,26 @@ def prepare(self, clean=True):
                     "--file", self.get_cfgpath(),
                     "--disable", "debug_info"]
             logging.info("disabling debuginfo: %s", args)
+            print args
```
veruu: ?

skt/__init__.py
```diff
@@ -476,16 +477,38 @@ def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
...
+            logging.info(
+                "Found kernel build output directory specifier in make "
+                "options: {}".format(self.kbuild_output_dir)
+            )
+
+            # Ensure the output directory exists
+            # NOTE(mhayden): os.makedirs will cause an error if the directory
+            # already exists, so we must check for the directory first.
+            if not os.path.isdir(self.kbuild_output_dir):
+                os.makedirs(self.kbuild_output_dir)
```
dzickusrh: Is this a problem?  If the directory already exists, does that mean another container is using it?  Is it clean?  An existing directory on shared storage makes me wonder.

skt/__init__.py
```diff
@@ -497,11 +520,19 @@ def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
 
     def prepare(self, clean=True):
         if (clean):
-            args = self.defmakeargs + ["mrproper"]
+            args = self.makeopts + ["mrproper"]
```
dzickusrh: Do we need 'mrproper' any more if we have a fresh build area.  I think this was a legacy use when the same output build dir was being used.

skt/__init__.py
```diff
@@ -476,16 +477,38 @@ def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
...
+            # Ensure the output directory exists
+            # NOTE(mhayden): os.makedirs will cause an error if the directory
+            # already exists, so we must check for the directory first.
+            if not os.path.isdir(self.kbuild_output_dir):
+                os.makedirs(self.kbuild_output_dir)
+
+            # Add the 'O=' make option to specify the output directory
+            self.makeopts.append("O={}".format(self.kbuild_output_dir))
+        else:
+            self.kbuild_output_dir = None
```
dzickusrh: Should we default this to self.path to avoid more if/else clauses below?

skt/executable.py
```diff
@@ -534,6 +535,11 @@ def setup_parser():
         default=False,
         help="Build kernel with debuginfo (default: disabled)"
     )
+    parser_build.add_argument(
+        "--kbuild-output",
```
dzickusrh: The argument name is fine, I just wonder if we could shorten it to '-o' or is that to confusing?

skt/__init__.py
```diff
@@ -497,11 +520,19 @@ def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
...
-            args = self.defmakeargs + ["mrproper"]
+            args = self.makeopts + ["mrproper"]
             logging.info("cleaning up tree: %s", args)
             subprocess.check_call(args)
 
-        shutil.copyfile(self.basecfg, "%s/.config" % self.path)
+        if self.kbuild_output_dir:
+            config_destination = "%s/.config" % self.kbuild_output_dir
+        else:
+            config_destination = "%s/.config" % self.path
```
dzickusrh: Any reason why we are not using below's get_cfgpath()?

skt/__init__.py
major: It catches a situation where the directory might exist because someone made it manually or mounted a filesystem on that directory.

skt/__init__.py
major: The `make mrproper` is required since `make rh-configs` and `make olddefconfig` will make the kernel source unclean. This will cause the `make targz-pkg` to fail with an error that says you must run `make mrproper` first.

skt/executable.py
major: I was just trying to be specific and verbose there. :)

skt/__init__.py
major: Oh good point - I missed that. That would simplify things.

skt/__init__.py
dzickusrh: That was my point.  Is that going to lead to collision with other work?

skt/__init__.py
dzickusrh: really?  That seems odd.  I just ran those commands on rhel-7 with no problem.  Is this leftover from output directory re-use?

major: That could be the case. However, I know if you put the `.config` into the kernel source directory, `make olddefconfig`, and then copy the `.config` to the output directory, `make` will complain at you for not running `make mrproper`. The build will fail.

This may only show up when building kernels with the output outside of the source directory.

skt/__init__.py
major: This is an implementation issue. ;) One would hope that the user would ensure that output directories are not shared.

Within OpenShift, the kernel source is kept inside an ephemeral volume within the container and the build output goes into storage with a guaranteed size allotment that persists after the container is gone.

skt/__init__.py
major: Well, in the single work directory model, `self.path` points to a work directory that contains the kernel source and eventually the kernel build output. However, with the kernel build output separate from the source, you end up with:

* `self.path`: Contains kernel source
* `self.kbuild_output_dir`: Contains build output

Setting `self.path` to the kernel build output directory would break things since skt expects `self.path` to contain the kernel source repository.

dzickusrh: You have it backwards.
else:
    self.kbuild_output_dir = self.path

:-)  Look at your various if/else below, they all have if self.kbuild_output_dir then <use it> else <use self.path>.  By assigning self.kbuild_output_dir to self.path in the absence of --kbuild-output, simplifies the code and makes it easier to read. :-)

skt/__init__.py
dzickusrh: Oh right, because the .config file is there or even worse all the include/autoconf stuff is there.  Yes the O= is the problem and is caused by the 'make rh-configs' cluttering up the local env.  
Ok, so I understand the problem now.  This is fine. I should probably file a separate issue to either comment that fact or fix it by using 'make O=<output> rh-configs'.  I just verified the latter works.

skt/__init__.py
```diff
@@ -479,16 +480,38 @@ def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
...
         self.enable_debuginfo = enable_debuginfo
 
         if makeopts is not None:
-            # FIXME: Might want something a bit smarter here, something that
-            # would parse it the same way bash does
-            self.makeopts = makeopts.split(' ')
-            self.defmakeargs += self.makeopts
+            self.makeopts = self.defmakeargs + shlex.split(makeopts)
+        else:
+            self.makeopts = self.defmakeargs
```
spbnick: This is getting too confusing due to several reasons. First, we assign `self.defmakeargs` and then never use it outside the constructor. Then we have the `makeopts` argument, and we have `self.makeopts`, but they're not the same. The argument is named `makeopts`, but it's not just "options", it's any arguments. Lastly, `self.makeopts` is not just "options", neither it's just "arguments", because it also contains `argv[0]`, the executable name (`"make"`), which, strictly speaking, is not an argument, as in most programs don't process it. Same goes to naming them `args` throughout the code, because those also include `argv[0]`.

With that in mind, how about this:
* rename the constructor argument to `make_args` (or similar, or perhaps even `extra_make_args`) and submit an issue to rename the command-line option accordingly later;
* remove `self.defmakeargs`;
* use `self.make_argv` (or similar, or even `self.make_argv_base`, or `self.make_argv_pfx`) to signify that these are more than just arguments (in the widely-accepted sense);
* use `argv` instead of `args` when building the make's argv array before execution.

I think it is better to name things as close as possible to their contents to minimize the cognitive load and confusion when reading the code, and that's what I'm targeting above.

PR #87, opened 2018-04-24
reporter: job result reporting enhancements

### LINKED ISSUE #76: The `report` command aborts if `run` command was interrupted
If a `run` command is interrupted after the Beaker job is submitted, but before it starts running (i.e. while it is still "queued"), the `report` command running afterwards aborts with the following in the (`-vv`) output:

    2018-04-12 12:09:46,314     INFO   runner type: beaker
    2018-04-12 12:09:46,314     INFO   beaker template: beakerjob.xml
      File "/home/worker/runner/workspace/rhel-7.5@2@tmp/virtualenv/bin/skt", line 11, in <module>
        load_entry_point('skt==1', 'console_scripts', 'skt')()
      File "/home/worker/runner/workspace/rhel-7.5@2@tmp/virtualenv/lib/python2.7/site-packages/skt/executable.py", line 710, in main
        args.func(cfg)
      File "/home/worker/runner/workspace/rhel-7.5@2@tmp/virtualenv/lib/python2.7/site-packages/skt/executable.py", line 332, in cmd_report
        reporter.report()
      File "/home/worker/runner/workspace/rhel-7.5@2@tmp/virtualenv/lib/python2.7/site-packages/skt/reporter.py", line 465, in report
        msg.attach(MIMEText(self.getreport()))
      File "/home/worker/runner/workspace/rhel-7.5@2@tmp/virtualenv/lib/python2.7/site-packages/skt/reporter.py", line 394, in getreport
        msg += self.getjobresults()
      File "/home/worker/runner/workspace/rhel-7.5@2@tmp/virtualenv/lib/python2.7/site-packages/skt/reporter.py", line 333, in getjobresults
        vresults = runner.getverboseresults(list(self.cfg.get("jobs")))
    TypeError: 'NoneType' object is not iterable

Instead it should proceed and report appropriate results.

veruu:
Enhancements related to reporter.getjobresults():
* Console log parsing
  * Move fetchdata() to init for ease of use
  * Refactor to make it clear what's going on
* More friendly wording and cleaner dividing
* Fix #76 - add default for
  `getverboseresults(list(self.cfg.get("jobs", [])))` to ensure we don't
  get a trace call if there are no jobs to report
* Attach console log for every failed job, not only the first one
* Only report failures for the kernel that is actually being tested. We
  don't want to send out reports if the panic occurred with the original
  (not our) kernel, which we did so far.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -332,11 +332,12 @@ class reporter(object):
     def getjobresults(self):
         result = ["\n-----------------------"]
         runner = skt.runner.getrunner(*self.cfg.get("runner"))
-        vresults = runner.getverboseresults(list(self.cfg.get("jobs")))
+        job_list = sorted(list(self.cfg.get("jobs", [])))
+        vresults = runner.getverboseresults(job_list)
 
         minfo = {"short": {}, "long": {}}
         jidx = 1
-        for jobid in sorted(self.cfg.get("jobs")):
+        for jobid in job_list:
             for (recipe, rdata) in vresults[jobid].iteritems():
                 if recipe == "result":
                     continue
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -330,6 +330,20 @@ class reporter(object):
         return result
 
     def getjobresults(self):
+        """
+        Retrieve job results which should be appended to the report.
+
+        Get job results from runner, check console logs (if present) to filter
+        out infrastructure issues and find call traces. For each test run, add
+            1. Number of test run
+            2. It's result
+            3. If present, info about the machine the test ran on
+        If the testing failed, add first found trace call and attach related
+        console log.
+
+        Returns:
+            A list of lines representing results of test runs.
+        """
         result = ['\n' + '=' * 70 + '\n']
         runner = skt.runner.getrunner(*self.cfg.get("runner"))
         job_list = sorted(list(self.cfg.get("jobs", [])))
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -344,24 +344,28 @@ class reporter(object):
 
                 (res, system, clogurl, slshwurl, llshwurl) = rdata
 
+                clog = consolelog(self.cfg.get("krelease"), clogurl)
+                if not clog.data and res != 'Pass':
+                    # The targeted kernel either didn't start booting or the
+                    # console wasn't logged. The second one isn't an issue if
+                    # everything went well, however reporting a failure without
+                    # any details is useless so skip it.
+                    continue
+
                 result.append("Test Run: #%d" % jidx)
                 result.append("result: %s" % res)
 
-                if clogurl is not None and res != "Pass":
-                    logging.info("Panic detected in recipe %s, "
-                                 "attaching console log",
-                                 recipe)
-                    clog = consolelog(self.cfg.get("krelease"), clogurl)
+                if res != "Pass":
+                    logging.info("Failure detected in recipe %s, attaching "
+                                 "console log", recipe)
                     ctraces = clog.gettraces()
                     if ctraces:
                         result.append("first encountered call trace:")
                         result.append(ctraces[0])
 
-                    if jidx == 1:
-                        clfname = "%02d_console.log.gz" % jidx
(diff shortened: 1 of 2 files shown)


skt/reporter.py
```diff
@@ -328,9 +330,9 @@ def getbuildfailure(self):
         return result
 
     def getjobresults(self):
-        result = ["\n-----------------------"]
+        result = ['\n' + '=' * 70 + '\n']
```
spbnick: I like how Perl has a separate operator for this. Operator overloading should be used carefully.

skt/reporter.py
```diff
@@ -343,24 +343,28 @@ def getjobresults(self):
```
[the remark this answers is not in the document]
spbnick: Comment for the whole commit:

This commit contains at least two logical changes: fixing extracting wrong console logs, and adding extracting console logs from other runs. Following the rule "one logical change per commit" this one should be at least two commits. That way, if e.g. it turns out that we don't need logs from all runs after all, it would be easy to revert it, or e.g. if we had to fix wrong console logs in production urgently, it would be easy to cherry-pick it there.

skt/reporter.py
```diff
@@ -117,30 +117,36 @@ def __init__(self, kver, url):
         """
```
spbnick: Commenting in the code, but this should be a comment to the commit as a whole, which GitHub doesn't allow at this moment, unfortunately.

This commit has several logical changes, as far as I can see:
* Fixing handling kernels not booting at all
* Switching to explicit fetching, upon creation (along with test update)
* Handling the missing URL
* Refactoring line extraction

Those would be better as separate commits for simpler reverting and cherry-picking.

skt/reporter.py
```diff
@@ -332,11 +332,11 @@ def getbuildfailure(self):
     def getjobresults(self):
         result = ["\n-----------------------"]
         runner = skt.runner.getrunner(*self.cfg.get("runner"))
-        vresults = runner.getverboseresults(list(self.cfg.get("jobs")))
+        vresults = runner.getverboseresults(list(self.cfg.get("jobs", [])))
 
         minfo = {"short": {}, "long": {}}
         jidx = 1
-        for jobid in sorted(self.cfg.get("jobs")):
+        for jobid in sorted(self.cfg.get("jobs", [])):
```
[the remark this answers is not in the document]
veruu: Do you want this as additional commit since the change it is unrelated to anything else?

skt/reporter.py
veruu: > - Fixing handling kernels not booting at all

It did the same thing before, I just added the comment to explicitly state it to avoid confusion.

> - Handling the missing URL

The missing URL was handled on the reporter side before, I just moved it here because it makes more sense (and in the followup commit, removed the URL check from reporter so it doesn't break anything).

I didn't really want to separate fetching upon extraction and it's refactoring since they are related and I wanted to make sure each added commit works separately.

spbnick: I agree, each commit should work separately, thank you for caring about that. Pity none of the CI solutions available on GitHub support verifying every commit in a PR, only the last.

I would like to understand your intention for the way you split the changes into commits better, and share my view on this, but I'm afraid I'll stray too far and delay us. This looks already better to me and I think it's good for the start. Let's return to the topic in the next PR :)

skt/reporter.py
```diff
@@ -343,24 +343,28 @@ def getjobresults(self):
...
                         result.append(ctraces[0])
 
-                    if jidx == 1:
-                        clfname = "%02d_console.log.gz" % jidx
-                        result.append("full console log attached: %s"
-                                      % clfname)
-                        self.attach.append((clfname, clog.getfulllog()))
+                    clfname = "%02d_console.log.gz" % jidx
+                    result.append("full console log attached: %s" % clfname)
+                    self.attach.append((clfname, clog.getfulllog()))
```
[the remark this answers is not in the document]
spbnick: Ah, I see you already looked at this and thought it through. I think this makes sense.

veruu:
Already working on those changes, will push them soon and ping you.

veruu:
@spbnick ready

Issue #91, opened 2018-04-25
User should be able to override CPU count for kernel builds


major:
At the moment, `skt` gets a count of CPUs on the system and uses that amount for kernel builds. However, there are some situations where this number needs to be increased or decreased. Users should be able to override this variable if needed.

PR #92, opened 2018-04-25
Allow user to specify make job count

major:
Users need the ability to increase or decrease the amount of CPU
cores used for kernel builds. This patch adds a `--jobcount` option
for `skt build` that allows a user to specify the amount of CPUs to
use. If a job count is not specified, `skt` will use the number of
CPUs installed in the system as a default.

Fixes: #91


major:
@spbnick Having two `-j` options in the make output might lead to confusion. I wasn't aware of which value would take precedence when the `make` command runs.

PR #93, opened 2018-04-25
Fix bug: skt merge creates relative workdir under workdir

### LINKED ISSUE #51: skt merge creates relative workdir under workdir
After executing the following command right outside the workdir (`skt-workdir`):

              --pw https://patchwork.ozlabs.org/patch/886637

The workdir has `skt-workdir` directory created under it.

danghai:
Hello, 
I create the PR to fix #51. 
I think `skt-workdir` in this issue is an untracked directory. Introducing the `git clean -df` is a convenience method for deleting untracked file in a repo's working tree. Please review my PR and give the feedback. Thanks!

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -295,6 +295,7 @@ class ktree(object):
         logging.info("checking out %s", self.ref)
         self.git_cmd("checkout", "-q", "--detach", dstref)
         self.git_cmd("reset", "--hard", dstref)
+        self.git_cmd("clean", "-df")
 
         head = self.get_commit()
         self.info.append(("base", self.uri, head))


veruu:
Hi! While this patch clears the symptom by deleting the directory, it doesn't solve the underlying problem which is that the directory is created in the first place. I'm afraid this solution might hide deeper related issues in the future. Do you think you can find the real cause here and fix it?

danghai:
I think the reason is `self.git_cmd("reset", "--hard", dstref)`  creating `untracked` files.

danghai:
I just rebase them and consider a different approach.

veruu:
I'm closing this since the source of original problem was solved by proper path expansion.

PR #95, opened 2018-04-25
Optimize configuration and split out state data

### LINKED ISSUE #27: Separate configuration and state
At the moment skt workflow state is stored in the configuration file, which makes its reuse difficult and error-prone. Move the state somewhere else.

### LINKED ISSUE #89: Relative paths to jobtemplate xml should work
If a user specifies `jobtemplate.xml` as the path to their Beaker XML template, we should look for that in the current working directory.

### LINKED ISSUE #94: Path expansion should occur as early as possible
When a user runs `skt`, the path expansion should happen as early as possible. This expansion includes expanding home directory paths (`~/some_dir/some_file.txt`) or relative paths into absolute paths. This prevents situations where some parts of `skt` might handle paths in different ways than other parts.

major:
This PR does a few things:

* Splits configuration and state information into separate files
* Gets absolute paths to directories and files very early in the `skt` run
* Updates Circle-CI to use the new state file

Fixes #27.
Fixes #89.
Fixes #94.


skt/__init__.py
```diff
@@ -468,12 +467,8 @@ def bisect_iter(self, bad):
...
     def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
                  enable_debuginfo=False):
-        # FIXME Move expansion up the call stack, as this limits the class
-        # usefulness, because tilde is a valid path character.
-        self.path = os.path.expanduser(path)
-        # FIXME Move expansion up the call stack, as this limits the class
-        # usefulness, because tilde is a valid path character.
-        self.basecfg = os.path.expanduser(basecfg)
+        self.path = path
+        self.basecfg = basecfg
```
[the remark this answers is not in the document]
major: Those are fixed up in 83862e8 and d02ba69.

major:
> It would be easier to split this up in to 2-3 patches. First patch brings in framework, second patch wires it up and third patch does some cleanup? Half of the patch is just a straight cfg->state rename.

@dzickusrh I split them up a bit more to make the diffs smaller and make them easier to review. I can consolidate some of them together if that helps.

skt/executable.py
```diff
@@ -222,33 +234,42 @@ def cmd_build(cfg):
     try:
         tgz = builder.mktgz(cfg.get('wipe'))
     except Exception as e:
-        save_state(cfg, {'buildlog': builder.buildlog})
+        save_state(
+            {
+                'buildlog': builder.buildlog
+            }
+        )
```
spbnick: Please avoid doing considerable reformatting together with logic changes. Please do the logic change in one commit, and reformatting in another, choose order as suits better. Then, reviews are easier, and we can drop or revert a problematic logic change, while still benefiting from reformatting.

skt/executable.py
```diff
@@ -177,19 +177,22 @@ def cmd_merge(cfg):
...
-            for patch in cfg.get('pw'):
-                save_state(cfg, {'patchwork_%02d' % idx: patch})
+            # Make a list of the patchwork patches provided by the user and
+            # save the list to the state file.
+            patches = [x for x in cfg.get("pw")]
+            save_state({"patchwork": patches})
+            for patch in patches:
                 ktree.merge_patchwork_patch(patch)
-                idx += 1
+
```
spbnick: Hey, nice that you put this into a separate commit! Also, this is where a structured storage format such as YAML shines!

skt/executable.py
```diff
@@ -688,6 +688,7 @@ def main():
     args = parser.parse_args()
 
     setup_logging(args.verbose)
+    global cfg
```
spbnick: Let's not do this. Moreover, let's pass only the state that the particular class or function needs. We need them more independent , not more entangled.

skt/executable.py
```diff
@@ -321,8 +321,32 @@ def cmd_run(cfg):
...
+        sys.exit(1)
+
+    # Ensure that the state file has a buildurl (comes from the publish step)
+    if not state.get('buildurl'):
+        print(
+            "ERROR:"
+            "The 'buildurl' item is missing from skt's state file. Please "
+            "run `skt publish` before attempting to test a kernel."
+        )
+        sys.exit(1)
```
spbnick: This will definitely help the user.

However, I would like this solved more efficiently. Instead of checking for
the presence of every variable in the state file, how about we have the state
recorded and verified explicitly?

We can e.g. have state "none" for nothing done, "merged" for code retrieved,
"built" for kernel built, and "tested" for kernel tested. Executing any of the
corresponding commands could then check that the state is the previous one or
above, and reset it accordingly.

Do you think you have time to implement the above in another PR? If not, then
let's not add these error messages lest they proliferate and increase our code
size.

Also, adding error messages like these would be better done in a separate
commit. I.e. refactor for the new interface in one, add functionality in
another.

skt/__init__.py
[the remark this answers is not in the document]
spbnick: This would make the all the previous patches break skt, which is what we're trying to avoid with the rule "every commit works". This also makes review difficult.

skt/executable.py
```diff
@@ -25,6 +25,7 @@
 import sys
 import time
 import traceback
+import yaml
```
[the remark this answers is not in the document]
spbnick: A faster approach can be continue using .ini for state, just split it out, and convert everything to YAML later. Would that work better for you, perhaps?

Issue #16, opened 2018-03-18
Running `skt publish` without --tarpkg produces stacktrace


spbnick:
Running `skt publish` without --tarpkg produces stacktrace:

```
Traceback (most recent call last):
  File "skt/skt.py", line 569, in <module>
    main()
  File "skt/skt.py", line 559, in main
    args.func(cfg)
  File "skt/skt.py", line 99, in wrapper
    func(cfg)
  File "skt/skt.py", line 202, in cmd_publish
    url = publisher.publish(cfg.get('tarpkg'))
  File "/home/nkondras/skt/skt/publisher.py", line 60, in publish
    shutil.copy(os.path.expanduser(source), self.destination)
  File "/usr/lib64/python2.7/posixpath.py", line 254, in expanduser
    if not path.startswith('~'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```
A readable error message about missing option should be produced instead.

PR #97, opened 2018-04-27
Fix: Running `skt publish` without --tarpkg produces stacktrace

danghai:
I create PR to fix: #16 . I added `raise exception` when skt publish is missing `tarpkg`. Please review my commit when you have a chance. Thanks!

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -249,6 +249,9 @@ def cmd_publish(cfg):
     """
     publisher = skt.publisher.getpublisher(*cfg.get('publisher'))
 
+    if not cfg.get('tarpkg'):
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -249,6 +249,9 @@ def cmd_publish(cfg):
     """
     publisher = skt.publisher.getpublisher(*cfg.get('publisher'))
 
+    if not cfg.get('tarpkg'):
+        raise Exception("skt publish is missing --tarpkg")
```
veruu: Can we have it say ```'skt publish is missing "--tarpkg <path>" option'``` to be clearer?

danghai:
@veruu thanks review. I have updated !

veruu:
Awesome, thanks!

PR #100, opened 2018-04-27
Reports humanization

veruu:
Split into a bunch of commits based on what part of report is reworked for ease of reviewing.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -416,11 +416,14 @@ class reporter(object):
         return result
 
     def getreport(self):
-        msg = list()
-
-        if self.cfg.get("krelease"):
-            msg.append("result report for kernel %s"
-                       % self.cfg.get("krelease"))
+        msg = ['Hello,\n',
+               'we appreciate your contributions to the Linux kernel and would'
+               ' like to help test',
+               'them. Below are the results of automatic tests we ran on a '
+               'patchset you\'re ',
+               'involved with, with hope it will help you find possible issues'
+               ' sooner.',
+               '\n']
 
         msg += self.getmergeinfo()
 
@@ -430,6 +433,14 @@ class reporter(object):
             msg += self.getbuildfailure()
         else:
             msg += self.getjobresults()
+
+        msg += ['\nPlease check out the report above and reply to this email '
+                'in case you find',
+                'an issue with the testing process or wish to not receive '
+                'these reports',
+                'anymore.',
+                '\nSincerely,',
+                '  Kernel CI Team']
 
         if self.attach and self.attach[0][0] == "config":
             self.attach.append(self.attach.pop(0))


spbnick:
Hey, a get on PR ID :smile: :tada:!

skt/reporter.py
```diff
@@ -415,27 +426,35 @@ def getreport(self):
             msg += self.gettested()
             msg += self.getjobresults()
 
+        msg += ['\nPlease review the above outcome and reply to this email in'
+                ' case',
+                'it\'s deemed incorrect.',
+                '\nSincerely,',
+                '  Kernel CI Team']
```
spbnick: In general, we can't really tell people what to do, so better just be helpful. Also, an outcome cannot really be incorrect, it's just a fact, and we're not passing any judgment here to object to. We can also put a "thank you" here.

Perhaps like this:

    Please reply to this message, if you wish to report an issue with our build
    setup or our tests, suggest an improvement, or exclude yourself from
    receiving these messages.

    Thank you for your contribution to the Linux kernel!

    Sincerely,
    Kernel CI Team,
    Red Hat

And, yeah, we'll need to provide a way for people to stop receiving the messages.

skt/reporter.py
```diff
@@ -436,21 +436,23 @@ def getreport(self):
...
+            subject += "Report"
+
+        if self.cfg.get("krelease"):
+            subject += " for kernel %s" % self.cfg.get("krelease")
+
+        if not self.cfg.get('mergelog') and not self.cfg.get('buildlog'):
+            if self.cfg.get('retcode') == '0':
+                subject += ': PASS'
+            else:
+                subject += ': FAIL'
```
spbnick: Hmm, I liked the PASS/FAIL at the start of the subject. I think it's the most important piece of information and should be at the start. I would even vote for removing `[skt]`. Also, I don't think the git repo name is that important. How often do people submit patches to different repos/maillists at the same time, anyway?

What do you think?

skt/reporter.py
```diff
@@ -414,11 +414,7 @@ def getjobresults(self):
         return result
 
     def getreport(self):
-        msg = list()
-
-        if self.cfg.get("krelease"):
-            msg.append("result report for kernel %s"
-                       % self.cfg.get("krelease"))
+        msg = ['Hi,\nthanks for your contribution to Linux kernel!\n']
```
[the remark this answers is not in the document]
veruu: "Hello" works for me, however since we will be running skt internally too I wanted the message to work with that as well and your version seems too much for that. Maybe we can think of something in the middle?

skt/reporter.py
veruu: I wanted to put the positive thing (thanking) at the beginning of the message so it doesn't sound too sarcastic in case everything failed.

Definitely agree with adding a way to exclude people but since we don't have it implemented yet, we shouldn't be advertising it in the message. We can add it as it's implemented.

skt/reporter.py
```diff
@@ -324,8 +328,9 @@ def getmergefailure(self):
 
     def getbuildfailure(self):
         attname = "build.log"
-        result = ["\n-----------------------",
-                  "Build failed: see attached %s" % attname]
+        result = ['however the build failed. We are attaching the occurred '
+                  'failure for',
+                  'you to review.']
```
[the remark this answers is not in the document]
spbnick: Yeah, if it's only one, then we don't need it, strictly speaking, but if we simply mention the attachment name with every such sentence, then we can combine them as we see fit by rearranging the code or function calls.

skt/reporter.py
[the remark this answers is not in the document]
spbnick: Are you talking about "Hello", or the whole introduction? I'm not sure what to replace "Hello" with, which would be similarly respectful. Regarding the paragraph, I think it's OK to mention our team and company names internally, we want to be known and recognized inside too :)

Could you give me suggestions or specific points you'd like changed?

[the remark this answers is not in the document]
veruu: "Hello" is fine. The rest of the introduction looks too formal, robotic and not very positive. Especially with combination of failed tests it sounds like "hey, you suck and here is a robot to call you out". From my experience with some upstream communities, I think it's very important to start positive.

Also, another point regarding mentioning the company and team name: since this project is in the wild now and anybody - including other people without any involvement with us - can use it, having it hardcoded (without any possibility to configure) would just cause confusion. We might want to add a configurable signature, but that's for another day.

skt/reporter.py
spbnick: I'm unsure about the thanking in general. I'm not sure if we're in the position to be thanking for the contribution to the Linux kernel directly, as we (Red Hat) don't own it. I.e. the sender of the patch is not sending it to us, therefore we can't really thank them. Let me think about that a little.

Regarding excluding, I think we *need* to put it in, especially at the start, when our reports and tests will be the worst. People need to know they're free to not receive those and we are not forcing them. It will help them see the reports better and accept them easier.

For now we can just respond to those e-mails ourselves, and quickly add an address filter to sktm setup when/if somebody asks that. It will also be acceptable to ask people to wait a little, while we implement that, as long as we promise and actually do that.

skt/reporter.py
```diff
@@ -273,24 +273,27 @@ def update_mergedata(self):
         self.mergedata = mergedata
 
     def getmergeinfo(self):
```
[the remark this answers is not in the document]
veruu: Added

skt/reporter.py
```diff
@@ -416,11 +416,7 @@ def getjobresults(self):
         return result
 
     def getreport(self):
-        msg = list()
-
-        if self.cfg.get("krelease"):
-            msg.append("result report for kernel %s"
-                       % self.cfg.get("krelease"))
+        msg = ['Hello,\nthanks for your contribution!\n']
```
spbnick: How about this:

    Hello,

    We at Red Hat use the Linux kernel heavily in our products, and really
    appreciate the work you put into your contributions. We'd like to help
    maintain the kernel at the best possible quality, as it would also benefit us.

    That's why Red Hat Kernel CI team runs automatic tests on patch sets posted to
    development maillists. Below you can find results of one such run on a patch
    set you've been involved with.

First of all, it doesn't assume the recipient is a patch author (who most of
the time won't be), but tries to grateful for all contributions (including
reviews). This also states from the get-go where the message is coming from
(not everyone looks at from), and why we're doing this. It makes it clear
we're benefitting from this, which I think is important. I realize my
language can be a bit dry, so I would appreciate your help in making it
softer. However, I think this is the information we need to present at the
beginning of the message.

skt/reporter.py
spbnick: I still think we should not ask people to read the report, as it would imply that we require them to heed it. Instead we should make it clear what it contains from the start, make it easy to read and understand, and give people the opportunity to use it, as a help. But we should in no way insist or even hint that they should.

How about this:

    Please reply to this message if you find an issue with our testing
    process, or wish to stop receiving these reports.

    Sincerely,
    Kernel CI Team,
    Red Hat

We can't really write "reply to these message if you want to start receiving
them again" as they shouldn't be getting them then. Instead I think it's OK to
let people know how to re-subscribe in response to their unsubscribe request.
Something like this:

    We're sorry to know you didn't find our test reports useful. We're working
    hard to improve them, and would be glad to hear what you found unsuitable
    and receive any suggestions you might have.

    If you wish to start receiving these reports again, simply send us a
    message at kernel-ci-list@redhat.com.

skt/reporter.py
```diff
@@ -333,8 +333,9 @@ def getmergefailure(self):
 
     def getbuildfailure(self):
         attname = "build.log"
-        result = ["\n-----------------------",
-                  "Build failed: see attached %s" % attname]
+        result = ['However, the build failed. We are attaching the build '
+                  'output for',
+                  'you to review (build.log).']
```
spbnick: Oh, and I think "for you to review" is also a no-go. We cannot ask people to review any of our results. We're offering help.

skt/__init__.py
```diff
@@ -186,7 +186,9 @@ def git_cmd(self, *args, **kwargs):
         args = list(["git", "--work-tree", self.wdir, "--git-dir",
                      self.gdir]) + list(args)
         logging.debug("executing: %s", " ".join(args))
-        subprocess.check_call(args, **kwargs)
+        subprocess.check_call(args,
+                              env=os.environ.update({'LC_ALL': 'C'}),
```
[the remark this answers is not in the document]
veruu: Damn, forgot about the return value. Will fix that. What do you mean by changing the environment for everything? It should only change for the subprocess call.

skt/reporter.py
veruu: Will change it to "... attaching the build output for more information"

skt/reporter.py
veruu: I can't reply to your comment below which GitHub shows as outdated so I'm replying to both here.

ad footer, I assumed most of the time the email will go to a mailing list as well so anybody who has access to it can reply. Explaining resubscription in unsubscribing message is a good idea though! We don't have the un/re-subscribing in action yet but again, I'd avoid hardcoding the email address there. This is an open project and doing so *without the ability to configure it* is another entry barrier for anyone who'd like to use it.

This applies to the introduction too. It reads like a marketing email instead of a test report (I mean, I want to save it for any propagation I might need in the future because I like it that much, but I don't think it has a place in a test report), and it's too long without actually saying anything that is important *to the developer*. I strongly believe that the report should get to the point fast, so the introduction should only contain necessities ("hi, we ran tests on these patches and this is the result").  Besides, who doesn't look at the sender of the message? Together with the subject it's the first thing you see before you even open it.

spbnick: Yes, we absolutely need to get these company-specific messages out of skt. I agree it's another barrier for use by general public. I think this should be a part of configurable message template. Moreover, once again, I think e-mail sending shouldn't be a part of skt at all, but of sktm. For now we'll have to keep it here, though.

Regarding sounding as a marketing message, I agree, but I also think we are representing the company here and if anything goes wrong the company will be responsible, so we should be open and forthcoming about who actually stands behind the test result. OTOH, it's a great opportunity to promote the company as spending extra money and computing power to get the kernel tested, which is good for everyone. If you make an effort for people, you should make it clear and yourself known, or it will not be appreciated (see e.g. the typical struggle of sysadmins being acknowledged for their work). The language of this can be improved, I agree, but let's get something merged and discuss all together and with other kernel developers.

However, I think you're right about saying the most important thing first. So,
let's try to reorder things (although moving it further might help clarity, I
think):

    Hello,

    We at Red Hat use the Linux kernel heavily in our products, and really
    appreciate the work you put into your contributions. Below you can find
    results of automatic tests we ran on a patchset you're involved with.

    We'd like to help maintain the kernel at the best possible quality, as it
    would also benefit us. That's why Red Hat Kernel CI team maintains a
    system running automatic tests on kernel contributions, and shares the
    results, hoping to help developers and maintainers find issues early.

veruu: > Yes, we absolutely need to get these company-specific messages out of skt. I agree it's another barrier for use by general public. I think this should be a part of configurable message template. Moreover, once again, I think e-mail sending shouldn't be a part of skt at all, but of sktm.

100% agreed. For anyone who wants to use it, having this introduction would make the project a no-go. Until we get the configuration, I don't think we should hardcode any company info. As you say, anything goes wrong and the company is responsible. Now imagine someone unrelated trying to help with testing kernel on their own but screwing up. While in the end it should come up it wasn't us, it still leaves the bad taste and a mess to clean up (and while people will be able to do it with the configuration too, it would be active impersonation and not accident).

Again, if I was a non-RH developer on the receiving end I'd stop reading after the half of the first sentence (TBH I'd do it even as a Red Hatter because it reminds me of a sponsorship spam which, as a developer, I am not interested in). And that would be the better case, since there is certainly a lot of devels with a short fuse who'd complain about us on social media etc. Why are we doing it belongs to an article or a blog post, possibly before we launch it live on upstream mailing lists so it's easily searchable, but definitely not here.

spbnick: Alright, perhaps you're right about putting it away to a blog post or such. We
can give a "read more" link, where things will be explained, then. I still
think we need to clearly state it's Red Hat, but perhaps not explain
everything.

I think our branding could go into skt for a short while, since we're not
promoting it yet, but since we're going live internally only, that can wait
and we can make it an external e-mail template, and return to branding and the
text later. And then, I think it might be easier to not let it in than to come
around to removing it later.

Alright, how about this:

    Hello,

    We appreciate your contributions to the Linux kernel and would like to help
    test them. Below are the results of automatic tests we ran on a patchset
    you're involved with. We hope it will help you find possible issues earlier.

veruu: "Read more" link at the end sounds good. That way it wouldn't take the attention away from the test report itself, while people who'd be interested in it could check it out.

This introduction sounds pretty good, I think we found the middle ground :) I'll add it shortly.

PR #101, opened 2018-04-30
Add coveralls to skt

danghai:
I make PR to add coveralls to skt in order to provide statistics about code coverage. Please review them when you have the chance. Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,17 @@
 language: python
 python:
     - "2.7"
+env:
+    - TOX_ENV=py27
 install:
     # Install skt using pip to ensure dependencies are downloaded correctly.
     - pip install .
+    - pip install coveralls tox
 script:
     # Run tests and linters.
     - flake8 --show-source .
     - pylint tests
     - python -m unittest discover tests
     # Test the uninstallation of skt,
+    - tox -e $TOX_ENV
     - pip uninstall -y skt
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,11 @@ language: python
 python:
     - "2.7"
 env:
-    - TOX_ENV=py27
+    - TOX_ENV=py27,flake8,pylint
 install:
     # Install skt using pip to ensure dependencies are downloaded correctly.
     - pip install .
     - pip install coveralls tox
 script:
-    # Run tests and linters.
-    - flake8 --show-source .
-    - pylint tests
-    - python -m unittest discover tests
-    # Test the uninstallation of skt,
     - tox -e $TOX_ENV
     - pip uninstall -y skt
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,14 @@
+[tox]
+envlist =
+    py27
+
+[testenv]
+passenv = TRAVIS TRAVIS_*
+deps =
+    coverage
+    mock
+    coveralls
+commands =
+    coverage run --source=skt,tests -m unittest discover tests
+    coverage report -m
+    coveralls
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,8 @@
 [tox]
 envlist =
     py27
+    flake8
+    pylint
 
 [testenv]
 passenv = TRAVIS TRAVIS_*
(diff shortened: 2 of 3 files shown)


veruu:
@major I like the idea too. But IMHO, if we want to use `tox` then we should also run the tests via `tox`, not use it only for coverage. If we don't want to change the current status of tests (since we are *very* far from py3) and only add coverage, then we should do it in the easiest way possible, and adding two dependencies, one of which is pretty much deprecated and [the maintainers advise not using it](https://nose.readthedocs.io/en/latest/) doesn't look like the best idea to me.

danghai:
@major Thanks your review.
@veruu I think I can add run the tests via `tox`. Or if more dependencies are not fine in this time, I am going to figure out the simpler way then, And I can reopen the PR when we are near to `py3`, ( may be more test env). What do you think?

danghai:
@veruu okay I removed nose, and run tests only once. 
Because I keep the `tox`, so I put all built from `travis` to `tox.ini` including `flake8`, `pylint`. Is it good enough? I may separate commits into small commits when someone approve all these changes. Thanks

tox.ini
```diff
@@ -0,0 +1,25 @@
...
+    py27
+    flake8
+    pylint
+
+[testenv]
+deps =
+    coverage
+    mock
+commands =
+    coverage run --source=tests -m unittest discover tests
```
veruu: This only runs the coverage on `tests` subdirectory (so it will be always 100%), not the whole repository. You can use `--source=.` to fix that.

veruu:
Commenting here since GitHub doesn't think you should comment on a commit, only on separate lines - @danghai can you also explain the change to tox in the commit message? A single sentence is enough, just so it doesn't go unnoticed.

@major , @spbnick the current coverage submission to coveralls.io doesn't work, I think because we don't have the account linked there. If anybody wants to deal with administration stuff it would be great to do that.

danghai:
@veruu I have updated!, I thought you only want coverage in test. 
So you do not want me separate them into small commit? You want me describe what I have changed in commit message relating to `tox`?
After link to coveralls.io, it is good to display their badges in `README`.

veruu:
@danghai you can do it either way. Doing it in two commits is cleaner since it's technically two logical changes, but at the same time it's a single "add coverage and run it via tox so we don't run the tests multiple times" commit. So I'm fine either way for this one.

danghai:
I am not sure why it has coverage `.tox/`. I think I already put them into `gitignore`. You can view result here:
https://coveralls.io/builds/16775435

spbnick:
@veruu, I enabled skt repo on coveralls.io, but am not sure if that will be sufficient.

@danghai, I think you need to run `coverage run --source=skt,tests` so it doesn't go into all the subdirectories, which we don't need anyway.

veruu:
@spbnick awesome, thanks!

@danghai looks like to make it work with Travis, we need to specify the [env vars in tox.ini](https://coveralls-python.readthedocs.io/en/latest/usage/tox.html). Can you add them?

danghai:
@veruu I have updated it!.

.travis.yml
```diff
@@ -1,13 +1,14 @@
...
 script:
-    # Run tests and linters.
-    - flake8 --show-source .
-    - pylint tests
-    - python -m unittest discover tests
-    # Test the uninstallation of skt,
+    - tox -e $TOX_ENV
     - pip uninstall -y skt
+after_success:
+    - coveralls
```
veruu: Based on the docs about coveralls and tox, coveralls should be run from tox too

major:
@danghai I agree with @veruu -- if you can get the `coveralls` command into the `tox.ini` so that it works for manual tox tests outside of Travis-CI, that would be great!

danghai:
@major  @veruu I get the `coveralls` command into the `tox.ini`. :)

danghai:
@veruu @major It is really weird. It pass all test in my private Travis: https://travis-ci.org/danghai/skt
But it is fail in here.

veruu:
I'm thinking it has to do with access restrictions between GitHub repos and coveralls. Checking it out now.

veruu:
@spbnick as the owner, you should get an email requesting the access

spbnick:
@veruu, should be done now!

veruu:
Looking good now! I had to enable the coverage for the repo in the coveralls too, but that could be done only after the owner acked the integration

veruu:
Thanks everyone for involvement (yay teamwork!) and especially @danghai for the patience!

danghai:
Thanks @veruu @spbnick @major review my PR :)

PR #104, opened 2018-05-03
Refine report messages

spbnick:
Refine report messages a bit, according to changes suggested on the internal maillist and in PR comments.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -295,9 +295,9 @@ class reporter(object):
             result += ['  - %s,' % pname,
                        '    grabbed from %s\n' % purl]
 
-        result += ['on top of commit %s from' % self.mergedata['base'][1][:12],
-                   '  %s' % self.mergedata['base'][0],
-                   'repository.']
+        result += ['on top of commit %s from the repository at' %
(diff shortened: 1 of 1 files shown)


skt/reporter.py
```diff
@@ -295,9 +295,10 @@ def getmergeinfo(self):
             result += ['  - %s,' % pname,
                        '    grabbed from %s\n' % purl]
 
-        result += ['on top of commit %s from' % self.mergedata['base'][1][:12],
-                   '  %s' % self.mergedata['base'][0],
-                   'repository.']
+        result += ['on top of commit %s from the repository at' %
+                   self.mergedata['base'][1][:12],
+                   '  %s' %
+                   self.mergedata['base'][0]]
```
veruu: Can you put this together with the previous line? Technically there's nothing wrong with it, but since it logically belongs together and fits < 80 chars it would be cleaner.

spbnick: Yeah, I changed it to be more readable for me, but your way is alright too :) Done!

spbnick:
Thank you, Veronika :)!

PR #105, opened 2018-05-03
Rework merge_patch_file

### LINKED ISSUE #99: merge_patch_file doesn't provide logs if the patch application fails
While the `ktree.merge_patchwork_patch` provides logs about the patch application failure, the sister method `ktree.merge_patch_file` does not. This means we don't have any feedback to report to the patch author if the patch failed to apply.

Because of the similarity between the functions that would be introduced by the fix, it would be nice if we could make the `merge_patchwork_patch` retrieve the patch file and then use `merge_patch_file` internally.

danghai:
I make PR to rework `merge_patch_file`. Fix #99.

1. It provides logs if the patch application fails. 
2. It must have `patch` file in `skt-workdir` to merge. So, in code, I copy the patch file in skt to `skt-workdir`, and then remove it. It is really awkward. But I am not sure how to handle it. 
3. I test by download the patch file `mbox` from https://patchwork.ozlabs.org/patch/886637/, and then use command:

```
$ skt --rc skt-rc --state --workdir skt-workdir -vv \
                    --patchlist net-next-cxgb4-notify-fatal-error-to-uld-drivers.patch
```
Could you review it when you have the chance? Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,22 @@ E.g. to apply a particular patch to a particular, known-good commit from the
                    --ref a870a02cc963de35452bbed932560ed69725c4f2 \
                    --pw https://patchwork.ozlabs.org/patch/886637
 
+To apply a local patch run:
+
+    $ skt --rc <SKTRC> --state --workdir <WORKDIR> -vv \
+             merge --baserepo <REPO_URL> \
+                   --ref <REPO_REF> \
+                   --patchlist <PATH_TO_PATCH>
+
+Here, `<PATH_TO_PATCH>` would be the patch file.
+E.g. to apply a particular patch to a particular, known-good commit from the
+"net-next" repo, run:
+
+     $ skt --rc skt-rc --state --workdir skt-workdir -vv \
+              merge --baserepo git://git.kernel.org/pub/scm/linux/kernel/git/davem/net-next.git \
+                    --ref a870a02cc963de35452bbed932560ed69725c4f2 \
+                    --patchlist net-next-cxgb4-notify-fatal-error-to-uld-drivers.patch
+
 #### Faster clones
 
 In some instances, a full git history is not needed. Shallow clones are git
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -364,10 +364,17 @@ class ktree(object):
     def merge_patch_file(self, path):
         if not os.path.exists(path):
             raise Exception("Patch %s not found" % path)
+        args = ["git", "am", path]
         try:
-            self.git_cmd("am", path)
-        except subprocess.CalledProcessError:
+            subprocess.check_output(args,
+                                    cwd=self.wdir,
+                                    env=dict(os.environ, **{'LC_ALL': 'C'}))
+        except subprocess.CalledProcessError as exc:
             self.git_cmd("am", "--abort")
+
+            with open(self.mergelog, "w") as fp:
+                fp.write(exc.output)
+
             raise Exception("Failed to apply patch %s" % path)
 
         self.info.append(("patch", path))
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -362,6 +362,8 @@ class ktree(object):
         self.info.append(("patchwork", uri, patchname.replace(',', ';')))
 
     def merge_patch_file(self, path):
+        if not os.path.exists(path):
+            raise Exception("Patch %s not found" % path)
         try:
             self.git_cmd("am", path)
         except subprocess.CalledProcessError:
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -157,7 +157,7 @@ def cmd_merge(cfg):
             idx = 0
             for patch in cfg.get('patchlist'):
                 save_state(cfg, {'localpatch_%02d' % idx: patch})
-                ktree.merge_patch_file(patch)
+                ktree.merge_patch_file(os.path.abspath(patch))
                 idx += 1
 
         if cfg.get('pw'):


veruu:
Hi, thanks for the pull request! However, there are a few larger problems with it:

1. It does *not* provide the missing `mergelog`.
2. `git am` is perfectly capable of applying a patch file given the path. Do you have any reason for the change, or why you copy the patch file?
3. `-` for `git am` marks input from stdin, not file path. I'm not even sure why it worked in your test.
4. Following up, empty `gam.communicate()` doesn't do anything.

danghai:
@veruu I am not sure why it does not work when the patch file in `skt folder` except the patch file in `skt-workdir`. The error: (ignore the name `test6.patch`, I just rename it)
```
2018-05-03 19:59:39,751    DEBUG   executing: git --work-tree skt-workdir --git-dir skt-workdir/.git am test6.patch
fatal: could not open 'test6.patch' for reading: No such file or directory
2018-05-03 19:59:39,815    DEBUG   executing: git --work-tree skt-workdir --git-dir skt-workdir/.git am --abort
fatal: Resolve operation not in progress, we are not resuming.
2018-05-03 19:59:39,878    DEBUG   state: mergelog -> skt-workdir/merge.log
  File "/usr/bin/skt", line 11, in <module>
  File "/home/danghai/skt/skt/executable.py", line 713, in main
    args.func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 116, in wrapper
    func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 172, in cmd_merge
    raise e
subprocess.CalledProcessError: Command '['git', '--work-tree', 'skt-workdir', '--git-dir', 'skt-workdir/.git', 'am', '--abort']' returned non-zero exit status 128
```
Have you try the `patchlist` option in merge and it works fine?

danghai:
@veruu I am not sure why it does not merge when the `patch file` in `skt folder`, except it only merges when the `patch file` in `skt-workdir`. The error (ignore the name `test6.patch`, I just rename it):
```
2018-05-03 19:59:39,751    DEBUG   executing: git --work-tree skt-workdir --git-dir skt-workdir/.git am test6.patch
fatal: could not open 'test6.patch' for reading: No such file or directory
2018-05-03 19:59:39,815    DEBUG   executing: git --work-tree skt-workdir --git-dir skt-workdir/.git am --abort
fatal: Resolve operation not in progress, we are not resuming.
2018-05-03 19:59:39,878    DEBUG   state: mergelog -> skt-workdir/merge.log
  File "/usr/bin/skt", line 11, in <module>
  File "/home/danghai/skt/skt/executable.py", line 713, in main
    args.func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 116, in wrapper
    func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 172, in cmd_merge
    raise e
subprocess.CalledProcessError: Command '['git', '--work-tree', 'skt-workdir', '--git-dir', 'skt-workdir/.git', 'am', '--abort']' returned non-zero exit status 128
```
Have you try with `patchlist` option in merge, and it works fine?

veruu:
@danghai you can use `os.path` module to create absolute path from relative and use that. Of course do check for exceptions (if the file exists, has right permissions, etc). As I said before, it will avoid the need to pass absolute path as argument to `skt`, and also avoid unhelpful message (file not found when the file obviously exists).

The difference between functions is that `merge_patch_file` passes filepath to `git am`, while `merge_patchwork_patch` grabs the *contents* of the mbox file as a string (*not* the file, but what's inside it) and passes those, similarly as you'd do in interactive mode of any other command. That's the reason why calling `git_cmd` in the first example is enough, but you need to open a communication tunnel and pass the data explicitly in the second.

Does it make sense now?

danghai:
@veruu Thanks! it makes sense. The `patch` from `mbox` is a string, so It can pass as an argument in `communicate` for `STDIN` process. But the `patch` in `merge_patch_file` is a file ( not string), I try to get content of file to pass. However using `git am <patch>` is good enough for `merge_patch_file`. 
So, what I should do for this patch now? Create the absolute path for file, and then pass them to `git_cmd`. 
I looks it does not provide the missing `mergelog` in origin `merge_patch_file` too.

veruu:
@danghai yes, that would be perfect. One commit for the absolute path usage (explain in the commit message why relative paths didn't work) and add example usage in the README as another commit since as you noticed, there isn't one.

> I looks it does not provide the missing mergelog in origin merge_patch_file too.

Yes, that's the reason why issue #99 was opened. So it can be added. If you want to add another commit that would create the mergelog and solve the issue, it would be more than awesome!

danghai:
@veruu Could you help me? I got stuck from getting `stdout` from `subprocess.check_call` in `git_cmd` I need to have `stdout` to write `merge.log`. How I can get this?

veruu:
@danghai see `stderr` and `stdout` passing in `merge_git_ref`. Check out the `subprocess` module documentation for more info, and don't hesitate to try everything out to see how it works!

danghai:
Yeah, I check carefully this function to get data from `stdout`. However, I cannot get them. So I switch to use `subprocess.Popen` along with `communicate` to get stdout then. Could you review it now?

skt/__init__.py
```diff
@@ -405,10 +405,27 @@ def merge_patchwork_patch(self, uri):
...
+        gam = subprocess.Popen(
+            ["git", "am", abspath],
+            cwd=self.wdir,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            env=dict(os.environ, **{'LC_ALL': 'C'})
+        )
+
+        (stdout, stderr) = gam.communicate()
```
veruu: Since we don't need to pass any input for communicate and only grab `stdout`, this can be implemented much simpler by `subprocess.check_output` and then saving the mergelog from exception's `output` attribute.

veruu:
@danghai yeah, I agree that the `merge_git_ref` is a bit confusing. We might rework it a bit later. For now, I have an idea that would be a lot easier to use.

skt/__init__.py
danghai: Could you elaborate? Because I see `git_cmd` already uses `subprocess.check_output. Do you have the way to get `output` attribute from this function without modifying?

danghai:
@veruu Sorry, I am not sure how to get `stdout` without `subprocess.Popen`. I think `output = subprocess.check_call(["git", "am", abspath], cwd=self.wdir)` will return `output` string. But it is not. That is your idea?

veruu:
@danghai no. I mentioned `subprocess.check_output`, not `subprocess.check_call`. Please check the documentation on how they differ and how to retrieve the output if it's needed.

danghai:
@veruu I have updated. I miss information both 'subprocess.check_output` and `subprocess.check_call` exception's output attribute when it raises `CalledProcessError`. I write the explanation in my commit. Thanks a lot

skt/__init__.py
```diff
@@ -405,11 +405,19 @@ def merge_patchwork_patch(self, uri):
                           patchinfo.get("name").replace(',', ';')))
 
     def merge_patch_file(self, path):
+        if not os.path.isfile(path):
+            raise Exception("Patch %s not found" % os.path.basename(path))
```
veruu: It would be less confusing to print out the original `path` passed, rather then only the file name. Imagine that two patches have the same file name, but are stored in different directories. It wouldn't be clear which one the exception is about.

skt/__init__.py
```diff
@@ -405,11 +405,19 @@ def merge_patchwork_patch(self, uri):
...
+        abspath = os.path.abspath(path)
         try:
-            self.git_cmd("am", path)
-        except subprocess.CalledProcessError:
+            subprocess.check_output(["git", "am", abspath], cwd=self.wdir)
```
veruu: Can you also add the `env` locale override (as you can see in the other function)?

skt/__init__.py
```diff
@@ -405,11 +405,19 @@ def merge_patchwork_patch(self, uri):
...
+        except subprocess.CalledProcessError, stdout:
             self.git_cmd("am", "--abort")
-            raise Exception("Failed to apply patch %s" % path)
+
+            with open(self.mergelog, "w") as fp:
+                fp.write(stdout.output)
+
+            raise Exception("Failed to apply patch %s" %
+                            os.path.basename(path))
```
veruu: Same as with the nonexistent patch here, it would be better to use either original, or absolute path here rather than file name.

skt/__init__.py
```diff
@@ -405,11 +405,19 @@ def merge_patchwork_patch(self, uri):
...
 
     def merge_patch_file(self, path):
+        if not os.path.isfile(path):
+            raise Exception("Patch %s not found" % os.path.basename(path))
+        abspath = os.path.abspath(path)
         try:
-            self.git_cmd("am", path)
-        except subprocess.CalledProcessError:
+            subprocess.check_output(["git", "am", abspath], cwd=self.wdir)
+        except subprocess.CalledProcessError, stdout:
```
veruu: This works with Python2 (which we currently support), but the syntax is deprecated in Python3. As we'd eventually like to support both in the future, it would be easier to use `except subprocess.CalledProcessError as <identifier_name>:` syntax (which is compatible) instead of `,`. Using `stdout` as identifier for exception is also confusing, usually `exc` is being used. Can you change it?

skt/__init__.py
danghai: Ah what does it mean `env`? I see it in several `subprocess.Open`. But I am not sure about it

skt/__init__.py
danghai: That makes sense!

danghai:
@veruu . Please help me with cleaner English, and properly formatted . I appreciate it. Thanks!

skt/__init__.py
danghai: Oh I see. That makes sense!

skt/__init__.py
veruu: `env` overrides process' environment. You can read more about what environment variables are for example [here](https://wiki.archlinux.org/index.php/environment_variables). The variable we add ensures git's output stays same (in English) and independent from user's language or charset settings, so we can easily search it and add to reports.

veruu:
@danghai okay, here is my take on the commit messages. I'm not a native speaker either, but I hope they are clearer:

The first one:
```Raise exception if the patch to apply doesn't exist```

Second:
```README: Document --patchlist option from merge command```

And the last one - I think we can make this one much shorter since most of what you describe is present in the documentation of `subprocess` module:
```
Provide mergelog if the application of a local patch failed

merge_patch_file() didn't provide logs if the patch failed to apply.
This means the reports sent about merge failures were incomplete.
Fix the issue by saving git's output if the application failed.

Fixes #99
```

Don't forget to rebase on top of current master!

**Comment by danghai** [comment]
@veruu Thanks! I think you provide a tool to check format, grammar, etc :) for in future that I can write it by myself. ^^
I have updated and rebase too

File: skt/__init__.py [hunk 6]
```diff
@@ -406,13 +406,19 @@ def merge_patchwork_patch(self, uri):
 
     def merge_patch_file(self, path):
         if not os.path.isfile(path):
-            raise Exception("Patch %s not found" % os.path.basename(path))
+            raise Exception("Patch %s not found" % os.path.abspath(path))
```
veruu: Can you please do this change in the commit that introduced the exception? That way if anything happens and the commit needs to be reverted it is easy to do so. We are currently working on an addition to contributing guide (PR #112) that mentions it.

danghai: Sorry my bad. Let me rebase and do again then

veruu:
@danghai no problem! We won't be providing any tools like that since they are usually integrated into standard editors. For example, I use `vim` for commit messages and it wraps the lines automatically, and I can use `:set spell` command to check grammar. I don't know which editor you use, but both features are available in most editors either by default, or via various plugins.

danghai:
@veruu I remember Nick mentioned it before. Unfortunately, I am using `atom` :(. I am not familiar with `vim`.  I rebased and fixd it.

veruu:
Looks good to me now Hai, thank you!

Calling in @major and @spbnick if they want to check this out in case I missed something. I can verify the `merge.log` is now provided.

skt/__init__.py
spbnick: This can produce incorrect error message. E.g. when somebody enters a path to a directory instead of a file (e.g. from botched auto-complete). So, instead use `os.path.exists`, and let the "directory instead of file" produce an exception. That will be less confusing.

skt/__init__.py
```diff
@@ -405,8 +405,11 @@ def merge_patchwork_patch(self, uri):
                           patchinfo.get("name").replace(',', ';')))
 
     def merge_patch_file(self, path):
+        if not os.path.isfile(path):
+            raise Exception("Patch %s not found" % os.path.abspath(path))
+        abspath = os.path.abspath(path)
```
spbnick: We already have many cases of normalizing paths deep in the code which need to be fixed. Let's not add another one. Instead, call `os.path.abspath` in `cmd_merge`, or better yet during configuration/command-line option parsing/loading.

The next step would be to document the function, mention that the path should be absolute, and verify that inside the function (e.g. with an `assert` and `os.path.isabs`). @veruu, can we have assertions in our Python code? What are the recommended practices?

README.md
```diff
@@ -165,6 +165,23 @@ E.g. to apply a particular patch to a particular, known-good commit from the
...
 
+To apply a local patch run:
+
+    $ skt --rc <SKTRC> --state --workdir <WORKDIR> -vv \
+             merge --baserepo <REPO_URL> \
+                   --ref <REPO_REF> \
+                   --patchlist <PATH_TO_PATCH>
+
+Here, `<REPO_REF>` would be the reference to checkout, and to apply the patch
+on top of, and `<PATH_TO_PATCH>` would be the patch file.
```
spbnick: Let's not repeat what `REPO_REF` is. Otherwise we would need to explain every placeholder every time, which would make the document hard to read.

skt/__init__.py
```diff
@@ -409,10 +409,16 @@ def merge_patch_file(self, path):
             raise Exception("Patch %s not found" % os.path.abspath(path))
         abspath = os.path.abspath(path)
         try:
-            self.git_cmd("am", abspath)
-        except subprocess.CalledProcessError:
+            subprocess.check_output(["git", "am", abspath],
+                                    cwd=self.wdir,
+                                    env=dict(os.environ, **{'LC_ALL': 'C'}))
```
spbnick: It would be nice to unify those somehow eventually, so we don't have to copy-paste things around.

skt/__init__.py
danghai: I think I agree with @veruu 's perspective

skt/__init__.py
veruu: Indeed, eventually factoring out the common code would be nice. Or even modifying `git_cmd` to be able to choose `check_call` or `check_output`. But that's for a different discussion

danghai:
@spbnick @veruu I have updated it!

skt/__init__.py
spbnick: > Technically we can, but assertions are usually associated with tests instead of production code.

So Python test framework authors hijacked `assert`, apparently not knowing its original purpose, and made it useless? Oh, well...

> If part X of the code base is responsible for making the path absolute, verification it works as expected should be part of tests for part X.

Writing tests checking that a function interface is observed for every invocation of that function doesn't scale. Letting the function check its arguments works better. Whether it's with an assertion, or with an if statement, doesn't really matter.

BTW, this is called "precondition checking" in "[Design by contract](https://en.wikipedia.org/wiki/Design_by_contract)" approach, which I find very effective for keeping developers' sanity, and would like to see used in our code as well.

> It is harder to debug the problem if you get an error in unrelated part of the code.

That's why we have backtraces :)

veruu: One of disadvantages of using `assert` instead of custom `try/except` block is that you always get `AssertionError` instead of `TypeError`, `ValueError` etc, which are more appropriate in the main code base. Indeed, you can (and should) pass your own message, but the exception class stays the same, which might create some issues in unexpected places.

For example, function A should raise `ValueError` if passed value doesn't fit expected range, and `TypeError` if the input is wrong type. Function B, which calls A, might want to do something specific (eg. cleanup) based on the exception A raises. Raising same `AttributeError` for both different issues will just complicate things (and doesn't go well with *explicit is better than implicit* rule).

Another

veruu: The explanation got a bit long so I made the relevant tl;dr bold for ease.

One of disadvantages of using assert instead of throwing exception is that you always get `AssertionError` instead of `TypeError`, `ValueError` etc, which are more appropriate in the main code base. Indeed, you can pass your own message, but the exception class stays the same, which might create some issues in unexpected places.

For example, function A should raise `ValueError` if passed value doesn't fit expected range, and `TypeError` if the input is wrong type. Function B, which calls A, might want to do something specific (eg. cleanup) based on the exception A raises. Raising same `AttributeError` for both different issues will just complicate things (and doesn't go well with *explicit is better than implicit* rule).

Similarly, it renders exception inheritance useless. You might want to catch all `OSError` exceptions, but if all you get is `AssertionError`, try to differentiate between `OSError` sub-exceptions and anything else.

While there isn't any written rule about it, using assert in production code creates more issues than it solves.

> Writing tests checking that a function interface is observed for every invocation of that function doesn't scale. Letting the function check its arguments works better. Whether it's with an assertion, or with an if statement, doesn't really matter.

If the function has nondeterministic interface, then the problem lays much deeper and it should be fixed, not hidden. **However, if you really need to check the attributes (eg. we don't have an init function that sanitize the values), please raise relevant exception (whether it's in `try/except` block or `if condition: raise X`) instead of assertion.**

> That's why we have backtraces :)

Indeed, so instead of "here is the function that messed up" we can go dig through a bunch of them :)

danghai: I think there are cons and pros for each approach including using `assertion` and `exception`. In my opinion, this PR focus on rework the `merge patch file`. We could probably consider the idea's @spbnick in another issue with another PR in order to have `precondition checking` not only for `merge_patch_file` function, but also for other functions relating to `precondition checking`. ( Of course with the approval of @veruu ). So I mean we keep the raise `exception` in this case in the PR. What do you think?

skt/__init__.py
danghai: @spbnick I see it is good idea to unify those parameter. So, I create another PR to do it: #117

skt/__init__.py
veruu: Works for me, what about you @spbnick ? Let's not mix in problems different from what this pull is trying to solve

skt/__init__.py
spbnick: @danghai, could you please address my comment here? I.e. use `os.path.exists`, not `os.path.isfile`.

skt/__init__.py
spbnick: @danghai, could you please address my [first comment](https://github.com/RH-FMK/skt/pull/105/commits/1817bd1bb06c3203c630ed5ecae6a9053011ea1e#r186373587) here? I.e. move calling `os.path.abspath` to `cmd_merge`.

danghai:
@spbnick thanks review. I have updated it.

danghai:
@spbnick I have rebased!

PR #107, opened 2018-05-03
Minor fixes

veruu:
Use generic exception name and fix patch name retrieval in reporter


skt/reporter.py
```diff
@@ -242,9 +242,13 @@ def stateconfigdata(self, mergedata):
...
-                (rpc, patchid) = skt.parse_patchwork_url(purl)
-                pinfo = rpc.patch_get(patchid)
-                mergedata['patchwork'].append((purl, pinfo.get("name")))
+                response = requests.get(os.path.join(purl, 'mbox'))
+                if response.status_code != requests.codes.ok:
+                    raise Exception('Failed to retrieve patch from %s, '
+                                    'returned %d' % (purl,
+                                                     response.status_code))
+                patchname = re.findall(r'Subject:\s(.*)', response.content)[0]
+                mergedata['patchwork'].append((purl, patchname))
```
spbnick: Oh, man, copy-paste...
Waaaait, is this what we're doing in the main code now?
What about header wrapping, escaping, character encoding?
This doesn't handle any of that.
See https://tools.ietf.org/html/rfc2822#section-2.2
and https://tools.ietf.org/html/rfc2047

Sorry, I'll have to revert cfef31e for now. As much as I would love to see Patchwork reliance gone from skt, we'll have to go to staging without this today. Do you think you could redo this so it doesn't need copy-pasting and extracts Subject properly? Thank you.

PR #112, opened 2018-05-04
CONTRIBUTING.md: Add commit and documentation reqs

spbnick:


spbnick:
@veruu, I think that would be good, but I don't subscribe under "tests for everything" approach, as it is not that efficient and would slow us down. If you agree, do you have an idea of how we can describe the criteria for needing tests?

veruu:
@spbnick I agree with the developer slowdown. I'd say that if a function is changed to behave differently, tests (if present) need to be fixed - in the end that would break the build if they weren't. And that if a new function is added and the class it's part of already has base tests (or the function is independent), add some simple unit tests for pass / fail modes. This would avoid having to deal with eg `runner` methods, while (hopefully) keeping the test coverage.

Does this make sense, or do you have a better idea how to articulate it? I think it's important to mention tests because then everybody (including me) would just ignore them and they are useful because you can find a bunch of bugs while writing them. But as you said, they shouldn't be the focus of our work right now.

CONTRIBUTING.md
```diff
@@ -17,6 +17,13 @@ a patch, please do the following:
 * Limit the first line of your commit message to 50 characters
 * Describe the bug you found or the feature that is missing from the project
 * Describe how your patch fixes the bug or improves the project
+* Make sure the code keeps working after each commit, and does not require any
+  further commits to fix it.
+* Make sure each commit contains only one complete logical change, no more, no
+  less.
+* When adding a new module, a class, or a function always document them.
```
major: Small nit: "Always document each new module, class, or function."

CONTRIBUTING.md
```diff
@@ -17,6 +17,13 @@ a patch, please do the following:
...
+* When changing the logic or the interface (as opposed to e.g. formatting) of
```
major: Perhaps:

> Update the documentation when the logic or the interface of an existing module, class, or function changes.

CONTRIBUTING.md
spbnick: Yes, that's better. Thanks, Major!

spbnick:
I addressed two of Major's comments and added requirements @dzickusrh suggested internally.
Please take a look and discuss :)

Regarding what @major said about tests, I think that would essentially mean requiring everything tested, which I'm not willing to accept (at least yet). I don't have an idea of how to describe "test important and complicated stuff" at the moment, and I think it will take us a long time to find out and agree upon, so I propose to merge this PR without it and think about and discuss this more otherwise.

spbnick:
@veruu, @major, @dzickusrh, are we OK with this?

spbnick:
@dzickusrh approved this on IRC. @veruu, could you please merge this once @major agrees? Thanks!

dzickusrh:
Works for me. :-)

PR #113, opened 2018-05-04
Rework patch retrieval from Patchwork

veruu:
Using XMLRPC for patch retrieval can pose a problem especially with
newer Patchwork2 which have the legacy interface disbled. For the
limited functionality skt needs, retrieving patch mbox is enough and
compatible with both Patchwork1 and Patchwork2 versions.

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -12,6 +12,9 @@
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+from email.errors import HeaderParseError
+import email.header
+import email.parser
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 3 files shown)


skt/__init__.py
```diff
@@ -12,117 +12,75 @@
...
+        # Emails return None if the header is not found so use a stub subject
+        # instead of it
+        return 'subject'
+
+    try:
+        # decode_header() returns a list of tuples (value, charset)
+        return email.header.decode_header(subject)[0][0]
+    except HeaderParseError:
+        # We can't parse the original subject so use a stub one instead
+        return 'subject'
```
spbnick: Argh, I'm really uncomfortable with just substituting this with "subject". First of all, it will lead to things like `Subject: subject`, so at least it would be better if it returned something like `<MISSING>` and `<INVALID>`, and those would be documented as well.

Or perhaps even better, we could stop pretending we're returning the "subject", and say we're returning patch "name" instead. I.e. rename the function to `get_patch_name`, loading it with the meaning that we give that patch a name as best we can, including naming the patch `<SUBJECT MISSING>`, and `<SUBJECT ENCODING INVALID>`. This would also help cover the gap where we assign "patch subject" to variable named "patch name". What do you think?

skt/__init__.py
```diff
@@ -12,117 +12,75 @@
...
+    headers = email.parser.Parser().parsestr(content, True)
+    subject = headers['Subject']
+    if not subject:
+        # Emails return None if the header is not found so use a stub subject
+        # instead of it
+        return 'subject'
+
+    try:
+        # decode_header() returns a list of tuples (value, charset)
+        return email.header.decode_header(subject)[0][0]
```
spbnick: Let's go the whole way here. Otherwise we can end with partial subjects with wrong encoding in our reports.

Here's an example of an encoded subject from RFC 2047:

    Subject: =?ISO-8859-1?B?SWYgeW91IGNhbiByZWFkIHRoaXMgeW8=?=
        =?ISO-8859-2?B?dSB1bmRlcnN0YW5kIHRoZSBleGFtcGxlLg==?=

`decode_header()` will return this for it:

    [('If you can read this yo', 'iso-8859-1'), ('u understand the example.', 'iso-8859-2')]

Which we can then feed into a simple loop, using `str.decode()`. Something
like this (not tested):

    decoded_piece_list = []
    for piece in email.header.decode_header(subject):
        decoded_piece_list.append(piece[0].decode(piece[1], "replace"))
    return "".join(decoded_piece_list)

tests/test_init.py
```diff
@@ -22,44 +22,35 @@
...
+
+    def test_ok_patch_subject(self):
+        """Ensure get_patch_subject() returns correct 'Subject' if present"""
+        mbox_body = 'From Test Thu May 2 17:49:51 2018\nSubject: GOOD SUBJECT'
+        self.assertEqual('GOOD SUBJECT', skt.get_patch_subject(mbox_body))
+
+    def test_encoded_patch_subject(self):
+        """Ensure get_patch_subject() correctly decodes 'Subject'"""
+        mbox_body = ('From Test Thu May 2 17:49:51 2018\n'
+                     'Subject: =?utf-8?q?=5BTEST=5D?=')
```
spbnick: Perhaps also test with that example from the RFC I mention in the other comment.

skt/__init__.py
veruu: Those are all valid points and good ideas, thanks!

skt/__init__.py
veruu: Ew. Will fix.

veruu:
@spbnick resolved all, thanks again for good suggestions!

spbnick:
Looks great, thank you, Veronika!

PR #116, opened 2018-05-07
Support passing of subject to mailreporter

veruu:
This allows non-default subjects, which is especially useful when
replying to an already existing message.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -488,7 +488,8 @@ class mailreporter(reporter):
     """A reporter sending results by e-mail"""
     TYPE = 'mail'
 
-    def __init__(self, cfg, mailfrom, mailto, mailinreplyto=None):
+    def __init__(self, cfg, mailfrom, mailto, mailinreplyto=None,
+                 subject=None):
         """
         Initialize an e-mail reporter
 
@@ -500,6 +501,9 @@ class mailreporter(reporter):
             mailinreplyto:  A string containing the value of the "In-Reply-To"
                             header to add to the message. No header is added
                             if evaluates to False.
+            subject:        Subject to use, for example subject of the message
+                            to reply to. Default subject describing test
+                            results is used if not present.
         """
         # The From: address string
         self.mailfrom = mailfrom
@@ -507,12 +511,13 @@ class mailreporter(reporter):
         self.mailto = [to.strip() for to in mailto.split(",")]
         # The value of "In-Reply-To" header
         self.mailinreplyto = mailinreplyto
+        self.subject = subject
         super(mailreporter, self).__init__(cfg)
 
     def report(self):
         self.update_mergedata()
         msg = MIMEMultipart()
-        msg['Subject'] = self.getsubject()
+        msg['Subject'] = self.subject if self.subject else self.getsubject()
         msg['To'] = ', '.join(self.mailto)
         msg['From'] = self.mailfrom
         if self.mailinreplyto:


spbnick:
Regarding describing `subject` as "the report subject", I meant that it would then allow us to say that the message subject consists of the "outcome" (PASS/FAIL) and "the report subject" (patch subject, or detailed outcome (ew, but OK)), and then perhaps add base kernel version to that later.

veruu:
> This is a good start, but I think it needs to be a bit more involved. In particular I think we should still have `PASS:` or `FAIL:` prefix.

I'm not so sure about this. From my point of view, using the original subject has a purpose
of 1) accommodating mail clients and 2) making it clear it's a reply.

There is a lot of mail clients which get confused if you reply to a message with a different
subject and act like it's a separate thread and not a reply; and a bunch of them (which get
the threading correctly based on other headers) will only show you the original subject which
the thread started with (unless you check the raw message) and not the subject of each
message in the thread. Unfortunately, most of mail clients do this, from my experience. From
this perspective, using a slightly modified subject has zero advantage.

Ignoring the above, using a slightly modified subject line is more confusing than going all
the way, whether by using the original or the constructed one. The original one looks like a
normal reply, the constructed one is visibly automated but again, it's clear. OTOH, the
slightly modified one is not - it doesn't provide enough information to make out it's an
automated test report but doesn't look like a reply either; especially the `FAIL: <original
subject>` looks like a mailserver sending you the message back because it failed to deliver.

Of course if you have an idea how to include enough information on top of the original subject
(so it's clear and not too long) I'm all ears but until then, I'd say going full custom or
full original subject is better.

spbnick:
Hmm, I don't think most e-mail clients have a problem handling the "Re: " prefix other clients are putting there in various languages (e.g. "Re: ", "AW: ", "На: ", etc.), so they will probably handle "PASS: " and "FAIL: " just as well. What do you think? We can skip adding the kernel version to the subject if that is a problem, it's not *very* important. However, I think the test outcome is hugely important, and it would be good if we could get it there.

spbnick:
And if any clients hide such a subject, well, then it's the problem of whoever uses them. We will have the result in the body anyway.

veruu:
`Re:` is fine, but some of clients I used got confused by the localization since I don't use it, and things got marked as a new thread because the prefix was not recognized.

Mail clients aside, for example Patchwork also treats emails not starting with `Re:` as separate patches or pulls and if the email can't be interpreted as such, tries to match it as cover letter and only afterwards tries to find a parent submission for a possible comment. So if we reply to a mailing list too (which we will, since we grab `To` and `Cc` headers), we might cause some problems if we a) reply to a cover letter or b) include any snippet in the report which can be interpreted as a diff or pull.

Right now we should be fine, but if we change the reply logic to cover letter (if it's present), or we'd like to dump a part of the patch which caused conflicts during application, people using it won't be happy. That said, problem b) will manifest with a custom subject without `Re:` too...

Anyways, I'd be more content with a solution that is not prone to cause us headaches in the future.

veruu:
> OK, how about we do `Re: <ORIGINAL SUBJECT> - <RESULT>` (`<RESULT>` == `PASS`/`FAIL`), or if that doesn't work, just `Re: <ORIGINAL SUBJECT>` (i.e. what a simple reply with a MUA would do)?

That looks better, as it shouldn't break stuff (at least what I know of). I'm not convinced about usefulness of appending to a long subject, and based on yesterday's meeting with @dzickusrh too I'd go with the latter option for now (which is what the pull does).

PR #117, opened 2018-05-08
Unify parameter for subprocess

danghai:
I make PR to unify parameter for `subprocess`. It would be nice to unify the parameter in `subprocess` so that we do not need to copy-paste everytime. Could you review it when you have the chance? Thanks


danghai:
I see. Close now, and reopen when I figure out your point.

PR #129, opened 2018-05-14
Expand basecfg path earlier

major:
Expanding the `basecfg` path earlier ensures that any function that uses the path to the base kernel config is getting the same absolute path to the file.

Partial fix for #94.

- [x] Requires #126 to merge first.
- [x] Requires #128 to merge next.


major:
@danghai Yes, but I will put that into a different PR.

danghai:
@major Could I take this with this commit: https://github.com/danghai/skt/commit/5cd7c3f36ffcb89f6202e8fe729ba5367665043e ?

PR #132, opened 2018-05-15
Fix report message if no patches were applied

### LINKED ISSUE #110: Handle absent patches in reports
At present, invocations of skt without patches applied produce the following
in reports:

    We applied the following patch:

    on top of commit 23b389c23139 from the repository at
       git://git.kernel.org/pub/scm/linux/kernel/git/mkp/scsi.git 

Fix the reports to produce appropriate text for these cases.

veruu:
If the merge was tested and no patches were applied, the message still
said 'We applied the following patch...' and '... tests we ran on a
patchset you are involved with'. Both parts are misleading and wrong
since there are no patches.

Check if there are any patches listed first and choose appropriate
wording based on the data.

Fixes #110

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -283,22 +283,29 @@ class reporter(object):
         Returns: A list of strings representing data about applied patches and
                  base repository.
         """
-        result = ['We applied the following patch']
-        if len(self.mergedata['localpatch'] + self.mergedata['patchwork']) > 1:
-            result[0] += 'es:\n'
+        patchlist = self.mergedata['localpatch'] + self.mergedata['patchwork']
+
+        if patchlist:
+            result = ['We applied the following patch']
+            if len(patchlist) > 1:
+                result[0] += 'es:\n'
+            else:
+                result[0] += ':\n'
+
+            for patchpath in self.mergedata['localpatch']:
+                result += ['  - %s' % patchpath]
+
+            for (purl, pname) in self.mergedata['patchwork']:
+                result += ['  - %s,' % pname,
+                           '    grabbed from %s\n' % purl]
+
+            result += ['on top of commit %s from the repository at' %
+                       self.mergedata['base'][1][:12],
+                       '  %s' % self.mergedata['base'][0]]
         else:
-            result[0] += ':\n'
-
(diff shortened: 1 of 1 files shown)


veruu:
@major can you please take a look? I'd like to get this fixed before today's deployment, since it was brought to my attention it bugs the hell out of me.

skt/reporter.py
```diff
@@ -420,13 +427,17 @@ def getreport(self):
...
                'would like to help',
-               'test them. Below are the results of automatic tests we ran on '
-               'a patchset',
-               'you\'re involved with, with hope it will help you find '
-               'possible issues sooner.',
-               '\n']
+               'test them. Below are the results of automatic tests we ran']
+        if self.mergedata['localpatch'] or self.mergedata['patchwork']:
+            msg[-1] += ' on a patchset'
+            msg += ['you\'re involved with, with hope it will help you find '
```
major: This wording is okay, but we could say something like `automatic tests we ran on a patchset that involves you`. I was trying to find a way to avoid using the contraction (`you're`) since that can be challenging for some non-native English speakers.

skt/reporter.py
```diff
@@ -283,22 +283,29 @@ def getmergeinfo(self):
...
-                   '  %s' % self.mergedata['base'][0]]
+            for (purl, pname) in self.mergedata['patchwork']:
+                result += ['  - %s,' % pname,
+                           '    grabbed from %s\n' % purl]
+
+            result += ['on top of commit %s from the repository at' %
+                       self.mergedata['base'][1][:12],
+                       '  %s' % self.mergedata['base'][0]]
+        else:
+            result = ['We grabbed git tree with HEAD at %s from the '
```
major: Perhaps: `We cloned the git tree and checked out %s from the repository at...`. Using `cloned` allows us to be a bit more specific than `grabbed`. ;)

skt/reporter.py
veruu: This was what @spbnick came up with after our long discussion about the intro message and I kept the wording here. We can discuss this all together if you think the wording should be changed a bit.

PR #138, opened 2018-05-15
Fix variable names in cmd_build()

### LINKED ISSUE #130: Rename makeopts to something more specific
Within `cmd_build()`, the `makeopts` argument is confusing since it actually contains extra make arguments and options that the user has asked to use during the build process. The variable should be named something more specific to what it is actually doing.

This was mentioned in @spbnick's comment in #85.

major:
Change makeopts to extra_make_args to signify that they are extras
provided by the user.

Change defmakeargs to make_argv_base since they are the base
arguments/options for kernel builds.

Simplify variable handling just before the kernel build.

Fixes #130.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,9 @@ and improve build times. In some cases, deep troubleshooting may require
 debug symbols. Use the `--enable-debuginfo` argument to build a kernel with
 debug symbols included.
 
+Provide additional arguments and options to `make` by using
+`--makeopts`.
+
 ### Publish
 
 To "publish" the resulting build using the simple "cp" (copy) publisher run:
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -438,22 +438,23 @@ class KernelTree(object):
 
 
 class KernelBuilder(object):
-    def __init__(self, path, basecfg, cfgtype=None, makeopts=None,
+    def __init__(self, path, basecfg, cfgtype=None, extra_make_args=None,
                  enable_debuginfo=False):
         self.path = path
         self.basecfg = basecfg
         self.cfgtype = cfgtype if cfgtype is not None else "olddefconfig"
         self._ready = 0
-        self.makeopts = None
         self.buildlog = "%s/build.log" % self.path
-        self.defmakeargs = ["make", "-C", self.path]
+        self.make_argv_base = ["make", "-C", self.path]
         self.enable_debuginfo = enable_debuginfo
 
-        if makeopts is not None:
+        # Split the extra make arguments provided by the user
+        if extra_make_args:
             # FIXME: Might want something a bit smarter here, something that
             # would parse it the same way bash does
-            self.makeopts = makeopts.split(' ')
-            self.defmakeargs += self.makeopts
+            self.extra_make_args = extra_make_args.split(' ')
+        else:
+            self.extra_make_args = []
 
         try:
             os.unlink(self.buildlog)
@@ -465,7 +466,7 @@ class KernelBuilder(object):
 
     def prepare(self, clean=True):
         if (clean):
-            args = self.defmakeargs + ["mrproper"]
+            args = self.make_argv_base + ["mrproper"]
             logging.info("cleaning up tree: %s", args)
             subprocess.check_call(args)
 
@@ -483,7 +484,7 @@ class KernelBuilder(object):
             logging.info("disabling debuginfo: %s", args)
             subprocess.check_call(args)
 
-        args = self.defmakeargs + [self.cfgtype]
+        args = self.make_argv_base + [self.cfgtype]
         logging.info("prepare config: %s", args)
         subprocess.check_call(args)
         self._ready = 1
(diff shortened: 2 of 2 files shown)


danghai:
@major Do you think is it good to mention `extra_make_args` in `README`. For example, to speed up the process considerably, we can build the kernel and its module with the -j option. The command:
```
$ skt --rc skt-rc --state --workdir skt-workdir -vv \
         build -c /boot/config-`uname -r` --extra_make_args=-j4
```

major:
Good idea, @danghai. I forgot about that. :)

skt/__init__.py
```diff
@@ -437,22 +438,21 @@ def bisect_iter(self, bad):
...
-        if makeopts is not None:
-            # FIXME: Might want something a bit smarter here, something that
-            # would parse it the same way bash does
-            self.makeopts = makeopts.split(' ')
-            self.defmakeargs += self.makeopts
+        # Split the extra make arguments provided by the user
+        if extra_make_args:
+            self.extra_make_args = shlex.split(extra_make_args)
+        else:
+            self.extra_make_args = []
```
spbnick: Could you please switch to `shlex.split` in a separate commit?

README.md
```diff
@@ -216,6 +216,9 @@ and improve build times. In some cases, deep troubleshooting may require
 debug symbols. Use the `--enable-debuginfo` argument to build a kernel with
 debug symbols included.
 
+Provide additional arguments and options to `make` by using
+``--extra-make-args``.
```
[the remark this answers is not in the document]
major: Good catch. Fixing.

major:
@spbnick Rebased and the conflict is resolved.

README.md
```diff
@@ -216,6 +216,9 @@ and improve build times. In some cases, deep troubleshooting may require
 debug symbols. Use the `--enable-debuginfo` argument to build a kernel with
 debug symbols included.
 
+Provide additional arguments and options to `make` by using
+``--makeopts``.
```
spbnick: Hmm, should this use double single quotes?

skt/__init__.py
```diff
@@ -438,22 +438,21 @@ def bisect_iter(self, bad):
...
         self._ready = 0
-        self.makeopts = None
         self.buildlog = "%s/build.log" % self.path
-        self.defmakeargs = ["make", "-C", self.path]
+        self.make_argv_base = ["make", "-C", self.path]
         self.enable_debuginfo = enable_debuginfo
 
-        if makeopts is not None:
-            # FIXME: Might want something a bit smarter here, something that
-            # would parse it the same way bash does
```
spbnick: This FIXME should be removed in the commit that switches to shlex.

skt/__init__.py
```diff
@@ -531,15 +531,23 @@ def mktgz(self, clean=True, timeout=60 * 60 * 12):
...
-                                   "targz-pkg"]
+        # Set up the arguments and options for the kernel build
+        targz_pkg_argv = [
+            "INSTALL_MOD_STRIP=1",
+            "-j%d" % multiprocessing.cpu_count(),
+            "targz-pkg"
+        ]
+        kernel_build_argv = (
+            self.make_argv_base +
+            targz_pkg_argv +
```
veruu: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

skt/__init__.py
veruu: I agree. @major can you move that to the other commit please?

skt/__init__.py
```diff
@@ -531,15 +530,23 @@ def mktgz(self, clean=True, timeout=60 * 60 * 12):
...
 
-        args = self.defmakeargs + ["INSTALL_MOD_STRIP=1",
-                                   "-j%d" % multiprocessing.cpu_count(),
-                                   "targz-pkg"]
+        # Set up the arguments and options for the kernel build
+        targz_pkg_argv = [
+            "INSTALL_MOD_STRIP=1",
+            "-j%d" % multiprocessing.cpu_count(),
+            "targz-pkg"
+        ]
```
veruu: What was wrong with the previous formatting that it needed to be changed? I won't stop the pull from being merged because of it, but it makes the definition longer without any reason (the arguments would fit on the line when starting on the same line as opening bracket).

skt/__init__.py
```diff
@@ -531,15 +531,23 @@ def mktgz(self, clean=True, timeout=60 * 60 * 12):
...
```
veruu: What was wrong with the previous formatting that it needed to be changed? I won't stop the pull from being merged because of it, but it makes the definition longer without any reason (the arguments would fit on the line when starting on the same line as opening bracket).

README.md
major: Ah, I've been doing too much RST lately. ;)

skt/__init__.py
major: Will do.

skt/__init__.py
major: Will do.

major:
@veruu / @spbnick: Should look better now.

PR #143, opened 2018-05-21
Fix the ordering of imports

major:
This PR works towards the goals of #3 by ordering imports properly and checking those imports with `flake8-import-order`.

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -12,9 +12,9 @@
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-from email.errors import HeaderParseError
 import email.header
 import email.parser
+import io
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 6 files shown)


skt/__init__.py
```diff
@@ -12,19 +12,20 @@
...
 import logging
 import multiprocessing
 import os
 import re
 import shutil
 import subprocess
 import sys
-import io
 import time
+from email.errors import HeaderParseError
```
veruu: Moving this introduces `pylint` error `C: 26, 0: Imports from package email are not grouped (ungrouped-imports)`.

skt/reporter.py
```diff
@@ -12,15 +12,16 @@
...
+import StringIO
 import gzip
 import logging
 import os
 import re
 import smtplib
-import StringIO
+from email.mime.application import MIMEApplication
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
```
veruu: PEP8 doesn't say anything about the actual order in the groupings, but lexical order is usually used because of readability and orientation reasons. This breaks the lexical order while actually not fixing anything and making the orientation around imports more confusing?

major:
Let's just forget about this one for now.

PR #147, opened 2018-05-21
Simplify response code check

major:
The response object has an `ok` method that returns True if the status
code is < 400.

This works towards #3.

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -50,7 +50,7 @@ def get_patch_mbox(url):
     except requests.exceptions.RequestException as exc:
         raise(exc)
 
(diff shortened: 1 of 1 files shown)


veruu:
While technically this should be ok (no pun intended), we are retrieving directly the mbox and no redirects should occur. If they do, there's a chance something is wrong, and this would hide it.

major:
@veruu True, but the `requests` module automatically follows redirects, so this should be okay as long as the resulting redirect returns something < 400.

http://docs.python-requests.org/en/master/api/

major:
I see your point. We can forget about this for now.

PR #152, opened 2018-05-21
Rename logger constant to LOGGER

major:
Pylint's C0103 requires that constants are in uppercase.

This PR works towards #3.


veruu:
@spbnick there is an "update branch" button, it creates a merge commit but you can then squash it when merging. Not sure how it works with multi-commit pulls but when there is a single one and no conflicts it's OK and avoids the constant rebase loop

PR #154, opened 2018-05-22
Get full path to source in publish

danghai:
I make PR for getting full path to source in publish

### DIFF
--- a/skt/publisher.py
+++ b/skt/publisher.py
@@ -58,9 +58,7 @@ class CpPublisher(Publisher):
     TYPE = 'cp'
 
     def publish(self, source):
-        # FIXME Move expansion up the call stack, as this limits the class
-        # usefulness, because tilde is a valid path character.
-        shutil.copy(os.path.expanduser(source), self.destination)
+        shutil.copy(source, self.destination)
         return self.geturl(source)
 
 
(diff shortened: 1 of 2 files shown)


danghai:
@veruu . In `cmd_publish`, it is going to publish the `buildinfo` and `buildconf`. Getting full path `buildinfo` and `buildconf` in configuration `cfg` can fix this issue? I am not sure about it.

skt/publisher.py
```diff
@@ -58,20 +58,16 @@ class CpPublisher(Publisher):
...
 class ScpPublisher(Publisher):
     TYPE = 'scp'
 
     def publish(self, source):
-        # FIXME Move expansion up the call stack, as this limits the class
-        # usefulness, because tilde is a valid path character.
         subprocess.check_call(["scp",
-                               os.path.expanduser(source),
+                               source,
                                self.destination])
```
veruu: Since the full call fits into one line now, can you merge those lines so it doesn't take up so much space?

danghai: Yeah, I have intend to ask you about make absolute path for `tarpkg`, but it seems to me the `load_config` become longer. 
Ah I got what you mean. We should get absolute path for all configuration early. And then, we are not worry about path for later design

danghai:
Hmm not sure why it has the change from previous patch

spbnick:
Thank you, Hai, Veronika! This looks good, merging. However, I would've been happier if the commit subject matched the change better :)

PR #155, opened 2018-05-23
Rename kernel source vars

### LINKED ISSUE #42: Use keyword arguments when initializing skt.kbuilder
The arguments list for `skt.kbuilder` is getting a bit long with the addition of `enable_debuginfo`. We should change this to use named arguments instead.

This was noted by @spbnick in #33.

major:
The workdir/kernel source path was defined as `self.path`, which is confusing. The kernel source won't be directly in the workdir after the work for #84 is complete.

Fixes #42.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -201,11 +201,11 @@ def cmd_build(cfg):
                                         "%Y%m%d%H%M%S")
 
     builder = skt.KernelBuilder(
-        cfg.get('workdir'),
-        cfg.get('baseconfig'),
-        cfg.get('cfgtype'),
-        cfg.get('makeopts'),
-        cfg.get('enable_debuginfo')
+        source_dir=cfg.get('workdir'),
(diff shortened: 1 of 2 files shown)


skt/executable.py
```diff
@@ -201,11 +201,11 @@ def cmd_build(cfg):
...
-        cfg.get('workdir'),
-        cfg.get('baseconfig'),
-        cfg.get('cfgtype'),
-        cfg.get('makeopts'),
-        cfg.get('enable_debuginfo')
+        kernel_source_dir=cfg.get('workdir'),
+        basecfg=cfg.get('baseconfig'),
+        cfgtype=cfg.get('cfgtype'),
+        extra_make_args=cfg.get('makeopts'),
+        enable_debuginfo=cfg.get('enable_debuginfo')
```
veruu: Can't comment on the commit, but can you please mention in the commit message that this fixes issue #42 ?

PR #156, opened 2018-05-23
[WIP] Store kernel source in a directory

major:
This PR renames `wdir` to `workdir` to make it more clear. Also, it clones/fetches the kernel source into `workdir/source`. This allows work for #84 to continue since it allows multiple directories to exist inside the workdir without conflicts.


major:
Forgot to make the source directory. Will fix that now.

major:
Ignore this one for now. I need to figure out how to handle the kernel source directory so that `skt.KernelBuilder` can use it as well.

PR #157, opened 2018-05-24
Add new YAML state file functions/tests

major:
This PR contains the initial work for YAML-formatted state files. These functions are not yet in use by `skt`, so merging this code should not affect an existing deployment.

The PR that adds code that uses these functions is in progress.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,7 @@ install_requires = flake8
                    pycodestyle!=2.4.0
                    pylint
                    requests
+                   PyYAML
 
 # The beaker-client package breaks some Python packaging rules and tries to
 # store content in the system /etc directory. This causes a Sandbox violation
--- a/skt/state_file.py
+++ b/skt/state_file.py
@@ -1,0 +1,87 @@
+#!/usr/bin/python2
+
+# Copyright (c) 2017 Red Hat, Inc. All rights reserved. This copyrighted
+# material is made available to anyone wishing to use, modify, copy, or
+# redistribute it subject to the terms and conditions of the GNU General
+# Public License v.2 or later.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Functions that manage the skt state file."""
+import logging
+import os
+import yaml
+
+
+def destroy(cfg):
+    """
+    Destroy the state file.
+
+    Args:
+        cfg: A dictionary of skt configuration
+    """
+    state_file = cfg.get('state')
+
+    if os.path.isfile(state_file):
+        try:
+            os.unlink(state_file)
+        except IOError as exception:
+            logging.error("Failed to delete state file: %s", exception)
+            raise exception
+
+
+def read(cfg):
+    """
+    Read the state file from the disk.
+
+    Args:
+        cfg: A dictionary of skt configuration
+    """
+    # Return an empty state dictionary if the state file does not exist
+    current_state = {}
+
+    state_file = cfg.get('state')
+    logging.debug("Reading state from: %s", state_file)
+
+    if os.path.isfile(state_file):
+        try:
+            with open(state_file, 'r') as fileh:
+                current_state = yaml.load(fileh)
+        except IOError as exception:
+            logging.error("Failed to read state file: %s", exception)
+            raise exception
+
+    return current_state
+
+
+def update(cfg, state_updates):
+    """
+    Update the state file on disk with new state data.
+
+    Args:
+        cfg:           A dictionary of skt configuration.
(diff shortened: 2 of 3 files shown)


skt/executable.py
```diff
@@ -44,6 +45,58 @@ def full_path(path):
...
+    if os.path.isfile(state_file):
+        with open(state_file, 'r') as fileh:
+            current_state = yaml.load(fileh)
+    else:
+        current_state = {}
+
+    return current_state
+
+
+def save_state_file(cfg, state_updates):
```
spbnick: Could you perhaps rename these functions so that it's clearer they're operating on the same stuff? E.g. name them `state_file_destroy`, `state_file_read`, and `state_file_save`? Or even make a module for them called `state_file.py` and name the functions `destroy`, `read`, and `save` respectively?

Oh, and also, `save_state_file` name is somewhat confusing and would imply that we're overwriting the file and not merging the data into it. I think something like `update_state_file` or `state_file_update` or `state_file.update` would be better.

setup.cfg
```diff
@@ -20,6 +20,7 @@ install_requires = flake8
                    pycodestyle!=2.4.0
                    pylint
                    requests
+                   PyYAML
```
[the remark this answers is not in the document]
veruu: I was able to find it in RHEL7 repositories (but again, it is possible to install any packages via `pip` without depending on the underlying system so it's not a problem if it's not there in the future).

skt/executable.py
```diff
@@ -44,6 +45,58 @@ def full_path(path):
...
+    new_state = current_state.copy()
+    new_state.update(state_updates)
+
+    # Save the new state to the state file
+    logging.debug("Saving state: {}".format(state_updates))
+    state_file = cfg.get('state')
+    with open(state_file, 'w') as fileh:
+        fileh.write(
+            yaml.dump(new_state, default_flow_style=False)
+        )
```
veruu: Can we have this shortened on a single line since the split formatting is not needed? :)

skt/executable.py
```diff
@@ -44,6 +45,58 @@ def full_path(path):
...
+def read_state_file(cfg):
+    """
+    Read the state file from the disk
+
+    Args:
+        cfg: A dictionary of skt configuration
+    """
+    state_file = cfg.get('state')
+    if os.path.isfile(state_file):
+        with open(state_file, 'r') as fileh:
```
veruu: We should account for a situation when the file doesn't exist / we don't have permission to open it etc

skt/executable.py
```diff
@@ -44,6 +45,58 @@ def full_path(path):
...
+            current_state = yaml.load(fileh)
+    else:
+        current_state = {}
```
veruu: Can we get rid of the `current_state` variable altogether, or at least initialize it at the beginning of the function (no `else` needed)? Together with the exceptions catching we don't want to enter a code path where it would be undefined. /me remembers a similar problem with reporter

skt/executable.py
major: Makes sense. I'll work on that.

skt/executable.py
major: Sure.

skt/executable.py
major: That makes sense. I'll add that logic.

setup.cfg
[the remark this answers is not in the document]
major: It's `python-pyyaml` in CentOS 7. I haven't checked RHEL yet.

major:
@spbnick Since we're the only users of `skt`, I planned to simply switch over to the new format in one patch. ;)

setup.cfg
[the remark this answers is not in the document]
spbnick: @veruu, we got to remember that we'll probably need an RPM package. And RPM package can only depend on other RPM packages, not pip packages.

skt/state_file.py
```diff
@@ -0,0 +1,89 @@
...
+        # Read from the state file
+        with open(state_file, 'r') as fileh:
+            current_state = yaml.load(fileh)
+
+    return current_state
+
+
+def update(cfg, state_updates):
+    """
+    Save state to the state file on disk.
```
spbnick: This would be nice to update to "update" :)

skt/state_file.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Destroy the state file.
+
+    Args:
+        cfg: A dictionary of skt configuration
+    """
+    state_file = cfg.get('state')
+
+    # Check if the state file exists
+    if os.path.isfile(state_file):
```
veruu: Can we use try/except blocks here and later too? Feel free to raise the exception you get, but they'd be much more readable (and pythonic) than multiple `if` checks, and would avoid race conditions too

skt/state_file.py
major: Oops. Thanks.

skt/state_file.py
```diff
@@ -0,0 +1,86 @@
...
+    Args:
+        cfg: A dictionary of skt configuration
+    """
+    state_file = cfg.get('state')
+
+    if os.path.isfile(state_file):
+        try:
+            os.unlink(state_file)
+        except IOError as exception:
+            logging.error("Failed to update state file: %s", exception)
```
veruu: I'm so sorry to bother you again in this pull, but shouldn't this be "Failed to *remove* state file"?

major: Right. I'll fix it.

PR #158, opened 2018-05-25
Use new state file

major:
- [x] Requires #157 to merge first.
- [x] Requires #160 to merge first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ defaults: &defaults
     - run:
         name: Test `skt merge`
         command: |
-          skt -vv --state -d workdir --rc sktrc merge \
+          skt -vv -d workdir --rc sktrc --state skt-state.yml merge \
             --pw https://patchwork.ozlabs.org/patch/895383/ \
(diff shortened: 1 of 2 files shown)


major:
Hold off on merging this one -- I need to roll through and see where `cfg` needs to be replaced with `state`.

major:
Nevermind, it looks like I misunderstood the logic there for a moment. This should be okay to review.

dzickusrh:
This can be come confusing unless we set the ground rules right.
I think 'state' should contain all the runtime needed data.  That includes the initial cfg data.  After the state file is created, we should no longer use 'cfg' for the rest of the commands.
Treat 'state' as a copy of 'cfg' plus whatever else is needed, sort of how it is mimic'd today (missing the copy).

dzickusrh:
This also avoids any changes to the cfg file during a run.

major:
That does make some sense, and I'd like to keep `cfg` immutable. We wouldn't change a user's command line input or their config file, so we shouldn't touch those variables either. The `load_config()` method makes this a little tricky.

I'll try to get a look at this tomorrow.

PR #160, opened 2018-05-29
Move KernelTree into its own file, add tests

major:
This PR moves the `KernelTree` class into its own file and adds tests.


skt/__init__.py
```diff
@@ -11,15 +11,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-
+"""Class for working with the kernel source tree."""
```
spbnick: This doesn't seem to be the right place for this comment.

spbnick: Sorry, missed the button, should've gone into the review.

.circleci/config.yml
```diff
@@ -5,11 +5,13 @@ defaults: &defaults
     - run:
         name: Install skt using pip
         command: pip install .
+    # Disable R0801 in pylint that checks for duplicate content in multiple
```
spbnick: I think this would've been clearer, if it said something like "Run pylint with R0801 disabled..." or "Running with R0801 disabled in pylint...". As the main thing here is that we actually run pylint, and not just disabling stuff.

Although a little confusing, this can go in as is.

major:
@spbnick Okay, those issues that you noted are resolved.

PR #161, opened 2018-05-29
Add coverage % to README

major:


spbnick:
Do you think it would be good to move the URLs to the references at the end of file, perhaps? This way readers of the raw README.md don't have to see as much garbage at the start. Something like what's done here: https://raw.githubusercontent.com/DIGImend/digimend-kernel-drivers/master/README.md ?

danghai:
I think it is good to add badge of build/pass `travis`  too?

major:
Oh, I think i see what we're saying here. I'll take another look and push an update this morning.

major:
Oh, I think i see what we're saying here. I'll take another look and push an update this morning.

PR #172, opened 2018-06-04
Allow the shell to expand the rc file path

major:

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -668,8 +668,11 @@ def load_config(args):
     Returns:
         Loaded configuration dictionary.
     """
+    # NOTE(mhayden): The shell should do any tilde expansions on the path
+    # before the rc path is provided to Python.
     config = ConfigParser.ConfigParser()
-    config.read(os.path.expanduser(args.rc))
+    config.read(os.path.abspath(args.rc))
+
     cfg = vars(args)
     cfg['_parser'] = config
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -760,7 +760,7 @@ def load_config(args):
         Loaded configuration dictionary.
     """
     config = ConfigParser.ConfigParser()
-    config.read(os.path.expanduser(args.rc))
+    config.read(full_path(args.rc))
```
spbnick: Doing tilde expansion to command line arguments is wrong, that should be done by shell. This instead should be a call to `os.path.abspath`.

major: But we're doing that for all other command line arguments below, including the workdir, state file, and others. Should all of those be changed?

spbnick: Yes, we shouldn't do tilde expansion on command-line arguments, and we should remove it from there. Otherwise we will end up with double or undesired expansion (on filenames which actually contain tilde). We can do it on parameters from the configuration file, although we should then provide an escaping mechanism to prevent it, if necessary.

Python does basic tilde expansion alright, it seems, but it's not the point. The point is that shell already does its own expansion.

major:
@spbnick Let me know if this latest revision works for you.

spbnick:
Looks good, thanks!

PR #174, opened 2018-06-05
KernelBuilder: Tests for prepare()

danghai:
I make PR to create test `prepare()` for `KernelBuilder`.


major:
@danghai I think I have this done in #175 after making the changes there for `make rh-configs`.

PR #175, opened 2018-06-05
KernelBuilder: Add support for `make rh-configs` and `make tinyconfig`

major:
This PR builds support into KernelBuilder for handling Red Hat kernel configs as well as `tinyconfig` for very quick build tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,9 +33,8 @@ defaults: &defaults
     - run:
         name: Test `skt build`
         command: |
-            make -C workdir tinyconfig
-            mv workdir/.config config
-            skt -vv --state -d workdir --rc sktrc build -c config
+            skt -vv --state -d workdir --rc sktrc build -c config \
+                --cfgtype tinyconfig
         working_directory: /tmp/skt_test
     - store_artifacts:
         path: /tmp/skt_test/workdir/*.log
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -77,7 +77,15 @@ class KernelBuilder(object):
 
     def prepare_kernel_config(self):
         """Prepare the kernel config for the compile."""
-        shutil.copyfile(self.basecfg, "%s/.config" % self.source_dir)
+        if self.cfgtype == 'rh-configs':
+            # Build Red Hat configs and copy the correct one into place
+            self.make_redhat_config()
+        else:
+            # Copy the existing config file into place
+            shutil.copyfile(self.basecfg, "%s/.config" % self.source_dir)
+            args = self.make_argv_base + [self.cfgtype]
+            logging.info("prepare config: %s", args)
+            subprocess.check_call(args)
 
         # NOTE(mhayden): Building kernels with debuginfo can increase the
         # final kernel tarball size by 3-4x and can increase build time
@@ -87,14 +95,10 @@ class KernelBuilder(object):
         if not self.enable_debuginfo:
             self.adjust_config_option('disable', 'debug_info')
 
-        args = self.make_argv_base + [self.cfgtype]
-        logging.info("prepare config: %s", args)
-        subprocess.check_call(args)
         self._ready = 1
 
     def make_redhat_config(self):
         """Prepare the Red Hat kernel config files."""
-        # These configs build into /redhat/configs
         args = self.make_argv_base + ['rh-configs']
         logging.info("building Red Hat configs: %s", args)
         subprocess.check_call(args)
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -12,6 +12,7 @@
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 """Class for building kernels"""
+import glob
 import io
 import logging
 import multiprocessing
@@ -70,6 +71,10 @@ class KernelBuilder(object):
         logging.info("cleaning up tree: %s", args)
         subprocess.check_call(args)
 
+    def glob_escape(self, pathname):
+        """Escape any wildcard/glob characters in pathname."""
+        return re.sub("[]*?[]", "[\g<0>]", pathname)
+
     def prepare_kernel_config(self):
         """Prepare the kernel config for the compile."""
         shutil.copyfile(self.basecfg, "%s/.config" % self.source_dir)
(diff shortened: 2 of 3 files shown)


major:
Quay is having issues right now and that's causing Circle-CI tests to fail.

skt/kernelbuilder.py
```diff
@@ -86,6 +87,24 @@ def prepare(self):
...
+        """Prepare the Red Hat kernel config files."""
+        # These configs build into /redhat/configs
+        args = self.make_argv_base + ['rh-configs']
+        logging.info("building Red Hat configs: %s", args)
+        subprocess.check_call(args)
+
+        # Copy the correct kernel config into place
+        # TODO Add multi-arch support later.
+        config_files = glob.glob(
+            "{}/redhat/configs/kernel*x86_64.config".format(self.source_dir)
```
spbnick: Technically, we need to make sure that `self.source_dir` doesn't have any glob characters. From the `glob` module documentation, the following might work:

    config_files = glob.glob(
        "{}/redhat/configs/kernel*x86_64.config".format(
            re.sub("[]*?[]", "[\g<0>]", self.source_dir)
        )
    )

Please do that or similar escaping.

spbnick: Also, how would `x86_64` play with our multi-arch plan?

skt/kernelbuilder.py
```diff
@@ -97,7 +98,7 @@ def make_redhat_config(self):
         # Copy the correct kernel config into place
         # TODO Add multi-arch support later.
         config_files = glob.glob(
-            "{}/redhat/configs/kernel*x86_64.config".format(self.source_dir)
+            "{}/configs/kernel*x86_64.config".format(self.source_dir)
```
spbnick: This doesn't seem related to renaming the function. Please do this in another, appropriate commit.

skt/kernelbuilder.py
```diff
@@ -71,7 +71,8 @@ def clean_kernel_source(self):
         logging.info("cleaning up tree: %s", args)
         subprocess.check_call(args)
 
-    def prepare(self):
+    def prepare_kernel_config(self):
```
spbnick: It would be nice give a reason or intent for the rename in the commit message.

Not necessary for an ACK from me, but I would be glad to see that in the future.

skt/kernelbuilder.py
```diff
@@ -73,7 +73,15 @@ def clean_kernel_source(self):
...
-        shutil.copyfile(self.basecfg, "%s/.config" % self.source_dir)
+        if self.cfgtype == 'rh-configs':
+            # Build Red Hat configs and copy the correct one into place
+            self.make_redhat_config()
+        else:
+            # Copy the existing config file into place
+            shutil.copyfile(self.basecfg, "%s/.config" % self.source_dir)
+            args = self.make_argv_base + [self.cfgtype]
+            logging.info("prepare config: %s", args)
+            subprocess.check_call(args)
```
spbnick: It would be nice to the reviewers to have a little explanation of what it means to handle rh-configs properly, in the commit message.

Not necessary for an ACK from me, but I would be glad to see that in the future.

skt/kernelbuilder.py
```diff
@@ -73,7 +73,15 @@ def clean_kernel_source(self):
 
     def prepare_kernel_config(self):
         """Prepare the kernel config for the compile."""
-        shutil.copyfile(self.basecfg, "%s/.config" % self.source_dir)
+        if self.cfgtype == 'rh-configs':
+            # Build Red Hat configs and copy the correct one into place
+            self.make_redhat_config()
```
spbnick: It would be good to have documentation explaining what config types we have and noting which use the base configs.

If you can't think of somewhere to put them right now quickly, or it's too long to write, could you perhaps post an issue?

skt/kernelbuilder.py
```diff
@@ -83,14 +91,10 @@ def prepare_kernel_config(self):
...
             self.adjust_config_option('disable', 'debug_info')
 
-        args = self.make_argv_base + [self.cfgtype]
-        logging.info("prepare config: %s", args)
-        subprocess.check_call(args)
         self._ready = 1
 
     def make_redhat_config(self):
         """Prepare the Red Hat kernel config files."""
-        # These configs build into /redhat/configs
```
spbnick: Hmm, why are we removing this comment? I'm just confused.

tests/test_kernelbuilder.py
```diff
@@ -180,10 +179,9 @@ def test_getrelease(self):
             result = self.kbuilder.getrelease()
             self.assertEqual(kernel_version.strip(), result)
 
-    def test_getrelease_regex_fail(self):
-        """Ensure get_release() fails if the regex doesn't match"""
-        mock_prepare = mock.patch("skt.kernelbuilder.KernelBuilder.prepare")
-
+    @mock.patch("skt.kernelbuilder.KernelBuilder.prepare_kernel_config")
+    def test_getrelease_regex_fail(self, mock_prepare):
+        """Ensure get_release() fails if the regex doesn't match."""
```
spbnick: These fixes need to go in the same commit which introduces the breaking change, so there is no commit where tests are broken. Please move appropriately.

skt/kernelbuilder.py
```diff
@@ -76,6 +76,9 @@ def prepare_kernel_config(self):
         if self.cfgtype == 'rh-configs':
             # Build Red Hat configs and copy the correct one into place
             self.make_redhat_config()
+        elif self.cfgtype == 'tinyconfig':
+            # Build an extremely small config file for quick testing
+            self.make_tinyconfig()
```
spbnick: Ah, this is a nice one!

tests/test_kernelbuilder.py
```diff
@@ -188,3 +188,37 @@ def test_getrelease_regex_fail(self, mock_prepare):
         with mock_popen, mock_prepare:
             with self.assertRaises(Exception):
                 self.kbuilder.getrelease()
+
+    @mock.patch("skt.kernelbuilder.KernelBuilder.adjust_config_option")
```
spbnick: Comment on the commit message: Please use imperative mood: write "Add tests ..." (or similar), not just "New tests ...". This way it's easier to reason about the commit when reverting/merging. Thanks!

.circleci/config.yml
```diff
@@ -33,9 +33,8 @@ defaults: &defaults
     - run:
         name: Test `skt build`
         command: |
-            make -C workdir tinyconfig
-            mv workdir/.config config
-            skt -vv --state -d workdir --rc sktrc build -c config
+            skt -vv --state -d workdir --rc sktrc build -c config \
+                --cfgtype tinyconfig
```
spbnick: Nice!

spbnick:
If we want to further speed this up, I will be OK with just the glob escaping added, and the answer to the multi-arch question. It would be nice to have the documentation issue posted, though.

.circleci/config.yml
major: Yay!

tests/test_kernelbuilder.py
major: Fixed.

skt/kernelbuilder.py
major: Done.

tests/test_kernelbuilder.py
major: Fixed.

skt/kernelbuilder.py
major: Support for multi-arch would come later. This does make it a lot easier to select the correct one, though.

major:
@spbnick I can't think of a situation why we should accept a workdir/sourcedir that contains a wildcard. This seems like a situation where we should detect those glob characters and abort the build.

PR #176, opened 2018-06-05
Travis-CI: Enable branch testing

major:
This PR ensures we are testing any branches found in the code.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ deps =
     mock
     coveralls
 commands =
-    coverage run --source=skt,tests -m unittest discover tests
+    coverage run --branch --source=skt,tests -m unittest discover tests
     coverage report -m
     coveralls
 


tox.ini
```diff
@@ -11,8 +11,8 @@ deps =
     mock
     coveralls
 commands =
-    coverage run --source=skt,tests -m unittest discover tests
-    coverage report -m
+    coverage run --branch --source=sktm,tests --omit="*/tests/*" -m pytest -v -s tests
```
danghai: I think  you have a typo here. It should be `skt` not `sktm`

major: Oh, good catch.

major:
Coveralls is upset because test coverage dropped with branch testing enabled. That's expected.

danghai:
I think I mentioned it in this PR: https://github.com/RH-FMK/skt/pull/83

major:
@spbnick Well, we're not using any `pytest`-specific features yet, so it will continue running unit tests in a similar way to `unittest`. However, it does offer us some handy capabilities around test fixtures and faked filesystems that will be handy as we increase test coverage.

spbnick:
Well, extra features are nice, yes. What I'm trying to avoid is pytest hijacking the assert statement and changing its function when running tests. If it does that, then I would prefer not to use it, as I would like to use `assert` for its original purpose.

major:
@spbnick I've used `pytest` several times and I've not ever seen it do that. It only runs the unit tests found in the directory specified unless you start adding in `pytest` specific features.

major:
@spbnick I've adjusted the patch to leave `unittest` alone and simply enable branch testing. Coveralls will complain that coverage is reduced, but that's expected.

PR #178, opened 2018-06-06
KernelBuilder: Detect cross compiling

### LINKED ISSUE #177: Multi-arch: skt must know which config to use with `make rh-configs`
When `skt` runs `make rh-configs` as part of `KernelBuilder.make_redhat_config()`, it needs to know what architecture is being built so that it can select the correct configuration file for the build. There are two possible situations:

**The kernel is being cross compiled for a different architecture.** In this case, we must detect the `ARCH` environment variable and use that value to select the correct configuration file.

**The kernel is being built in the machine's native architecture.** The `ARCH` environment variable will not be set and the output of `uname -m` (or equivalent python) must be used.

major:
This PR adds support to `skt` for detecting a cross compile and choosing an appropriate Red Hat kernel configuration file.

Fixes #177.

### DIFF
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -17,6 +17,7 @@ import io
 import logging
 import multiprocessing
 import os
+import platform
 import re
 import shlex
 import shutil
@@ -36,6 +37,7 @@ class KernelBuilder(object):
         self.buildlog = "%s/build.log" % self.source_dir
         self.make_argv_base = ["make", "-C", self.source_dir]
         self.enable_debuginfo = enable_debuginfo
+        self.build_arch = self.get_build_arch()
 
         # Split the extra make arguments provided by the user
         if extra_make_args:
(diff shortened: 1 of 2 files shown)


skt/kernelbuilder.py
```diff
@@ -124,6 +129,15 @@ def make_tinyconfig(self):
         logging.info("building tinyconfig: %s", args)
         subprocess.check_call(args)
 
+    def get_build_arch(self):
+        """Determine the build architecture for the kernel build."""
+        # Detect an architecture specified for the kernel build (for cross
+        # compiling kernels) and set the build arch
```
veruu: The function itself isn't setting anything, so this part of the comment is confusing. Can you mention cross-compilation in the docstring and remove the additional multiline comment instead?

major: Sure.

major:
@veruu I also removed the `TODO` from make_redhat_configs(). ;)

PR #182, opened 2018-06-06
CircleCI: Remove unit/lint testing

### LINKED ISSUE #181: Remove unit/lint tests from CircleCI
CircleCI is great for functional testing since we get access to more CPU cores for a quick kernel build and we can run within Docker containers there easily. However, we're doing lint/unit testing in both Travis-CI and CircleCI, which is wasteful.

We should limit our testing in CircleCI to functional testing only.

major:
We are using CircleCI for functional testing since it's easier to use
Docker containers there for real kernel builds. It is not necessary
to duplicate the same tests that we run in Travis-CI.

Fixes #181.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,14 +5,6 @@ defaults: &defaults
     - run:
         name: Install skt using pip
         command: pip install .
-    # Run pylint without R0801 that checks for duplicate content in multiple
-    # files. See https://github.com/PyCQA/pylint/issues/214 for details.
-    - run:
-        name: Run Python linters and tests
-        command: |
-          flake8 --show-source .
-          pylint -d R0801 tests
-          python -m unittest discover tests
     # TODO(mhayden): Add --depth support to skt and then let skt to the clone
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,16 +5,6 @@ defaults: &defaults
...
         command: pip install .
-    # Run pylint without R0801 that checks for duplicate content in multiple
-    # files. See https://github.com/PyCQA/pylint/issues/214 for details.
-    - run:
-        name: Run Python linters and tests
-        command: |
-          flake8 --show-source .
-          pylint -d R0801 tests
-          python -m unittest discover tests
-    # TODO(mhayden): Add --depth support to skt and then let skt to the clone
```
veruu: The removed TODO seems unrelated to "not run linters" commit. If it's being removed because it was already fixed (I assume it is, since we are using the `--fetch-depth` option below), can you do it in separate commit please?

major: Fixed!

PR #184, opened 2018-06-07
Fix kernel config issues in reporter

veruu:
The name of the attachment changed from 'config' to 'config.gz' as we
decided to compress the data, which wasn't reflected when working with
the attachment later. That said, printing gzip-ed data in StdiioReporter
is fairly useless too, so remove that.

Add a comment to explain moving config attachment to the end of
attachment list, and iterate over the list of attachments instead of
checking only the first one, in case the order of adding attachments
changes.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -458,8 +458,15 @@ class Reporter(object):
                 '\nSincerely,',
                 '  Kernel CI Team']
 
-        if self.attach and self.attach[0][0] == "config":
-            self.attach.append(self.attach.pop(0))
+        # Move configuration attachments to the end because some mail clients
+        # (eg. mutt) inline them and they are huge
+        # It's not safe to iterate over changing list so let's use a helper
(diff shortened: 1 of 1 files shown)


skt/reporter.py
```diff
@@ -458,8 +458,15 @@ def getreport(self):
                 '\nSincerely,',
                 '  Kernel CI Team']
 
-        if self.attach and self.attach[0][0] == "config":
-            self.attach.append(self.attach.pop(0))
+        # Move configuration attachments to the end because some mail clients
+        # (eg. mutt) inline them and they are huge
+        # It's not safe to iterate over changing list so let's use a helper
+        config_attachments = [attachment for attachment in self.attach
```
major: Nice idea. It's easy to follow. ;)

major:
@veruu Can you rebase this one? It looks good otherwise.

veruu:
@major rebased

PR #186, opened 2018-06-08
Arch support

dzickusrh:
Add arch support for multi-arch beaker jobs

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -428,7 +428,7 @@ class BeakerRunner(Runner):
 
         return jobid
 
-    def run(self, url, release, wait=False, host=None, uid="",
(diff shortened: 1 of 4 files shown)


skt/runner.py
```diff
@@ -428,7 +428,7 @@ def jobsubmit(self, xml):
 
         return jobid
 
-    def run(self, url, release, wait=False, host=None, uid="",
+    def run(self, url, release, wait=False, host=None, uid="", arch="x86_64",
```
major: I'm not sure if hard-coding `x86_64` is a good idea long term, but it should work in the short term. We could check to see if it's set, and if not, detect it with `platform.machine()`.

PR #187, opened 2018-06-08
Run tests

dzickusrh:
Add more tests for skt/runner.py


major:
@dzickusrh You had a pylint failure:

```
************* Module tests.test_runner
R: 35, 0: Too many public methods (22/20) (too-many-public-methods)
```
You can do something similar to what we did here:

* https://github.com/RH-FMK/skt/blob/da8ada6152d726266c4f8bbc781e4758477b7201/tests/test_kerneltree.py#L35

PR #198, opened 2018-06-12
Use shutil.move to move files, not os.rename

spbnick:
Switch to using shutil.move instead of os.rename to move files in
cmd_build. The os.rename uses the rename syscall, which only works
within a single filesystem, failing with EXDEV otherwise. Moving build
artifacts across filesystems is currently necessary for the Docker
build.


major:
@spbnick Rebase when you can and I will merge it.

PR #204, opened 2018-06-12
Remove tests directory from coverage

### LINKED ISSUE #203: Test coverage is inflated since tests are included
Our test coverage percentage is inflated because the `tests` directory is included when calculating coverage.

major:
The test coverage percentage is higher than it should be. This is
because the tests themselves are being included in the results.

Fixes #203.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ deps =
     mock
     coveralls
 commands =
-    coverage run --branch --source=skt,tests -m unittest discover tests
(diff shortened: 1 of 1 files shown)


veruu:
Isn't it enough to remove tests from `--source` in `coverage run` command so the test coverage isn't even calculated for them, instead of removing them from the report only?

major:
@veruu Ah, I didn't realize that would work. I'll try that.

PR #206, opened 2018-06-12
junit:  Create directory if it doesn't exist

dzickusrh:
This day and age, we shouldn't expect the user to create a directory
ahead of time.  Try and create it early on.

This also fixes the problem of waiting until the end of the command to find
out that the user forgot that simple step.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -793,6 +793,12 @@ def load_config(args):
     for idx, statefile_path in enumerate(cfg.get('states') or []):
         cfg['states'][idx] = full_path(statefile_path)
 
+    if cfg.get('junit'):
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -793,6 +793,12 @@ def load_config(args):
     for idx, statefile_path in enumerate(cfg.get('states') or []):
         cfg['states'][idx] = full_path(statefile_path)
 
+    if cfg.get('junit'):
+        try:
+            os.mkdir(cfg.get('junit'))
```
major: Could you wrap this in `full_path()` so we get an absolute path to the directory?

PR #210, opened 2018-06-12
Set target

dzickusrh:
In order to merge report results, we need to create an arch dict() of results.  This PR achieves that.
Now we can grab all the results and merge them in one dict() and process them in one go.

### DIFF
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -27,10 +27,37 @@ import time
 from threading import Timer
 
 
+def verify_target(target):
+    """
+    Verify the given target is something we support.
+
+    Args:
+        target:    A build target archtecture
+
+    Returns:
+        The value of the ARCH env and the config file variant as a tuple
+    """
+    if not target:
+        target = platform.machine()
+
+    if re.match('x86_64', target):
+        return ('x86_64', target)
+    elif re.match('s390x', target):
+        return ('s390', target)
+    elif re.match('ppc64', target):
+        return ('powerpc', target)
+    elif re.match('ppc64le', target):
+        return ('powerpc', target)
+    elif re.match('aarch64', target):
+        return ('arm64', target)
+
+    return (None, None)
+
+
 class KernelBuilder(object):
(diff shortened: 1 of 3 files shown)


skt/kernelbuilder.py
```diff
@@ -27,19 +27,53 @@
 from threading import Timer
 
 
+def verify_target(target):
```
agustinhenze: The name verify, implies raising an exception or return True or False to me. It's something more like translate_target and judging for the implementation a dict satisfies the necessity very well instead of matching word by word.

skt/kernelbuilder.py
```diff
@@ -27,19 +27,53 @@
...
+    Verify the given target is something we support.
+
+    Args:
+        target:    A build target archtecture
+
+    Returns:
+        The value of the ARCH env and the config file variant as a tuple
+    """
+    if not target:
+        target = platform.machine()
```
agustinhenze:
```python
target = target or platform.machine()
```
not important at all, just extra sugar

skt/kernelbuilder.py
```diff
@@ -27,19 +27,53 @@
...
         self.buildlog = "%s/build.log" % self.source_dir
         self.make_argv_base = ["make", "-C", self.source_dir]
         self.enable_debuginfo = enable_debuginfo
-        self.build_arch = self.get_build_arch()
-        self.rh_configs_glob = rh_configs_glob
+        self.target = target
+
+        (build_arch, config_arch) = verify_target(target)
+        if build_arch is None:
+            raise Exception("Unsupported target: %s" % target)
```
agustinhenze: Raising a generic exception and not a custom one makes harder catch it later. It'd be more readable if you create a custom exception, e.g.
```python
class UnsupportedTarget(Exception):
   "Raised when target is not supported"
```

agustinhenze: Also, I think it'd be better if you raise the exception inside the function to avoid repeated code.

PR #212, opened 2018-06-13
Add support for remote smtp url

dzickusrh:
Some of us suck at setting up a local smtp server.  Make
it easy by allowing one to add an entry in sktrc to point
to a remote server.  Use the 'smtp_url' entry.

    [config]
    smtp_url = smtp.example.com

Otherwise if the above is not defined (common case), it defaults
to 'localhost'.

Only used by the 'reporter' command.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -690,6 +690,8 @@ class MailReporter(Reporter):
         self.headers = [headers.strip() for headers in
                         cfg['reporter']['mail_header']]
         self.subject = cfg['reporter']['mail_subject']
+        self.smtp_url = cfg.get('smtp_url') or 'localhost'
+
         super(MailReporter, self).__init__(cfg)
 
     def report(self):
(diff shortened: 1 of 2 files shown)


veruu:
This is a good idea. Do you think you can extend the patch so the option can be used on command line too?

skt/reporter.py
```diff
@@ -688,6 +688,10 @@ def __init__(self, cfg):
         self.headers = [headers.strip() for headers in
                         cfg['reporter']['mail_headers']]
         self.subject = cfg['reporter']['mail_subject']
+        self.smtp_url = 'localhost'
+        if cfg.get('smtp_url'):
+            self.smtp_url = cfg.get('smtp_url')
```
veruu: Unless I'm missing overrides somewhere, `self.smtp_url = cfg.get('smtp_url', 'localhost')` can be used instead of those 3 lines.

dzickusrh: Yeah that didn't work.  I implemented the parser arg like you requested and it auto-sets it to None, which means the 'localhost' will never be used.  Setting it as a default in the parser blocks the config file from overriding it. :-/
Instead I am using: self.smtp_url = cfg.get('smtp_url') or 'localhost'
which maintains the spirit of what you wanted. :-)

veruu: Right, those two ideas don't play well together. This was just a shorthand in case you didn't want to do the arg :) Ternary operators for assignments work just as fine. Will check the changes tomorrow!

PR #214, opened 2018-06-13
WIP: Add stub code to test general workflow with dummy data

dzickusrh:
Testing the tools is hard when you need a lot of pieces to run.  Further
the amount of time it takes for those pieces to complete is long.

Short-circuit it, by quickly return the success/failure you want.

WIP because it is quick and dirty way of updating the state fail and
properly generating the expected report response.

Use:

skt --state --test 0 merge   # generates a successful merge in state data
skt --state --test 1 merge   # generates a failed merge in state data
skt --state report  # will report on the stub'd state data

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -141,6 +141,20 @@ def cmd_merge(cfg):
         wdir=cfg.get('workdir'),
         fetch_depth=cfg.get('fetch_depth')
     )
+
+    # setup stub code for quick testing
+    test = cfg.get('test')
+    if test is not None:
+        if test == 0:
+            save_state(cfg, {'baserepo': cfg.get('baserepo'),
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -141,6 +141,20 @@ def cmd_merge(cfg):
         wdir=cfg.get('workdir'),
         fetch_depth=cfg.get('fetch_depth')
     )
+
+    # setup stub code for quick testing
+    test = cfg.get('test')
+    if test is not None:
+        if test == 0:
+            save_state(cfg, {'baserepo': cfg.get('baserepo'),
```
spbnick: Hmm, do we have to set those to the same value here and everywhere else?

PR #217, opened 2018-06-13
Test coverage for init, reporter

major:

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -51,6 +51,65 @@ def gzipdata(data):
     return tstr.getvalue()
 
 
+def load_state_cfg(statefile):
+    """Load state information from a state file.
+
+    It takes the current config and adjusts state information within the
+    config based on the state file provided.
+
+    Args:
+        statefile:      Path to a skt state file to read
+
+    Returns: A cfg dictionary.
+
+    """
+    cfg = {}
+    state_to_report = ConfigParser.ConfigParser()
+    state_to_report.read(statefile)
+
+    # FIXME This can be simplified or removed after configuration and
+    # state split
+    for (name, value) in state_to_report.items('state'):
+        if not cfg.get(name):
+            if name.startswith('jobid_'):
+                if 'jobs' not in cfg:
+                    cfg['jobs'] = set()
+                cfg['jobs'].add(value)
+            elif name.startswith('mergerepo_'):
+                if 'mergerepos' not in cfg:
+                    cfg['mergerepos'] = list()
+                cfg['mergerepos'].append(value)
+            elif name.startswith('mergehead_'):
+                if 'mergeheads' not in cfg:
+                    cfg['mergeheads'] = list()
+                cfg['mergeheads'].append(value)
+            elif name.startswith('localpatch_'):
+                if 'localpatches' not in cfg:
+                    cfg['localpatches'] = list()
+                cfg['localpatches'].append(value)
+            elif name.startswith('patchwork_'):
+                if 'patchworks' not in cfg:
+                    cfg['patchworks'] = list()
+                cfg['patchworks'].append(value)
+            cfg[name] = value
+
+    # Get runner info
+    if state_to_report.has_section('runner'):
+        runner_config = {}
+        for (key, val) in state_to_report.items('runner'):
+            if key != 'type':
+                runner_config[key] = val
+            cfg['runner'] = [
+                state_to_report.get('runner', 'type'),
+                runner_config
+            ]
+    else:
+        logging.debug('No runner info found in state file, test runs will'
+                      ' not be reported')
+
+    return cfg
+
+
 class ConsoleLog(object):
     """Console log parser"""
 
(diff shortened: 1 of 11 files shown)


tests/test_reporter.py
```diff
@@ -99,3 +124,171 @@ def test_match_three_traces(self):
...
+
+
+class TestReporter(unittest.TestCase):
+    """Test cases for Reporter class."""
+
+    def setUp(self):
+        """Set up test fixtures."""
+        # Write a kernel .config file
+        self.tmpdir = tempfile.mkdtemp()
+        self.tempconfig = "{}/.config".format(self.tmpdir)
```
veruu: Can we have the config created only if the merge passed, or would it be too complicated? That way we can check merge failures don't require it for reports. Eventually, we should be grabbing config only if the build was attempted, but we need to implement that first.

skt/reporter.py
```diff
@@ -294,6 +294,7 @@ def update_mergedata(self):
             else:
                 with open("%s/.config" % self.cfg.get("workdir"),
                           "r") as fileh:
+                    print("%s/.config" % self.cfg.get("workdir"))
```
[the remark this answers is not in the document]
major: Darn! I thought I had removed all of those. :/

tests/test_reporter.py
major: I'll just delete the temporary config file before I test for the merge failure.

tests/assets/beaker_job_fail_results_full.xml
```diff
@@ -0,0 +1,175 @@
+<job id="2547021" owner="someone@example" result="Fail" retention_tag="scratch" status="Completed">
+	<whiteboard>skt 4.17.0-rc1+ [patchwork] e6fbae817eec812bf6686c131bb8452d21c83504.tar.gz [noavc] [noselinux]</whiteboard>
+	<recipeSet id="4243047" priority="Normal" response="ack">
+		<recipe arch="x86_64" distro="Fedora-26" duration="0:14:56" family="Fedora26" finish_time="2018-06-15 10:57:25" id="5273166" job_id="2547021" kernel_options="selinux=0" kernel_options_post="" kickstart_url="http://beaker.example.com/kickstart/4336389" recipe_set_id="4243047" result="Fail" role="None" start_time="2018-06-15 10:42:29" status="Completed" system="hp-fail.example.com" variant="Server" whiteboard="4.17.0-rc1+">
```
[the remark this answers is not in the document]
major: Ah, I was testing something. I'll fix that.

tests/assets/beaker_job_pass_results_full.xml
```diff
@@ -0,0 +1,175 @@
+<job id="2547022" owner="someone@example" result="Pass" retention_tag="scratch" status="Completed">
+	<whiteboard>skt 4.17.0-rc1+ [patchwork] e6fbae817eec812bf6686c131bb8452d21c83504.tar.gz [noavc] [noselinux]</whiteboard>
+	<recipeSet id="4243047" priority="Normal" response="ack">
+		<recipe arch="x86_64" distro="Fedora-26" duration="0:14:56" family="Fedora26" finish_time="2018-06-15 10:57:25" id="5273166" job_id="2547022" kernel_options="selinux=0" kernel_options_post="" kickstart_url="http://beaker.example.com/kickstart/4336389" recipe_set_id="4243047" result="Pass" role="None" start_time="2018-06-15 10:42:29" status="Completed" system="hp-pass.example.com" variant="Server" whiteboard="4.17.0-rc1+">
```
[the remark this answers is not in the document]
major: Ah, I was testing something. I'll fix that.

major:
@veruu I forgot to commit the change to remove the config file before. It's here now: https://github.com/RH-FMK/skt/pull/217/files#diff-220ba5bc7cb2e37572c72a4f464eeb44R223

veruu:
@major , shouldn't the kernel config file be removed in `test_merge_failure_empty_log`test too? Otherwise looks good :)

major:
Good catch, @veruu. Fixing.

skt/reporter.py
```diff
@@ -513,48 +513,50 @@ def load_state_cfg(self, statefile):
...
         state_to_report = ConfigParser.ConfigParser()
         state_to_report.read(statefile)
 
         # FIXME This can be simplified or removed after configuration and
         # state split
         for (name, value) in state_to_report.items('state'):
-            if not self.cfg.get(name):
+            if not cfg.get(name):
                 if name.startswith('jobid_'):
                     if 'jobs' not in self.cfg:
```
veruu: I'm totally kicking myself for not noticing it sooner... but since you are changing the loading from `self.cfg` to `cfg`, it should be changed here (and on a bunch of other lines) too. Otherwise the reporting breaks. I think we need a test for `load_state_cfg` as well :D

major: Ah, the docstring is broken, too. I'll go fix those.

major:
83% coverage for `reporter.py`. ;)

skt/reporter.py
```diff
@@ -51,6 +51,67 @@ def gzipdata(data):
...
+
+    It takes the current config and adjusts state information within the
+    config based on the state file provided.
+
+    Args:
+        current_cfg:    Existing cfg dict to adjust
+        statefile:      Path to a skt state file to read
+
+    Returns: A cfg dictionary.
+    Raises: Exception if required 'runner' section is missing
```
veruu: There should be no exception raised.

skt/reporter.py
```diff
@@ -51,6 +51,67 @@ def gzipdata(data):
...
+
+    """
+    cfg = current_cfg
```
veruu: I don't think this is a good idea. We want to read a clean new state each time, not add it to previous ones. Checking out reporting with more state files to report, I get the first file passed reported multiple times, instead of having each of them reported once.

tests/test_reporter.py
```diff
@@ -14,34 +11,116 @@
...
+
+    def setUp(self):
+        """Test fixtures."""
+        self.statefile = "{}/assets/testing_state.cfg".format(SCRIPT_PATH)
+        self.statefile_no_runner = (
+            "{}/assets/testing_state_no_runner.cfg".format(SCRIPT_PATH)
+        )
+
+    def test_load_state_cfg(self):
+        """Test load_state_cfg() with an empty cfg."""
```
veruu: What does `with an empty cfg` mean here? Is it just a residue of previous implementation?

major: Ah, I meant to remove that.

PR #221, opened 2018-06-14
Reporting fixes

veruu:


skt/reporter.py
```diff
@@ -694,7 +694,8 @@ def report(self):
         msg = MIMEMultipart()
 
         # Add the most basic parts of the email message
-        msg['Subject'] = self.subject
+        if self.subject:
+            msg['Subject'] = self.subject
```
major: Should we remove the `default` option here in `executable.py`?

https://github.com/RH-FMK/skt/blob/master/skt/executable.py#L631-L636

veruu: Simply removing the default doesn't fix since then a header with `None` as a value is created, but I agree that the default doesn't make much sense here. I'll remove that.

PR #226, opened 2018-06-15
Merge fixes

spbnick:
These are various fixes for merging issues encountered during merging in containers. We need these in ASAP, but please take a careful look through them. Thanks!

### DIFF
--- a/skt/kerneltree.py
+++ b/skt/kerneltree.py
@@ -300,7 +300,12 @@ class KernelTree(object):
         retcode = gam.wait()
 
         if retcode != 0:
-            self.git_cmd("am", "--abort", cwd=self.wdir)
+            try:
+                self.git_cmd("am", "--abort", cwd=self.wdir)
+            except subprocess.CalledProcessError as exc:
+                logging.debug("%s failed with status %d and "
+                              "following output:\n%s",
+                              exc.cmd, exc.returncode, exc.output)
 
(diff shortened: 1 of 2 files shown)


spbnick:
Argh, completely missed your review. How would you prefer this done?

spbnick:
Merging now, but let's discuss the proper approach/report issues to refactor this. Thank you, Veronika!

veruu:
I'd have to take a proper look at the containers and things around to see how it can be simplified, if `git am --abort` failing doesn't mean a serious problem which shouldn't be caught etc. It's not a huge problem that would need immediate solving, I just find nesting like this slightly unclear.

spbnick:
I see, thank you the explanation. Yeah, I think we should abort completely, if `git am --abort` fails, because that would indicate a serious problem, likely breaking Git repo consistency. We shouldn't just e.g. go on applying patches. I didn't think this through well enough. However, we should preserve information on the original failure too, even if `git am --abort` fails.

spbnick:
Now reading the above, I don't think I'm thinking well enough today :)

veruu:
Sounds good! Let's save the failure first, and then run the `--abort`. If that fails, we already have all the info and can safely exit. Can you file an issue for it, so we don't forget to take a look later?

spbnick:
Alright, my rather lame attempt at it is in #229 :smiley:

PR #232, opened 2018-06-15
Use jinja2 for sending reports

major:
Merge single and multireports into one function and render the entire
email message with jinja2. This makes it easier to see what will end up
in the email sent to developers.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -21,13 +21,16 @@ import logging
 import os
 import re
 import smtplib
+import sys
 import StringIO
 
+from jinja2 import Environment, FileSystemLoader
 import requests
 
 import skt
 import skt.runner
 
+SCRIPT_PATH = os.path.dirname(os.path.realpath(__file__))
 
 MULTI_PASS = 0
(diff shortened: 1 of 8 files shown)


major:
The tests don't really examine the output yet. Still working on that.

skt/reporter.py
```diff
@@ -434,61 +374,87 @@ def getjobresults(self):
...
+
+            # If the compile failed, get the build log
+            temp_report['buildlog_output'] = None
+            if self.cfg.get('buildlog'):
+                self.getbuildfailure()
+                with open(self.cfg.get('buildlog'), 'r') as fileh:
+                    temp_report['buildlog_output'] = fileh.read()
+
+            # Get the test results
+            temp_report['jobresults'] = self.getjobresults()
```
veruu: This still causes failures for me :( In the original report, we are checking if the runner is specified before getting job results, but there's no check here so the report results in a traceback.

veruu:
Because of the planned reporting rework and move to sktm, I think we can close this one. @major what do you think?

PR #234, opened 2018-06-15
Detect local arch if it's not specified

### LINKED ISSUE #188: Detect arch instead of defaulting to x86_64
In #186, we default to `x86_64` when submitting a beaker job. We should detect the local machine arch if the arch is not specified.

https://github.com/RH-FMK/skt/blob/646f1aa3700a2d25d73a6eb1c29fe18d662ed76d/skt/runner.py#L431

veruu:
Instead of defaulting to x86_64, we should be using the architecture of
the local machine we run on.

Fixes #188

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -14,6 +14,7 @@
 
 import logging
 import os
+import platform
 import re
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


veruu:
Rebased.

PR #236, opened 2018-06-15
Add overall status to subject line of email

dzickusrh:
When reading through 10s of status emails, it is hard to detect quickly
which jobs failed.  Those are jobs I want to prioritize to verify the
pipeline is working correctly.

Stick the status in the subject line as [PASS] or [FAIL].  This requires
stripping an prepending Re:.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -630,6 +630,21 @@ class Reporter(object):
 
         return '\n'.join(intro + self.get_multireport_summary() + results)
 
+    def get_status(self):
+        if self.multireport:
+            if self.multireport_failed == MULTI_PASS and \
+               not self.multiretcode:
+                return 'PASS'
+            else:
+                return 'FAIL'
+        else:
+            if not self.cfg.get('mergelog') and \
+               not self.cfg.get('buildlog') and \
+               self.cfg.get('retcode') == '0':
+                return 'PASS'
+            else:
+                return 'FAIL'
+
     def getsubject(self):
         if not self.cfg.get('mergelog') and \
            not self.cfg.get('buildlog') and \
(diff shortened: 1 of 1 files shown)


skt/reporter.py
```diff
@@ -723,7 +738,10 @@ def report(self):
 
         # Add the most basic parts of the email message
         if self.subject:
-            msg['Subject'] = self.subject
+            # Strip all leading Re: s
+            subject = self.subject.replace('re: ', '', re.IGNORECASE)
```
veruu: Testing it with subject `Re: [PATCH] CONFIGURE: add xxx`, I'm getting `Re: [<get_status output here>] Re: [PATCH] CONFIGURE: add xxx` as new subject, which we probably don't want. This is because the normal `replace` method is not case-insensitive, and what you're specifying as the last parameter is actually the [max number of replacements](https://www.tutorialspoint.com/python/string_replace.htm). What you want for case-insensitive replacements is [`re.sub`](https://docs.python.org/3/library/re.html#re.sub). BUT, doing so with the testing subject will remove the `RE:` in `CONFIGURE:` as well, so you'd need to write a more complicated regex to only remove `re:` from before the square brackets.

spbnick: This won't catch all the possible language-specific "Re:" incarnations, such as "AW:" and "На:", but OK.

Here's the regex you need, I think:

    subject = re.sub("^ *(Re: *)*",
                     "Re: [" + self.get_status() + "] ",
                     self.subject,
                     flags=re.IGNORECASE)

veruu: We are the ones passing the subject with `Re: ` in the pipeline, so I think that part should be OK

PR #237, opened 2018-06-15
Add reporter option --dry-run 

dzickusrh:


major:
Isn't this why we have `--reporter stdio`? That prints out the same exact text as the email.

skt/executable.py
```diff
@@ -654,6 +654,12 @@ def setup_parser():
         type=str,
         help='Use smtp url instead of localhost to send mail',
     )
+    parser_report.add_argument(
```
major: Could we call this --mail-dry-run so that it lines up better with the other --mail arguments?

skt/reporter.py
```diff
@@ -748,6 +751,8 @@ def report(self):
             msg.attach(MIMEText(self.getreport()))
 
         for (name, att) in self.attach:
+            if self.dry_run:
```
veruu: Wouldn't it be better to skip checking the attachments if the dry run option is detected instead of iterating over them and skipping attaching?
```
if not self.dry_run:
    for (name, att) in self.attach:
        .....
```

PR #240, opened 2018-06-18
merge: Convert multi-value options

### LINKED ISSUE #227: Consider not using multi-value options
Options which accept multiple values are generally confusing and have surprising behavior. Command lines using both these options, and positional arguments, or other multi-value options, are hard to read. Consider not using them and instead using repeated option approach. This concerns `--patchlist`, `--pw` and `-m/--merge-ref` options.

major:
Convert multi-value options to arguments that can be repeated.
This improves readability on the command line.

Fixes #227.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## June 2018
 
+* Arguments for `skt merge` should now be used multiple times instead of
+  supplying multiple values for each option. This affects the following
+  arguments:
+
+  * `--patch` (formerly `--patchlist`)
+  * `--pw`
+  * `--merge-ref`
+
+  Example: `skt merge --pw URL1 --pw URL2 --pw URL3`
+
 * Arguments for reporter are now explicitly defined and JSON strings are no
   longer required. Users can specify the following arguments:
 
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -159,10 +159,10 @@ def cmd_merge(cfg):
             if retcode:
                 return
 
-        if cfg.get('patchlist'):
+        if cfg.get('patch'):
             utypes.append("[local patch]")
             idx = 0
-            for patch in cfg.get('patchlist'):
+            for patch in cfg.get('patch'):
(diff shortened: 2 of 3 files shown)


danghai:
@major It is nice! I tried it before. I see it will break when one of them patch cannot merge. But it does not provide which patch is wrong.

skt/executable.py
```diff
@@ -502,21 +502,29 @@ def setup_parser():
     parser_merge.add_argument(
         "--patchlist",
         type=str,
-        nargs="+",
-        help="Paths to each local patch to apply (space delimited)"
+        action='append',
+        help=(
+            "Paths to each local patch to apply "
+            "(use multiple times for multiple patches)"
```
veruu: Shouldn't the name of the option be changed too, since the list shouldn't be provided anymore?

skt/executable.py
```diff
@@ -502,21 +502,29 @@ def setup_parser():
...
+        )
     )
     parser_merge.add_argument(
         "--pw",
         type=str,
-        nargs="+",
-        help="URLs to each Patchwork patch to apply (space delimited)"
+        action='append',
+        help=(
+            "URLs to each Patchwork patch to apply"
```
veruu: Same here

skt/executable.py
major: Right. I'll fix that.

skt/executable.py
major: I'll fix this one , too.

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## June 2018
 
+* Arguments for `skt merge` can now be used multiple times instead of
```
spbnick: I would change this to "should now be used multiple times", and to "supplying multiple values".

skt/executable.py
```diff
@@ -504,21 +504,29 @@ def setup_parser():
     parser_merge.add_argument(
         "--patchlist",
```
spbnick: I think this should be changed to just "--patch" now.

PR #241, opened 2018-06-18
run: Use command line args instead of JSON strings

### LINKED ISSUE #125: Get rid of raw json / dicts on runner commandline
1. Bad UX, using explicit options is easier, cleaner and makes more sense
2. `ast.literal_eval()` requires some special escaping so passing strings with newlines (eg. patch subjects) is a real PITA, and it *shouldn't*

major:
Use multiple explicit arguments for running tests with `skt`. This
makes it easier to assemble the command line arguments for Beaker
jobs.

Fixes #125


skt/executable.py
```diff
@@ -293,7 +292,11 @@ def cmd_run(cfg):
         cfg:    A dictionary of skt configuration.
     """
     global retcode
-    runner = skt.runner.getrunner(*cfg.get('runner'))
+
+    # Attempt to import the reporter provided by the user
```
veruu: Should this be 'runner'? ;)

major: Good catch. I'll fix it.

skt/executable.py
```diff
@@ -775,15 +790,17 @@ def load_config(args):
...
+    if cfg.get('_name') == 'run' and cfg.get('type'):
+        # Use the runner configuration from the command line
+        cfg['runner'] = {
+            'type': cfg.get('type'),
+            'jobtemplate': cfg.get('jobtemplate'),
+            'jobowner': cfg.get('jobowner')
+        }
+    elif config.has_section('runner'):
+        # Use the runner configuration from the configuration file
+        cfg['runner'] = config.items('runner')
```
spbnick: Could you make the command-line options overwrite specific settings from the config file, instead of all of them? E.g. allow setting just the jobtemplate, but not jobowner?

skt/runner.py
```diff
@@ -460,21 +461,18 @@ def run(self, url, release, wait=False, host=None, uid="",
...
+        reporter_class = getattr(module, class_name)
+    except AttributeError:
+        sys.exit(
+            "Unable to find specified runner type: {}".format(class_name)
+        )
+
+    return reporter_class(
+        jobtemplate=cfg['runner']['jobtemplate'],
+        jobowner=cfg['runner']['jobowner']
+    )
```
spbnick: Could you please make this refactoring a separate commit, and also please keep documenting both arguments and return values?

spbnick: Actually, I'm not sure this is a good idea. It makes the class naming tied to the external interface and would make code refactoring difficult while keeping compatibility. Also, this is way too fancy/complicated. I would prefer to just have an explicit, simple dictionary of runner names vs types (classes).

Also, please don't forget to remove the "TYPE" variable from the runner class if you're going to do this in any shape.

skt/executable.py
major: Why would we want to ignore any settings coming from the config file?

skt/runner.py
major: But the class naming was tied to the runner type passed via `cfg` already. Are you suggesting we make a dictionary to map a runner name to a particular class?

spbnick: Hmm, I think it was tied to the "TYPE" member value. Yeah, I was thinking we make a dictionary.

And actually I think we shouldn't have it in the runner module, but instead in the executable.py, where the configuration/options are being interpreted. And the same for all other classes chosen by configuration/options. That can be a separate PR, if we decide to do it.

major: So something like this?

```python
runner_types = { 'beaker': 'BeakerRunner', 'something_else': "OtherRunner" }
```

And just instantiate whatever class name is returned from `runner_types['beaker']`?

major: Okay, I backed out some of the refactoring of `runner.getrunner()` and I'll work that dictionary stuff in another PR.

skt/executable.py
spbnick: Ideally, I would like to be able to override any config file setting with a command-line option. E.g. if the config file says "jobtemplate=A", I should be able to override it with the "--beaker-job-template" option. At the same time I would like e.g. the "jobowner" specified in the config file to stay unchanged.

How I read the code above is that we either take all the parameters from the command line, or from the config file.

major: I think I see what you're saying now. Let me see if I can rework some of the code there.

skt/executable.py
```diff
@@ -782,15 +802,19 @@ def load_config(args):
...
+        # Use the runner configuration from the command line
+        cfg['runner']['type'] = cfg.get('type')
+        cfg['runner']['jobtemplate'] = cfg.get('jobtemplate')
+        cfg['runner']['jobowner'] = cfg.get('jobowner')
+
+    # If the config file has a 'runner' section, we should find any values
+    # set there that were not set on the command line.
+    if cfg.get('_name') == 'run' and config.has_section('runner'):
+        for option, value in config.items('runner'):
+            cfg['runner'][option] = value
```
spbnick: I would prefer a more general way of merging configuration file and command-line parameters, one where we don't have to list all the parameters explicitly, but this will have to do. Not required for my ACK.

agustinhenze: I agree with @spbnick, it'd be good if we can use https://github.com/bw2/ConfigArgParse

skt/executable.py
```diff
@@ -887,6 +911,16 @@ def check_args(parser, args):
...
 
+    # Check required arguments for 'run'
+    if args._name == 'run':
+
+        if not args.jobtemplate:
+            parser.error(
+                "A job template XML file is required to run jobs in Beaker. "
+                "Use the --beaker-job-template argument to specify the path "
+                "to a valid file."
+            )
```
agustinhenze: Mmm I understand the mismatch now, it wouldn't be better a custom action to validate this?

PR #243, opened 2018-06-18
PR to simplify path of patchlist in cmd_merge

danghai:
I make PR to simplify path of `patchlist` in cmd_merge. In some cases, the path might contain multiple slashes `/` together.  Such as `/skt-workdir//example.patch/`.
In this case, it should ignore redundant slashes and return `/skt-workdir/example.patch`. It will remove redundant the slashes of `localpatch` in `skt-rc`.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -163,8 +163,9 @@ def cmd_merge(cfg):
             utypes.append("[local patch]")
             idx = 0
             for patch in cfg.get('patchlist'):
+                patch = os.path.abspath(patch)
                 save_state(cfg, {'localpatch_%02d' % idx: patch})
-                ktree.merge_patch_file(os.path.abspath(patch))
+                ktree.merge_patch_file(patch)
                 idx += 1
 
         if cfg.get('pw'):


danghai:
I think , in a situation, I create the folder `skt-patch` in where it can store all patches that I want to merge, I need to pass a relative path in command skt merge ( `--patchlist  <relative-path>`), it will save it under `localpatch` in `skt-rc` file. I think it is better to remove these slashes.
Similarly email url, If I pass `https://patchwork.ozlabs.org/patch/886637/, it can merge it. It is better to remove the slashes at the end , and save it in `skt-rc`.

veruu:
@danghai I don't understand. You are able to pass both relative and absolute paths. If you pass an absolute path to skt, your patch makes a relative path out of it and the patch will not be found because the path is messed up. If the path ends with a slash, it's a directory and not a file, and passing directories to skt is not supported. Patches retrieved from Patchwork URLs are applied differently and have nothing to do with `--patchlist` option.
As far as I know, the only way how to get double slashes is when you type them in the patch path directly. And that's a user error which `os.path.abspath()` cleans up. If the only reason for this patch is to solve the aesthetics in the state file, then why not retrieve the `abspath` sooner and save that?

danghai:
@veruu 

> As far as I know, the only way how to get double slashes is when you type them in the patch path directly

Yeah that is what I mean. Using `os.path.abspath()` is better. Thanks

danghai:
@veruu I am not sure why it usually has this error when I try to run command merge: 
```
  File "/usr/bin/skt", line 11, in <module>
  File "/home/danghai/skt/skt/executable.py", line 895, in main
    args.func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 123, in wrapper
    func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 180, in cmd_merge
    raise e
requests.exceptions.InvalidSchema: No connection adapters were found for '/home/danghai/skt/https:/patchwork.ozlabs.org/patch/886637/mbox'
```
It never happens before.
Do you think it is good to merge a folder in where it store many patches. It may merge them based on a certain priority. So, we do not need to type each patch by hand

veruu:
@danghai you are probably using wrong options (it's hard to guess if you don't provide the command you ran). Check the help output provided by `skt merge --help` -- the `--patchlist` option is for *already existing local patches*, the `--pw` option is for Patchwork URLs (this is also mentioned in the readme)

danghai:
@veruu Sorry I make you confused. I mean the old command: 
```
          --pw https://patchwork.ozlabs.org/patch/886637
```
Now, it show error: 
```
  File "/usr/bin/skt", line 11, in <module>
  File "/home/danghai/skt/skt/executable.py", line 895, in main
    args.func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 123, in wrapper
    func(cfg)
  File "/home/danghai/skt/skt/executable.py", line 180, in cmd_merge
    raise e
requests.exceptions.InvalidSchema: No connection adapters were found for '/home/danghai/skt/https:/patchwork.ozlabs.org/patch/886637/mbox'
```

veruu:
@danghai what skt revision are you using? I'm pretty sure the master works just fine.

danghai:
@veruu the newest from master. But the kernel source tree is old.

danghai:
@veruu Let me remove everything, and do it at first step. Anyway, my PR is good enough for state file?

veruu:
@danghai yes, the current version looks good, thanks! Can you rebase it on top of current master?

danghai:
@veruu I have rebased it!

veruu:
@danghai thanks! However, only now I've noticed you kept the commit message same, so it doesn't reflect the current patch. Can you edit it to say something like `Get absolute path to patch before saving state` or similar pleas?

danghai:
@veruu Okay. I fixed it

PR #249, opened 2018-06-19
reporter: Clean up load_state_cfg() docstring

### LINKED ISSUE #246: load_config() for reporter still says exception is raised when runner is missing
This is not true anymore, the docstring should be fixed.

major:
The reporter continues on without runner data instead of raising
an exception.

Fixes #246.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -508,10 +508,10 @@ class Reporter(object):
         return '\n'.join(msg)
 
     def load_state_cfg(self, statefile):
-        """
-        Load state info from statefile and reassign to self.cfg.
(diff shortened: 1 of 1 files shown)


skt/reporter.py
```diff
@@ -497,11 +497,7 @@ def getreport(self):
         return '\n'.join(msg)
 
     def load_state_cfg(self, statefile):
-        """
-        Load state info from statefile and reassign to self.cfg.
-
-        Raises: Exception if required 'runner' section is missing
-        """
+        """Load state info from statefile and reassign to self.cfg."""
```
spbnick: Could you please document the argument as well, while you're at it?

PR #251, opened 2018-06-19
[WIP] Test reports in Circle-CI

### LINKED ISSUE #223: skt report should be able to run after any stage, including merge

major:
Fixes #223.


skt/reporter.py
```diff
@@ -234,6 +234,24 @@ def __init__(self, cfg):
         # multireporting
         self.multi_job_ids = []
 
+    def get_config_local(self):
+        """Retrieve a local config file from the disk."""
```
spbnick: I think this needs to say "the local config file", and we need to define better where that file comes from and when it's created.

skt/reporter.py
```diff
@@ -234,6 +234,24 @@ def __init__(self, cfg):
...
+        config_file = "{}/.config".format('workdir')
+        if os.path.isfile(config_file):
+            # The config file
+            with open(config_file, "r") as fileh:
+                return fileh.read()
+
+        return None
+
+    def get_config_remote(self):
+        """Retrieve a config file from a remote location."""
```
spbnick: Same here, it needs to say "the published config file", and needs to either define publishing, or to refer to another place where it's done. Major part of why it's so difficult to describe these things here is that we keep passing the "cfg" object (essentially a bunch of global variables) around, instead of just what we need and no more. Each place has to keep track of what the rest of the program does, and know about (and refer to) all the events and entities it can possibly interact with.

skt/reporter.py
```diff
@@ -286,15 +304,11 @@ def update_mergedata(self):
...
-                    mergedata['config'] = response.text
-            else:
-                with open("%s/.config" % self.cfg.get("workdir"),
-                          "r") as fileh:
-                    mergedata['config'] = fileh.read()
+        # Retrieve the kernel config file
+        if not self.cfg.get('mergelog') and self.cfg.get('cfgurl'):
+            mergedata['config'] = self.get_config_remote()
+        elif not self.cfg.get('mergelog'):
+            mergedata['config'] = self.get_config_local()
```
spbnick: Wouldn't this be better written like this?

    if not self.cfg.get('mergelog'):
        if self.cfg.get('cfgurl'):
            mergedata['config'] = self.get_config_remote()
        else:
            mergedata['config'] = self.get_config_local()

skt/reporter.py
```diff
@@ -234,6 +234,24 @@ def __init__(self, cfg):
...
+            with open(config_file, "r") as fileh:
+                return fileh.read()
+
+        return None
+
+    def get_config_remote(self):
+        """Retrieve a config file from a remote location."""
+        response = requests.get(self.cfg.get("cfgurl"))
+        if response.status_code == requests.codes.ok:
+            return response.text
```
spbnick: Hmm, shouldn't we throw an exception if we get anything but 200 OK here?

skt/reporter.py
```diff
@@ -234,6 +234,24 @@ def __init__(self, cfg):
...
+
+        return None
```
spbnick: Can we have this return None, only if "cfgurl" is not set? Then we'll be able to simply try to get the remote config, and if it fails, get the local one?

Actually, why do we even get the remote config?

skt/reporter.py
```diff
@@ -349,11 +349,20 @@ def getmergeinfo(self):
         return result
 
     def get_kernel_config(self, suffix=None):
-        cfgname = "config.gz" if not suffix else "config_{}.gz".format(suffix)
+        """Add the kernel config to the list of attachments."""
```
spbnick: Could you please document the arguments and the return value, as well, including the behavior when there's no config?

skt/reporter.py
major: Fixed.

skt/reporter.py
major: Fixed.

skt/reporter.py
major: I can do that. I was told not to stack the `if` statements in another review. ;)

skt/reporter.py
major: A 301/302 redirect would be fine and requests should follow those. For example, the published location might redirect 80 to 443 and I doubt we should throw an exception for that.

skt/reporter.py
major: This function only gets called if `cfg['cfgurl']` is set.

skt/reporter.py
veruu: If redirecting is allowed, the `status_code` is the status of the last redirect, which should be `OK`, not 3XX.

skt/reporter.py
major: Fixed.

skt/reporter.py
spbnick: Hmm, surely not repeating a condition is good for maintenance? @veruu, was it a request from you? What do you think of this?

veruu: I don't remember saying anything like that (please correct me if I did) and I agree that @spbnick 's solution is more readable and clear

skt/reporter.py
spbnick: Well, if it's a 30x code, then we shouldn't be returning `response.text` in any case. Instead we need to follow the redirect, and raise an exception if we couldn't get the configuration.

skt/reporter.py
spbnick: Yeah, that's what I noticed too. So I was thinking, make it return None only if `cfgurl` is not set, and make it throw an exception if it failed to retrieve the configuration. Then we'll be able to write this:

    config = self.get_config_remote() or self.get_config_local()

and the `update_mergedata()` function won't need to know anything about `cfgurl`.

skt/reporter.py
```diff
@@ -350,11 +350,33 @@ def getmergeinfo(self):
...
+        Args:
+            suffix: An optional filename suffix to add to the config file
+                    name. This is helpful for reporting config files for
+                    different kernel architectures.
+
+        Returns:
+            If the 'config' item exists in self.mergedata, the method returns
+            a new line for the report as a single item within a list. If the
+            'config' item does not exist, the method returns an empty list
+            and no extra lines are added to the report.
```
spbnick: This is a good explanation, thank you. Although I would drop the last part - the "and no extra lines are added to the report" phrase. Because it is not for this function to decide what gets added to the report and what doesn't. It certainly makes no reports itself.

major:
@spbnick Allowing `get_config_local()` to throw an exception breaks the whole idea of allowing reports to be sent at any stage. A config file isn't required when the merge step runs and there needs to be a graceful way to skip that step.

veruu:
I might be wrong since I'm entering this discussion without actually reviewing the pull and just reading the comments, but I think that `allowing reports to be sent at any stage` has nothing to do with throwing an exception if something goes wrong. If the config file isn't needed, we shouldn't attempt to retrieve it. If it is, we should try to get it and throw an error if we can't.

major:
Config files are only needed when the merge has completed successfully and a build has been attempted. That's why I reworked the methods to ignore a missing config file if it doesn't exist.

spbnick:
I think I've been rather too verbose and confusing here, sorry. I'll try to boil it down:

* [ ] We need to find out if we really need to get the remote file. If not, remove that code. If yes, then we need to complete `get_config_remote()` documentation by adding a section on return values.
* [ ] We need to decide what `get_config_local()` means when it returns `None`, make sure that makes sense and we don't need to raise an exception if the file is missing, and then complete its documentation.

major:
@spbnick I agree on the fixes needed for `get_config_local()`. I can't think of a time when `get_config_remote()` is really needed since we must have the config stored locally if we want to compile the kernel. We should always have a local copy at reporting time.

The only exception would be if we've performed the merge and haven't started the compile yet. At that point, we don't have a local config.

major:
So the only action we take is to attach the config to the email. I think we can simplify this a bunch by renaming the method to something like `attach_kernel_config()` and moving the call down to the actual reporting method.

skt/reporter.py
```diff
@@ -273,29 +273,26 @@ def stateconfigdata(self, mergedata):
...
+        """
         mergedata = {
             'base': None,
             'merge_git': [],
             'localpatch': [],
             'patchwork': [],
-            'config': None
         }
 
         if self.cfg.get("infourl"):
```
spbnick: I feel it's the same story with the "info file". I.e. we don't really need to download it. However, let's deal with this in a separate PR.

skt/reporter.py
```diff
@@ -273,29 +273,26 @@ def stateconfigdata(self, mergedata):
         return mergedata
 
     def update_mergedata(self):
+        """Get information about the patch(es) being merged.
+
+        Returns: A dictionary of information about the merge, including the
```
spbnick: Google Python Style Guide says the "Returns" description should start on the next line. It would be good to keep the code consistent. This fix is not required for my ACK, though.

skt/reporter.py
```diff
@@ -334,12 +331,41 @@ def getmergeinfo(self):
...
+
+        if suffix:
+            attachment_filename = "config_{}.gz".format(suffix)
+
+        # Read the kernel config file, if present. The config file may not be
+        # present if we have completed a merge but we have not yet started the
+        # kernel compile.
+        config_filename = "{}/.config".format('workdir')
+        if os.path.isfile(config_filename):
+            with open(config_filename, 'r') as fileh:
```
spbnick: I would strongly prefer that we stopped doing this: checking the file exists and then opening it. While in the overall scheme, right now, this is insignificant, as we're full of race conditions, this is a bad pattern that needs to be avoided in general. Just open the file, catch the exception, and handle ENOENT.

This is not a requirement for my ACK at this moment, though.

veruu:
I didn't really keep up with the pull but since Nick asked me to check it out... I'm slightly concerned about the removal of remote configs. Are we really sure that the reports will have access to the config in the workdir? This makes it a requirement for the report to be always run on the same host the kernel was built on. Will this work especially if we want to move the reporting to sktm, where we might need to reimplement the functionality again?

If the answer for all of the above is that it will be fine, then `cfgurl` should be removed from skt altogether, not only from reporting part, since it's only point of publishing it with the build is to have it available for the report.

skt/reporter.py
major: Fixing.

skt/reporter.py
major: Fixing.

spbnick:
I'm on the fence here. Yeah, remote reporting might be useful with e.g. containers in theory, but we don't use it yet (so it might as well not even work) and I'm not sure if architecturally it's a good idea. I also agree, that if we're taking this away, we need to remove all support for remote reporting and not leave it broken/half-working. That might be more work than we have time to do right now.

In the end perhaps, let's try to see what kind of design we come up with for moving reporting to sktm (how much effort it will be, and how fast we can do it), and wait with this until then.

PR #255, opened 2018-06-21
Add LTP lite reports on failures

veruu:
LTP lite was added to tests that are being run, but the report wasn't
updated to include links to logs. Since console logs aren't very useful
with LTP most of the time, these logs are important. Because we skip
reporting of failures we had no console logs for (residue of only
boot-testing), we also reported a failure but without any indication of
what actually failed.

The change boils down to addition of BeakerRunner method to retrieve LTP
logs for each recipe in the jobs and adding them to reports.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -227,6 +227,46 @@ class BeakerRunner(Runner):
 
         return result
 
+    def get_ltp_lite_logs(self, job_id):
+        """
+        Get logs produces by LTP lite, for each recipe in specified job.
+
+        Args:
+            job_id: ID of the job for which to retrieve the logs.
+
+        Returns:
+            A dictionary in format
+              R:<recipe_id>: (<link_to_full_run_log>, <link_to_results>),
+            for each recipe in the job.
+        """
+        job_results = self.getresultstree(job_id, True)
+        ltp_logs = {}
+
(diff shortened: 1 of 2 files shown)


skt/runner.py
```diff
@@ -227,6 +227,46 @@ def getverboseresults(self, joblist):
...
+        job_results = self.getresultstree(job_id, True)
+        ltp_logs = {}
+
+        for recipe in job_results.findall('recipeSet/recipe'):
+            recipe_id = 'R:%s' % recipe.attrib.get('id')
+
+            ltp_node = recipe.find(
+                "task[@name='/kernel/distribution/ltp/lite']"
+            )
+            if ltp_node is None:
```
major: `xpath` could help a lot here to avoid lots of `if` statements, but this is okay for now. Dealing with XML is a huge pain.

skt/runner.py
```diff
@@ -227,6 +227,46 @@ def getverboseresults(self, joblist):
 
         return result
 
+    def get_ltp_lite_logs(self, job_id):
+        """
```
major: `pydocstyle` complains if there's no content in the first line of the comment, but we can fix that later if we ever enable it.

PR #258, opened 2018-06-22
Runner improvements

veruu:
Fix some pylint errors and bug with infinite jobs resubmitting

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -149,9 +149,11 @@ class BeakerRunner(Runner):
             jobid = self.lastsubmitted
 
         root = self.getresultstree(jobid, True)
-        el = root.find("recipeSet/recipe/logs/log[@name='console.log']")
-        if el is not None:
-            url = el.attrib.get("href")
+        console_log = root.find(
+            "recipeSet/recipe/logs/log[@name='console.log']"
+        )
+        if console_log is not None:
+            url = console_log.attrib.get("href")
 
         return url
 
@@ -172,7 +174,7 @@ class BeakerRunner(Runner):
             for (jid, rset) in self.j2r.iteritems():
                 if cid in rset:
                     rset.remove(cid)
-                    if len(rset) == 0:
+                    if not rset:
                         deljids.add(jid)
             for jid in deljids:
                 del self.j2r[jid]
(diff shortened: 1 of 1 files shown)


skt/runner.py
```diff
@@ -149,9 +149,9 @@ def getconsolelog(self, jobid=None):
             jobid = self.lastsubmitted
 
         root = self.getresultstree(jobid, True)
-        el = root.find("recipeSet/recipe/logs/log[@name='console.log']")
-        if el is not None:
-            url = el.attrib.get("href")
+        console = root.find("recipeSet/recipe/logs/log[@name='console.log']")
+        if console is not None:
+            url = console.attrib.get("href")
```
spbnick: Do you think this would be better to name `console_log` instead?

skt/runner.py
```diff
@@ -325,12 +325,17 @@ def getresults(self, jobid=None):
...
         if samehost:
-            hreq = tmp.find("hostRequires")
             hostname = etree.Element("hostname")
             hostname.set("op", "=")
             hostname.set("value", tmp.attrib.get("system"))
             hreq.append(hostname)
+        else:
+            hostname = hreq.find('hostname')
+            if hostname is not None:
+                hreq.remove(hostname)
```
spbnick: Hmm, shouldn't we simply remove the `hostname` requirement every time, and only add it back when `samehost` is True?

veruu: I don't understand how you mean it. If the `samehost` is true, we keep the hostname and if not, we check if there is a hostname specified in the recipe and remove it if there is one. Can you elaborate how you imagine the solution?

spbnick: Like this:

        hreq = tmp.find("hostRequires")
        hostname = hreq.find('hostname')
        if hostname is not None:
            hreq.remove(hostname)
        if samehost:
            hostname = etree.Element("hostname")
            hostname.set("op", "=")
            hostname.set("value", tmp.attrib.get("system"))
            hreq.append(hostname)

This way we make sure there's no `hostname` element when `samehost` is `False`, and there's only one `hostname` element, when it's `True` (that is unless there was more than one originally, which we can also handle, if needed).

veruu: Thanks for the explanation! Yes, that makes sense. I'll update the pull.

veruu:
@spbnick updated, thanks for the idea again!

PR #262, opened 2018-06-26
CONTRIBUTING.md: Require testing from the reviewer

spbnick:
I think we should test each submitted pull request (if not each patch). Unless we're very sure the change is automatically-tested by CI, we should test it ourselves, manually if need be. Ultimately, the reviewer is responsible for the change quality, so they should test the change as well. That doesn't mean the submitter shouldn't test them, of course.


major:
While I agree with this, I would hope we could improve unit/functional testing coverage so we don't need to worry about this. ;)

spbnick:
I agree, we need more tests, and integration tests especially. However, presence of automatic tests should not really be an excuse for letting some breakage in through review, no matter how many of them we have.

Rebased, could I get an approval, if we agree :)?

spbnick:
Or rather, having automatic tests should not be a reason not to read the code and not to do at least basic, sanity testing. We'll let *some* breakage in despite everything we try anyway, of course.

spbnick:
Rebased! Can I have another approval please :)?

PR #263, opened 2018-06-26
Fix placing of binary operators

veruu:
PEP8 encourages writing binary operators on the beginning of
continuation line, not end of the previous one (so one can easily see
what's going on). pycodestyle by default doesn't check for these.


major:
@veruu Feel free to rebase it and we can get it merged. :)

veruu:
@major rebased

PR #265, opened 2018-06-27
Mark private and protect methods

### LINKED ISSUE #10: Mark all private class methods as such
To avoid exposing too much interface, prepend all private class method names with __ to have them recognized as such.

danghai:
I make PR to mark private and protect methods. 
Fix #10

### DIFF
--- a/skt/kerneltree.py
+++ b/skt/kerneltree.py
@@ -251,7 +251,7 @@ class KernelTree(object):
         if self.fetch_depth:
             git_fetch_args.extend(['--depth', self.fetch_depth])
 
-        # The git_cmd() method expects a list of args, not a list of strings,
+        # The __git_cmd() method expects a list of args, not a list of strings,
         # so we need to expand our list into args with *.
         self.__git_cmd(*git_fetch_args)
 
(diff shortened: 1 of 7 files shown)


skt/kerneltree.py
```diff
@@ -251,11 +253,11 @@ def checkout(self):
 
         # The git_cmd() method expects a list of args, not a list of strings,
```
veruu: Can you change the comment to use the new method name (`__git_cmd()` instead of `git_cmd()`) as well?

danghai: If you want to change the name in comment, could I create another PR to do that because there are a lot comment need to be changed including the previous PR (`KernelBuilder`)?

veruu: You can add a commit for fixing all of the comments then, that would work too

danghai:
@veruu Thanks review. I reword these commits. I hope it should be merged soon because it creates a lot conflict with other PR. Haha :). I am not really good to rebase, and then fixing conflict.

danghai:
@veruu I add a commit to reword all private method names in comment too

PR #269, opened 2018-06-29
Test_kerneltree: Remove pylint: disable= too-many-public-methods 

danghai:
I make PR to remove pylint: disable= too-many-public-methods.
Although I mark the private and protected methods in Runner, I cannot remove this disable in `test_runner`. 
```
************* Module tests.test_runner
R: 35, 0: Too many public methods (22/20) (too-many-public-methods)
```

### DIFF
--- a/tests/test_kerneltree.py
+++ b/tests/test_kerneltree.py
@@ -32,7 +32,6 @@ def make_process_exception(*args, **kwargs):
 
 
 class KernelTreeTest(unittest.TestCase):
-    # (Too many public methods) pylint: disable=too-many-public-methods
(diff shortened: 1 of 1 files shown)


veruu:
Thanks Hai, but I'd leave the disabling as it is. It's not related to marking methods in tested classes, but rather to how many methods the test classes themselves have. Number of `KernelTreeTest` class methods is currently 20, which is the threshold for this warning. This means that adding a single test to the class, the warning will reappear, and will need to be disabled again.

danghai:
@veruu oh that is interesting! I did not know it before. Thanks

PR #270, opened 2018-06-29
Various pylint improvements

danghai:
I make PR to make various pylint improvements.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -45,7 +45,7 @@ retcode = 0
 
 
 def full_path(path):
-    """Get an absolute path to a file"""
+    """Get an absolute path to a file."""
     return os.path.abspath(os.path.expanduser(path))
 
 
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -481,7 +481,7 @@ def setup_parser():
         "-w",
         "--wipe",
         help=(
-            "Clean build (make mrproper before building) and remove workdir"
+            "Clean build (make mrproper before building) and remove workdir "
             "when finished"
         ),
         action="store_true",
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -14,6 +14,7 @@
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+from __future__ import print_function
 import ConfigParser
 import argparse
 import ast
@@ -104,7 +105,7 @@ def junit(func):
         global retcode
         if cfg.get('junit'):
             tstart = time.time()
-            tc = junit_xml.TestCase(func.__name__, classname="skt")
+            testcase = junit_xml.TestCase(func.__name__, classname="skt")
 
             try:
                 func(cfg)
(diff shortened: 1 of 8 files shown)


skt/executable.py
```diff
@@ -509,21 +510,21 @@ def setup_parser():
         type=str,
         action='append',
         help="Path to a local patch to apply "
-             + "(use multiple times for multiple patches)"
+        + "(use multiple times for multiple patches)"
```
veruu: While this line (and those below) fix the pylint warning, the changes make the code less readable since it's not easy to see where the help actually ends. If you can solve the warning in a way it stays clear that would be more than awesome, if not, I'd prefer for it to stay as it was, even if it produces a warning.

skt/executable.py
```diff
@@ -912,5 +913,5 @@ def main():
         main()
     except KeyboardInterrupt:
         # cleanup??
-        print("\nExited at user request.")
+        print"\nExited at user request."
```
veruu: I thought we switched to `print_function` in skt already. Seems not. Can you use that instead?

skt/kerneltree.py
```diff
@@ -117,14 +117,14 @@ def __git_cmd(self, *args, **kwargs):
...
+            raise exc
 
-    def __git_cmd_pipe(self, input, *args, **kwargs):
+    def __git_cmd_pipe(self, iinput, *args, **kwargs):
         """
         Feed input to a git command and return its output.
 
         Args:
-            input:      Input to feed to the git command.
+            iinput:      Input to feed to the git command.
```
veruu: While this fixes the pylint error, the new name of variable is unclear and prone to errors by oversight (it's easy to write additional code with `input` since having two `i` at the beginning is not something people expect). Can you use a different name that's still clear, maybe something like `command_input` or similar? Also, the indentation in docstring needs to be adjusted for the variable name to fit.

skt/reporter.py
```diff
@@ -784,6 +785,6 @@ def report(self):
 
             msg.attach(tmp)
 
-        s = smtplib.SMTP(self.smtp_url)
-        s.sendmail(self.mailfrom, self.mailto, msg.as_string())
-        s.quit()
+        server = smtplib.SMTP(self.smtp_url)
```
veruu: Can you use a more specific name here, like `mailserver`?

skt/executable.py
danghai: I am not sure what you mean `the changes make the code less readable since it's not easy to see where the help actually ends` ? 
Furthermore, I am not sure there are some different format for `help`. For example: https://github.com/RH-FMK/skt/blob/master/skt/executable.py#L531

danghai: I do not know the name for `mb`. Could you give the name for it? https://github.com/RH-FMK/skt/blob/master/skt/executable.py#L152

veruu: @danghai please check the git history of why the formatting was changed, and the internet for "how to split long string in Python" which should show you both of these (and maybe some others too) options with description.

> I am not sure what you mean the changes make the code less readable since it's not easy to see where the help actually ends ?

Python is based on code indentation and readability, as I'm sure you noticed. If the continuation line is indented same as the line above, it doesn't look like continuation line but separate statement on the first look.

danghai: @veruu I see the reason why in commit `Fix placing of binary operators`. So I will keep the same indentation in spite of the warning in `pylint`.

danghai:
@veruu I see the reason why in commit `Fix placing of binary operators`. So I will keep the same indentation in spite of the warning in `pylint`. 
I also fix some docstrings and add one space in parser argument.

PR #271, opened 2018-06-29
Distinguish between test failures and infrastructure errors

### LINKED ISSUE #268: skt does not differentiate between Beaker infrastructure and test failures
We've had an "skt run" fail due to Beaker service interruption. However, skt doesn't seem to provide a way for whoever executed it to differentiate between a test failure and an infrastructure/code failure. A good start could be a separate exit code.

See [the (redacted) output of multi-arch pipeline](https://github.com/RH-FMK/skt/files/2144081/failure.log).

veruu:
Hopefully fixes #268

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -132,6 +132,10 @@ class BeakerRunner(Runner):
         with open(fname, 'w') as fileh:
             bkr = subprocess.Popen(args, stdout=fileh)
             bkr.communicate()
+        if bkr.returncode != 0:
+            raise Exception(
+                'Unable to get Beaker job results for job %s' % jobid
+            )
 
     def __getconsolelog(self, jobid=None):
         """
@@ -496,16 +500,22 @@ class BeakerRunner(Runner):
             hostname = "(%s) " % host
             hostnametag = '<hostname op="=" value="%s"/>' % host
 
-        jobid = self.__jobsubmit(self.__getxml({'KVER': release,
-                                                'KPKG_URL': url,
-                                                'UID': uid,
-                                                'ARCH': arch,
-                                                'HOSTNAME': hostname,
-                                                'HOSTNAMETAG': hostnametag}))
+        try:
+            jobid = self.__jobsubmit(self.__getxml(
+                {'KVER': release,
+                 'KPKG_URL': url,
+                 'UID': uid,
+                 'ARCH': arch,
+                 'HOSTNAME': hostname,
+                 'HOSTNAMETAG': hostnametag}
+            ))
 
-        if wait:
-            self.wait(jobid, reschedule)
-            ret = self.__getresults()
+            if wait:
+                self.wait(jobid, reschedule)
+                ret = self.__getresults()
+        except Exception as exc:
+            logging.error(exc)
+            ret = 2
 
         return ret
 
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -485,6 +485,30 @@ class BeakerRunner(Runner):
 
     def run(self, url, release, wait=False, host=None, uid="",
             arch=platform.machine(), reschedule=True):
+        """
+        Run tests in Beaker.
+
+        Args:
+            url:        URL pointing to kernel tarball.
+            release:    NVR of the tested kernel.
+            wait:       False if skt should exit after submitting the jobs,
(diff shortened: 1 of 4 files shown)


skt/runner.py
```diff
@@ -496,16 +500,22 @@ def run(self, url, release, wait=False, host=None, uid="",
...
+                 'HOSTNAME': hostname,
+                 'HOSTNAMETAG': hostnametag}
+            ))
+
+            if wait:
+                self.wait(jobid, reschedule)
+                ret = self.__getresults()
+        except Exception as exc:
+            logging.error(exc)
+            ret = 2
```
spbnick: Could you please document the function, including the exit status?

veruu: Added!

skt/runner.py
```diff
@@ -485,6 +485,29 @@ def __jobsubmit(self, xml):
...
+            url:
+            release:    NVR of the tested kernel.
+            wait:       False if skt should exit after submitting the jobs,
+                        True if it should wait for them to finish.
+            host:       Force testing on a machine with provided hostname.
+            uid:        Username jobs should be submitted under. Can be
+                        retrieved by running `bkr whoami`.
+            arch:       Architecture of the machine the tests should run on.
+                        Defaults to architecture of the current machine skt is
+                        running on if not specified.
```
spbnick: It would be good to mention the architecture names are those expected by Beaker. I'm not sure if they will always match `platform.machine()`, BTW.

skt/runner.py
```diff
@@ -485,6 +485,29 @@ def __jobsubmit(self, xml):
 
     def run(self, url, release, wait=False, host=None, uid="",
             arch=platform.machine(), reschedule=True):
+        """
+        Main method for testing in Beaker.
+
+        Args:
+            url:
```
spbnick: Could you please document this argument as well?

skt/runner.py
```diff
@@ -485,6 +485,29 @@ def __jobsubmit(self, xml):
...
+                        retrieved by running `bkr whoami`.
+            arch:       Architecture of the machine the tests should run on.
+                        Defaults to architecture of the current machine skt is
+                        running on if not specified.
+            reschedule: True to try to rule out infrastructure / host-specific
+                        failures by resubmitting the job (on both the same and
+                        different host), False otherwise.
+
+        Returns:
+            ret: 0 if everything passed
```
spbnick: Google Python Style Guide doesn't seem to require documenting variables used internally to store return values. And why would the function user need to know?

Is there some reason/purpose I'm missing, or could you perhaps remove it?

skt/runner.py
```diff
@@ -485,6 +485,29 @@ def __jobsubmit(self, xml):
 
     def run(self, url, release, wait=False, host=None, uid="",
             arch=platform.machine(), reschedule=True):
+        """
+        Main method for testing in Beaker.
```
spbnick: Could you please use imperative mood, and descriptive instead of prescriptive language? Also, "method" would be redundant here. Something like "Run the tests" would work.

skt/runner.py
veruu: That's the name of the variable being returned, and explanation of possible values. I can remove the name if that's what's confusing

skt/runner.py
veruu: From what I've seen so far they do match, but I can't find any official docs. If you have any information about that, please link them and I'll make sure things work.

veruu:
@spbnick added the docs for `url`, removed the name of variable and changed the main wording. Let me know if there's anything else you see

skt/runner.py
spbnick: I don't have any information about that. A quick look through Beaker documentation, hasn't revealed anything of value. I think it's just good enough to mention those are what Beaker expect, and e.g. say that you need to go to Beaker UI and see what it accepts, or just say "Consult your Beaker installation for accepted architectures". Main thing is that it's not just some abstract architecture, and you can find a supported list somehow. If they do match what `platform.machine()` or anything else returns, great, but we cannot really guarantee that, and it's best to consult Beaker.

veruu:
@spbnick added `in a format accepted by Beaker` to arch description. I don't want you to be sad! :D

PR #272, opened 2018-07-01
Switch to use rpartition function instead of split function

danghai:
I make PR to switch to use `rpartition` function instead of `split` function for clear. 
In order to get the last part in a path, instead of using split: `split('/')[-1]`, Python supports the `rpartition(sep)` function. It splits the string at the last occurrence of sep, and returns a 3-tupple containing the part before the separator, the separator itself, and the part after the separator.

### DIFF
--- a/skt/kerneltree.py
+++ b/skt/kerneltree.py
@@ -240,7 +240,7 @@ class KernelTree(object):
         Returns:
             Full hash of the last commit.
         """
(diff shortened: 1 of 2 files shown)


danghai:
@veruu yeah I follow the C/C++ coding with strict rule for accessing element. The way to access in Python is really flexible. I do not know this is a pythonic idiom. I do not know much about the tip, trick, or pythonic idiom. I thought I just make the code more clear when I try to get the last element when I read the code. :) .

PR #275, opened 2018-07-04
Raise exception if file to report doesn't exist

### LINKED ISSUE #274: skt report produces obscure exception if --report file doesn't exist
When executing the following command:

    skt -vv report --reporter mail --mail-from 'CI <ci@example.com>' --mail-to 'Maintainers <maintainers@example.com>' --result NON-EXISTENT

skt produces this exception:

```
  File "/home/nkondras/projects/github.com/rh-fmk/skt/skt/executable.py", line 927, in main
    args.func(cfg)
  File "/home/nkondras/projects/github.com/rh-fmk/skt/skt/executable.py", line 384, in cmd_report
    reporter.report()
  File "/home/nkondras/projects/github.com/rh-fmk/skt/skt/reporter.py", line 762, in report
    msg.attach(MIMEText(self._get_multireport()))
  File "/home/nkondras/projects/github.com/rh-fmk/skt/skt/reporter.py", line 578, in _get_multireport
    self.cfg = load_state_cfg(statefile)
  File "/home/nkondras/projects/github.com/rh-fmk/skt/skt/reporter.py", line 72, in load_state_cfg
    for (name, value) in state_to_report.items('state'):
  File "/usr/lib/python2.7/ConfigParser.py", line 642, in items
    raise NoSectionError(section)
ConfigParser.NoSectionError: No section: 'state'
```

Instead, skt should produce a clear error message about a missing sktrc.

veruu:
ConfigParser.read() returns a list of successfully parsed filenames.
This can be checked and exception about nonexistent file can be raised
if it's empty.

This makes it clear where the underlying problem is, instead of failing
when retrieving state section and producing cryptic exceptions.

Fixes #274

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -65,7 +65,8 @@ def load_state_cfg(statefile):
     """
     cfg = {}
     state_to_report = ConfigParser.ConfigParser()
-    state_to_report.read(statefile)
+    with open(statefile, 'r') as fileh:
+        state_to_report.readfp(fileh)
 
     # FIXME This can be simplified or removed after configuration and
     # state split


skt/reporter.py
```diff
@@ -65,7 +65,12 @@ def load_state_cfg(statefile):
     """
     cfg = {}
     state_to_report = ConfigParser.ConfigParser()
-    state_to_report.read(statefile)
+    read_file = state_to_report.read(statefile)
+    if not read_file:
+        raise OSError(
+            2,
+            'Can\'t create report for %s, file not found!' % statefile
+        )
```
spbnick: This would still fire, if we e.g. didn't have permission to read the file (common thing when messing around with Jenkins users), and the error would be confusing.

Could you use `open()` and `state_to_report.readfp()` instead and produce appropriate error messages, or exceptions?

veruu:
@spbnick fixed

skt/reporter.py
```diff
@@ -60,12 +60,16 @@ def load_state_cfg(statefile):
     Args:
         statefile:      Path to a skt state file to read
 
+    Raises:
+        Exception in case statefile can't be read.
```
spbnick: Thank you, Veronika! Now we only have documentation to update :)

veruu: Wait, what documentation do you mean? AFAIK we aren't listing which function can throw what exceptions anywhere

spbnick: Alright, this docstring can be read as "an exception", or `Exception`. If it's the former, then it's better to remove it, since it is of no use to the caller, because they can't catch the specific exception. If it's the latter, then it's just wrong, because we're not throwing `Exception` :) So, either specify the exception type and conditions it's being thrown at, or just remove the section :)

veruu: It's supposed to be the second one. There are multiple subclasses of `Exception` that can be raised (eg. `OSError` if the file doesn't exist, `IOError` if we don't have the permission to read the file). I *think* it should be safe to name those two, but am not 100% sure if there is a possibility for a different one, so I went with the base class. I can remove it if you don't like the generic info though.

veruu: Removed.

veruu:
Sure, makes sense :)

PR #278, opened 2018-07-12
Don't try to report aborted jobs

veruu:
If the test job aborts, there are no results to report. Skip test
reporting if all of the jobs aborted, and skip reporting of the
actual aborted jobs, instead of getting an exception on trying to
report nonexistent results.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -469,12 +469,9 @@ class Reporter(object):
 
         minfo = {"short": {}, "long": {}}
         jidx = 1
-        for jobid in job_list:
-            for (recipe, rdata) in vresults[jobid].iteritems():
-                if recipe == "result":
-                    continue
-
-                (res, system, clogurl, slshwurl, _) = rdata
+        for jobid, values in vresults.items():
+            for recipe in [key for key in values if key.startswith('R:')]:
+                (res, system, clogurl, slshwurl, _) = values[recipe]
 
                 clog = ConsoleLog(self.cfg.get("krelease"), clogurl)
                 if not clog.data and res != 'Pass':
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -526,6 +526,10 @@ class Reporter(object):
                 result.append('')
                 jidx += 1
 
+            if self.multireport and self.cfg.get('retcode') != '0' and \
+                    self.multireport_failed == MULTI_PASS:
+                self.multireport_failed = MULTI_TEST
+
         return result
 
     def _getreport(self):
@@ -624,9 +628,6 @@ class Reporter(object):
                     results += self.__getjobresults()
 
             results.append('\n')
-
-            if self.cfg.get('retcode') != '0':
-                self.multireport_failed = MULTI_TEST
 
         results += ['Please reply to this email if you find an issue with our '
                     'testing process,',
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -470,6 +470,12 @@ class Reporter(object):
         minfo = {"short": {}, "long": {}}
         jidx = 1
         for jobid, values in vresults.items():
+            job_result = values.get('result')
+            job_status = values.get('status')
+            if job_result == 'Warn' and job_status == 'Aborted':
+                logging.info('Skipping aborted job %s', jobid)
+                continue
+
             for recipe in [key for key in values if key.startswith('R:')]:
                 (res, system, clogurl, slshwurl, _) = values[recipe]
 
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -326,10 +326,10 @@ class BeakerRunner(Runner):
         fhosts = set()
         tfailures = 0
 
-        if jobid is not None:
+        if jobid:
             (ret, _) = self.__jobresult(jobid)
 
-        if jobid is None or ret != 0:
+        if not jobid or ret != 0:
             if self.failures:
                 all_aborted = all([data[1] == ('Warn', 'Aborted')
                                    for data in self.failures.values()])
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -21,6 +21,9 @@ import time
 import xml.etree.ElementTree as etree
 
 
+MAX_ABORTED = 3
+
+
 class Runner(object):
     """An abstract test runner"""
     # TODO This probably shouldn't be here as we never use it, and it should
@@ -66,6 +69,7 @@ class BeakerRunner(Runner):
         self.jobs = set()
         self.lastsubmitted = None
         self.j2r = dict()
+        self.aborted_count = 0
 
         logging.info("runner type: %s", self.TYPE)
         logging.info("beaker template: %s", self.template)
@@ -338,6 +342,9 @@ class BeakerRunner(Runner):
                     logging.warning('All jobs aborted, possible infrastructure'
                                     ' failure.')
                     return 2
+            if self.aborted_count == MAX_ABORTED:
+                logging.error('Max count of aborted jobs achieved, please '
+                              'check your infrastructure!')
 
             for (recipe, data) in self.failures.iteritems():
                 # Treat single failure as a fluke during normal run
@@ -423,9 +430,16 @@ class BeakerRunner(Runner):
                             logging.warning("%s failed before kernelinstall, "
                                             "resubmitting", cid)
                             self.__forget_cid(cid)
-                            newjob = self.__recipe_to_job(root, False)
-                            newjobid = self.__jobsubmit(etree.tostring(newjob))
-                            self.__add_to_watchlist(newjobid, reschedule, None)
+
+                            if self.aborted_count < MAX_ABORTED:
+                                newjob = self.__recipe_to_job(root, False)
+                                newjobid = self.__jobsubmit(
+                                    etree.tostring(newjob)
+                                )
+                                self.__add_to_watchlist(newjobid,
+                                                        reschedule,
+                                                        None)
+                                self.aborted_count += 1
                         else:
                             if origin is None:
                                 origin = cid
@@ -549,6 +563,7 @@ class BeakerRunner(Runner):
         self.failures = {}
         self.recipes = set()
         self.watchlist = set()
+        self.aborted_count = 0
 
         # FIXME Pass or retrieve this explicitly
         uid += " %s" % url.split('/')[-1]
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -202,6 +202,7 @@ class BeakerRunner(Runner):
             root = self.getresultstree(jobid, True)
 
             result[jobid]['result'] = root.attrib.get("result")
+            result[jobid]['status'] = root.attrib.get('status')
 
             for recipe in root.findall("recipeSet/recipe"):
                 rid = "R:%s" % recipe.attrib.get("id")
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -367,17 +367,18 @@ class BeakerRunner(Runner):
             for (cid, reschedule, origin) in self.watchlist.copy():
                 root = self.getresultstree(cid)
 
-                if root.attrib.get("status") in ["Completed", "Aborted",
-                                                 "Cancelled"]:
+                result = root.attrib.get('result')
+                status = root.attrib.get('status')
+                if status in ['Completed', 'Aborted', 'Cancelled']:
                     logging.info("%s status changed to '%s', "
                                  "removing from watchlist",
                                  cid, root.attrib.get("status"))
                     self.watchlist.remove((cid, reschedule, origin))
 
-                    if root.attrib.get("status") == "Cancelled":
+                    if status == 'Cancelled':
                         continue
 
-                    if root.attrib.get("result") != "Pass":
+                    if result != 'Pass':
                         tinst = root.find(
                             ".//task[@name='/distribution/install']"
                         )
@@ -399,13 +400,11 @@ class BeakerRunner(Runner):
                             self.failures[origin][0].append(root.attrib.get(
                                 "system"
                             ))
-                            self.failures[origin][1].add(root.attrib.get(
-                                "result"
-                            ))
+                            self.failures[origin][1].add(result)
 
                             if reschedule:
                                 logging.info("%s -> '%s', resubmitting",
-                                             cid, root.attrib.get("result"))
+                                             cid, result)
 
                                 newjob = self.__recipe_to_job(root, False)
                                 newjobid = self.__jobsubmit(etree.tostring(
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -342,6 +342,12 @@ class BeakerRunner(Runner):
                     logging.warning('All jobs aborted, possible infrastructure'
                                     ' failure.')
                     return 2
+
+                job_cancelled = any(status == 'Cancelled'
+                                    for data in self.failures.values()
+                                    for _, status in data[1])
+                if job_cancelled:
+                    return 2
             if self.aborted_count == MAX_ABORTED:
                 logging.error('Max count of aborted jobs achieved, please '
                               'check your infrastructure!')
@@ -419,7 +425,20 @@ class BeakerRunner(Runner):
                     self.watchlist.remove((cid, reschedule, origin))
 
                     if status == 'Cancelled':
-                        continue
+                        logging.error('Cancelled job detected! Cancelling the'
+                                      ' rest of running jobs and aborting!')
+                        self.failures[cid] = [[root.attrib.get('system')],
+                                              set([(result, status)]),
+                                              1]
+                        jobs_to_cancel = [job for (job, _, _)
+                                          in self.watchlist.copy()]
+                        if jobs_to_cancel:
+                            ret = subprocess.call(['bkr', 'job-cancel']
+                                                  + jobs_to_cancel)
+                            if ret:
+                                logging.info('Failed to cancel the remaining '
+                                             'jobs')
+                        return
 
                     if result != 'Pass':
                         tinst = root.find(
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -282,16 +282,31 @@ class BeakerRunner(Runner):
         return max(set(fhosts), key=fhosts.count)
 
     def __jobresult(self, jobid):
-        ret = 0
+        """
+        Get results for a specific job.
+
+        Args:
+            jobid: A Beaker job ID, in 'J:<id>' format.
+
+        Returns:
+            A tuple (ret, result), where ret is 0 if the job ended with Pass,
+            1 if a test failure occurred and 2 if it aborted (infrastructure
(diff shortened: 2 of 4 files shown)


spbnick:
What does "aborted" job mean? Shouldn't we report this as "error", i.e. infrastructure failure, so that sktm could resubmit the job?

spbnick:
Or skt, for that matter?

veruu:
> What does "aborted" job mean? Shouldn't we report this as "error", i.e. infrastructure failure, so that sktm could resubmit the job?

Good idea. skt does resubmit aborted jobs (when trying to find out if the failure is host-specific), but in case all jobs are aborted, we should report an error.

Let me check if there is any chance beaker can report aborted job status in case the failure isn't infrastructure-related, and I'll do the appropriate changes.

veruu:
@spbnick I included some code simplifications and fixes of other issues I found in the meanwhile, so this turned out into more commits than I thought (tried to honor your "single change" comment from the last time). Let me know if you have additional ideas or notice an issue

skt/runner.py
```diff
@@ -281,16 +281,29 @@ def get_mfhost(self):
...
+        result = root.attrib.get("result")
+        status = root.attrib.get('status')
 
-            if result != "Pass":
-                ret = 1
-            logging.info("job result: %s [%d]", result, ret)
+        if result == 'Warn' and status == 'Aborted':
+            ret = 2
+        elif result != "Pass":
+            ret = 1
```
spbnick: Can we instead check for known "Pass" and "Fail" conditions explicitly, and consider everything else an error and return 2? This way we'll keep all the weird results away from developers, and get to learn all possible failure modes. I.e. if Beaker produces anything unanticipated we get to investigate and decide whether that is a "Pass" or a "Fail", instead of just assuming it's a "Fail" and sending the message?

skt/runner.py
```diff
@@ -316,6 +327,14 @@ def __getresults(self, jobid=None):
             (ret, _) = self.__jobresult(jobid)
 
         if jobid is None or ret != 0:
+            if self.failures:
+                all_aborted = all([data[1] == ('Warn', 'Aborted')
+                                   for data in self.failures.values()])
```
spbnick: `all_aborted` will be true if `self.failures.values()` is empty. Would that be possible and would that be correct?

skt/reporter.py
```diff
@@ -469,12 +469,9 @@ def __getjobresults(self):
...
         jidx = 1
-        for jobid in job_list:
-            for (recipe, rdata) in vresults[jobid].iteritems():
-                if recipe == "result":
-                    continue
-
-                (res, system, clogurl, slshwurl, _) = rdata
+        for jobid, values in vresults.items():
+            for recipe in [key for key in values if key.startswith('R:')]:
+                (res, system, clogurl, slshwurl, _) = values[recipe]
```
spbnick: I have no idea if this is equivalent, so I'll have to trust you :)

skt/reporter.py
```diff
@@ -470,6 +470,12 @@ def __getjobresults(self):
         minfo = {"short": {}, "long": {}}
         jidx = 1
         for jobid, values in vresults.items():
+            job_result = values.get('result')
+            job_status = values.get('status')
+            if job_result == 'Warn' and job_status == 'Aborted':
+                logging.info('Skipping aborted job %s', jobid)
+                continue
+
```
spbnick: Can we follow similar logic here? Accept only particular results as valid, and drop the rest?

Also, what happens to the dropped job? I would say that a single (errored) aborted job invalidates the whole run. I.e. we basically skipped one test and we don't know its outcome. We should re-run that job. Do we do that, or do we send a report after that?

spbnick:
The idea behind those result-handling suggestions is that we should consider *all* failures as "infrastructure failures" (errors) *by default*, in general, and only mark specific ones as failures worthy of developer's attention.

That being said, I can't really say what's going on in the code you're changing, you might be already doing that, and I would appreciate your explanation :)

skt/runner.py
veruu: Yes, that's exactly why there's the `if self.failures:` check before. So the check is skipped if there are no values and this quirk of `all()` doesn't bite us.

skt/runner.py
veruu: Not really. 'Panic' is a separate result (same as 'Pass' and 'Fail'), so doing so we wouldn't be reporting kernel panics. Also, 'Warn' itself is a valid test result. Only the combination of 'Warn' and 'Aborted' should show infrastructure failures.

I totally understand if this is confusing. I wasn't able to find any useful Beaker documentation and had to dig through a bunch of stuff to find out what's going on.

skt/runner.py
spbnick: I'm blind. I looked and looked for the check, and still didn't find it :grin:.

skt/reporter.py
veruu: The resubmission of a single aborted job is already done (by that code that tries to rule out host-specific failures). What we are doing here is just not reporting it since it's not useful for the developer. Does that answer you question?

skt/runner.py
spbnick: > I totally understand if this is confusing. I wasn't able to find any useful Beaker documentation and had to dig through a bunch of stuff to find out what's going on.

Perhaps we should write our own documentation, then. Right here, in the code, maybe?

Otherwise, how about we rewrite it this way, then:
```
if result == "Pass":
    ret = 0
elif result != "Warn" or status != "Aborted":
    ret = 1
else:
    ret = 2
```
and perhaps narrow the "Fail" (`ret = 1`) condition down a bit, if we can?
We won't have to initialize `ret` too.

skt/reporter.py
spbnick: Rriiight, so do you mean that we still send the message, but don't include this job's result into the report?

spbnick: Or rather, does skt return 1 or 2, if an aborted (failed due to infrastructure) job is detected?

skt/runner.py
veruu: The warn/aborted condition has to be `and`, not `or`, but otherwise I'm fine with rewriting it to this style.

skt/reporter.py
veruu: > Rriiight, so do you mean that we still send the message, but don't include this job's result into the report?

Yes.

> Or rather, does skt return 1 or 2, if an aborted (failed due to infrastructure) job is detected?

Do you mean the runner stage now? If all the jobs abort, 2. If the aborts are host-specific, then 0 (as per the host-specific resubmission logic)

skt/runner.py
spbnick: Hmm, I just reverted the condition using the `!(A&&B) => !A||!B` rule. Did I miss something?

skt/reporter.py
spbnick: How do we detect host-specific failure?

skt/runner.py
veruu: Oh, right. I got confused by the comparison of my version and this one, and didn't notice that the switch is setting 1 and not 2. Since it kind of makes more sense to me to have the specific condition in the `if ` and all the rest in the `else` (so `if specific infrastructure failure else 1`, if that makes sense).

It should work then, thanks for checking me :)

spbnick: OK, great :) Now, can we narrow that "ret=1" condition further somehow?

skt/reporter.py
veruu: The resubmission logic is pretty confusing and I didn't dive into it, but here's what I gathered so far:

For all non-pass jobs, the host and failure data are saved. By default, all these jobs are then resubmitted to a random machine from the pool, as well as the same host which the job failed on. When the testing results are gathered (after all the jobs finished), the failure data is checked and if there's only one unique hostname for the job, it's a host-specific problem.

skt/runner.py
veruu: I don't think so. For the completed jobs, you can get 'Pass', 'Warn', 'Fail' and 'Panic'. 'Panic' is for kernel panics, which we should report, 'Fail' and 'Warn' are normal test results which mean something went wrong.

spbnick: Ah, OK, then. Do you know what's the difference between "Warn" and "Fail"? And why do we even have "status" in addition to "result"?

skt/reporter.py
spbnick: I'm thinking that an infrastructure failure shouldn't even count into the host-specific failure detection. I.e. it should be as if the job simply didn't run, and it should be restarted, no matter if it's the first job in the host failure detection routine, or any other one.

But to be sure, what exactly are we calling infrastructure failures here? What would be the actual reason for the condition we're checking here for?

skt/runner.py
veruu: What they mean is test-specific. The test author decides what s/he considers a warning and failure. For example, I've seen NFS tests which considered a warning if the tests completed, but the speed was under a specific threshold; and a failure if they didn't complete.

Well, 'result' is the result of the job and 'status' it's status. Results are those 'Pass', 'Fail', 'Warn' things, while status talks about whether the job completed or not - status can be eg. 'Completed' and 'Aborted', 'Scheduled' (most of them aren't relevant once the job is finished).

skt/reporter.py
veruu: > But to be sure, what exactly are we calling infrastructure failures here? What would be the actual reason for the condition we're checking here for?

I'd say things like "network is down so we can't connect anywhere", "Beaker is under maintenance so we can't submit any jobs", "somebody messed with lab controller and now all jobs abort before even starting" etc.

The fact that a single job aborts is IMHO not an infrastructure failure. It's annoying, sure, but that's why we have the check there.

skt/runner.py
spbnick: I feel there are some very dark corners there, so let's leave the condition at that for now. Thank you for all the explanations :)

skt/reporter.py
spbnick: Ah, I see, good, those all seem to be perfect "infrastructure failures", indeed.

What I'm saying is that those shouldn't be the basis for host failure check. Because the fact that such failure repeats or doesn't repeat is not saying anything about the host. I would say we should then simply resubmit the job exactly as it is (whether it's a part of a host failure check or not), and if e.g. we get three infrastructure failures in a row, then give up with exit status of 2.

veruu: Oh, I see what you mean, I think. I'll add a check for that as well.

veruu:
@spbnick I added the max resubmission threshold and fixed the multireport summary (in case someone wants to ignore errors and run the reporting anyways)

spbnick:
OK, I think I spent enough time trying to understand the Beaker runner mess, and it gets only worse. I don't have the time and capacity (heat wave is tough on me) to get through to the end of it right now, so I'd like to ask you to make sure the following is true (highest priority first), and I'll have to trust you on that:

* [ ] If *any* (and I mean **any**) job submitted to Beaker fails due to an **infrastructure** failure, it is resubmitted, exactly as it was submitted the first time, up to three times (i.e. we retry it three times). If we get three infrastructure failures of any specific job, in a row, we consider the whole "skt run" failed due to infrastructure failure, and skt returns 2. This has **nothing** to do with host-specific failure detection. We can consider this as going a level *below* host-specific failure detection.
* [ ] If any job of those submitted by "skt run" is canceled, then skt exits **immediately** (yeah, abandon the rest of the jobs, or rather cancel them to be nice, if we can), with an appropriate error message, and exit status of 2.
* [ ] If we detect a host-specific failure, then we retry on another host (hope Beaker XML can have a `host != A && host !=B && ..` condition). If we detect three host-specific failures in a row, we can give up, declare an infrastructure failure, and exit with status 2.

Do you think the above makes sense? Do you think you can get at least the first one in this PR? The rest can go into issues and be fixed later.

Thank you.

veruu:
Okay, let me see if I can easily solve your points:

1. So you want to resubmit *each* aborted job up to 3 times? I thought 3 aborted jobs in general are enough (we only submit a single job and resubmit if it didn't pass so teechnically all the resubmissions come from the same original job already?)

2. Should be easy to implement. It would remove the testing aspect playing around with job submissions but I don't think we are using that anymore anyways.

3. If a job *fails* on a specific host, we resubmit it to one random host and the same host, there's no more magic to it with no more resubmissions. If it *aborts*, then it's not a failure and it counts as infrastructure failure (step 1.) Now that I think about it, we don't implement any "any host except X" logic in the code, just submit those 2 jobs. While I can implement that, I think it would complicate things with no added benefit. We can have a Beaker instance with a single eligible host so the resubmission to a different host would abort, then it will get submitted to a random host (without that condition) and run on that single host again, so we actually don't solve anything.

Let me know your thoughts on 1. and 3., and I'll solve 2 in the meanwhile

spbnick:
I think I understand where you're going with "if any three jobs got an infrastructure failure, then quit with status 2". This might be good enough, or even better than what I'm thinking about. I might be overdoing it. So, please do that if you think it works better with the code. What I want to avoid and am trying to say all this time is that we shouldn't take infrastructure failures as a test result or make any decisions based on them except should we give up testing or not. Because an infrastructure failure is an absence of test result, not a test result.

veruu:
Okay, let's break down the 1. - we submit `J:1`, that aborts, we submit `J:2` with a note that it's derived from `J:1`, it fails, we submit `J:3` which is again the same as `J:1`. At least that's how I understand it should be working. So the result would be the same but with more complicated code.

> What I want to avoid and am trying to say all this time is that we shouldn't take infrastructure failures as a test result or make any decisions based on them except should we give up testing or not. Because an infrastructure failure is an absence of test result, not a test result.

Agreed.

> We could keep it, but then if somebody cancels a job for a patch series developer sent, what do we do? Do we send them the result and say: sorry, we won't tell why, but one of your tests got canceled, see here what's left?

Added a commit for cancellation of the rest of the jobs and aborting.

spbnick:
Regarding 1., yes, that sounds pretty much what I meant. However, if you do it just "three infrastructure failures across all jobs", as I think you were implementing it before, I think it would be good enough, and if it keeps the code simpler for now, even better :)

Thanks!

spbnick:
I still feel we have a gap in understanding, but we'll have to work with that :)
The current code inspires no confidence in me, mostly because I don't understand it fully, but it'll have to wait its turn for thorough reading and improvement.

One last question before I try to test it. What would happen if the host-specific job (host==<SPECIFIC_HOST>) aborts with an infrastructure failure? Does it get resubmitted? Or does skt exit with ret=2? Or does something else happen?

veruu:
> The current code inspires no confidence in me, mostly because I don't understand it fully, but it'll have to wait its turn for thorough reading and improvement.

Agreed, I'd be very happy to simplify the runner (help welcome!) but it's so far down the priority list I'm not sure when I get to it :)

> What would happen if the host-specific job (host==<SPECIFIC_HOST>) aborts with an infrastructure failure? Does it get resubmitted? Or does skt exit with ret=2? Or does something else happen?

Infrastructure errors don't care if the job was bind to a specific host. The host-specific detection is for *test failures*, not infrastructure -- those are two separate problems. So what would happen depends on if we have reached the max number of resubmissions overall - we'd either abort with ret=2, or try to resubmit the job (to a random host, since we can argue that the host could have been taken out of the Beaker pool, which is an unexpected infrastructure error (as it already happened to us)).

Does this reflect what you'd like to happen? Oh, and thanks for poking into the logic, since it can help with finding better alternatives or bugs :)

spbnick:
OK, but does the check make sense then, if it's not on the same host?

Glad to hear you're getting something else besides frustration from my incessant questioning :)

veruu:
I'd say it does. If the failure is host-specific, then the job submitted to a different host shouldn't fail, right? If it does, it's a real problem.

spbnick:
OK, canceling the job during an "skt run" lead to this:
```
[x86_64] 2018-07-18 04:47:41,863     INFO   runner type: beaker
[x86_64] 2018-07-18 04:47:41,863     INFO   beaker template: beakerjob.xml
[x86_64] 2018-07-18 04:47:42,463     INFO   submitted jobid: J:2620751

[x86_64] 2018-07-18 04:47:42,856     INFO   added R:5396530 to watchlist

[x86_64] 2018-07-18 05:06:50,943     INFO   R:5396530 status changed to 'Cancelled', removing from watchlist
[x86_64] 2018-07-18 05:06:50,943    ERROR   Cancelled job detected! Cancelling the rest of running jobs and aborting!
[x86_64] Usage: bkr job-cancel [options] [J:<id> | RS:<id> ...]
[x86_64] 
[x86_64] bkr: error: Please specify a taskspec to cancel
[x86_64] 2018-07-18 05:06:51,198     INFO   Failed to cancel the remaining jobs
[x86_64] 2018-07-18 05:06:51,515    DEBUG   state: jobid_0 -> J:2620751
[x86_64] 2018-07-18 05:06:51,515    DEBUG   state: retcode -> 2
```

And then when reporting:
```
2018-07-18 05:07:05,291     INFO   runner type: beaker
2018-07-18 05:07:05,291     INFO   beaker template: beakerjob.xml
2018-07-18 05:07:05,686     INFO   Failure detected in recipe R:5396530, attaching console log
  File "/home/worker/runner/workspace/test-nkondras@2/virtualenv/bin/skt", line 11, in <module>
    sys.exit(main())
  File "/home/worker/runner/workspace/test-nkondras@2/virtualenv/lib/python2.7/site-packages/skt/executable.py", line 906, in main
    args.func(cfg)
  File "/home/worker/runner/workspace/test-nkondras@2/virtualenv/lib/python2.7/site-packages/skt/executable.py", line 363, in cmd_report
    reporter.report()
  File "/home/worker/runner/workspace/test-nkondras@2/virtualenv/lib/python2.7/site-packages/skt/reporter.py", line 767, in report
    msg.attach(MIMEText(self._get_multireport()))
  File "/home/worker/runner/workspace/test-nkondras@2/virtualenv/lib/python2.7/site-packages/skt/reporter.py", line 628, in _get_multireport
    results += self.__getjobresults()
  File "/home/worker/runner/workspace/test-nkondras@2/virtualenv/lib/python2.7/site-packages/skt/reporter.py", line 513, in __getjobresults
    result += ['    ' + ltp_log for ltp_log in ltp_logs]
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

Could you please fix that?

veruu:
I'll check why the job cancellation fails. The reporting is expected to fail since it's not expected to be ran after runner exited with code 2 - we said that we shouldn't be reporting canceled jobs or infrastructure failures.

spbnick:
Right, we shouldn't be reporting that. Working on implementing that right now. Would be nice to have a clearer error message, but that can be another PR, or an issue even, to be fixed with reporting overhaul.

spbnick:
OK, canceled jobs are handled alright now, thanks :) Testing a bit more.

spbnick:
Correction: I tested only passing jobs and canceled jobs. We have no means of testing for infrastructure failures, so we'll have to wait and see.

PR #282, opened 2018-07-18
Enhance string concatenations for url and file path

danghai:
I make PR to enhance string concatenations for url and file path for `skt`

### DIFF
--- a/skt/__init__.py
+++ b/skt/__init__.py
@@ -15,10 +15,29 @@
 from email.errors import HeaderParseError
 import email.header
 import email.parser
-import os
 import re
 
 import requests
+
+
+def join_with_slash(base, *suffix_tuple):
+    """
+    Join parts of URL or path by slashes. Trailing slash of base, and each
+    arg in suffix_tupple are removed. It only keeps trailing slash at the
+    end of the part if it is specified.
+
+    Args:
+        base:          Base URL or path.
+        *suffix_tuple: Tuple of suffixes
+
+    Returns:
+        The URL or path string
+    """
+    parts = [base.rstrip('/')]
+    for arg in suffix_tuple:
+        parts.append(arg.strip('/'))
+    ending = '/' if arg.endswith('/') else ''
+    return '/'.join(parts) + ending
 
 
 def get_patch_mbox(url):
@@ -34,9 +53,7 @@ def get_patch_mbox(url):
     Raises:
         Exception in case the URL is currently unavailable or invalid
     """
-    # Use os.path for manipulation with URL because urlparse can't deal
-    # with URLs ending both with and without slash.
-    mbox_url = os.path.join(url, 'mbox')
+    mbox_url = join_with_slash(url, 'mbox')
 
     try:
         response = requests.get(mbox_url)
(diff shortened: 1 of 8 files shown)


danghai:
@veruu It is fail because of `__glob_escape` method. I am not sure about what this method does? Could you help me clarify it?

veruu:
@danghai no. It's because in the kernelbuilder tests, the kbuilder is initialized without setting `rh_configs_glob`, which is then set to `None` by default. You'll need to modify those two tests which use it and overwrite the kbuilder used in them with one that is setting the attribute.

danghai:
@veruu Thanks. I fixed it. But I am still confused because it is not fail before using this function. Do you know why?

veruu:
@danghai previously `.format` was used, and it implicitly formats `None` object into a string. The new function doesn't retype things so the tests needed to be changed to reflect it.

danghai:
@veruu I replace them. I am not sure what the function `__glob_escape` does? For example: 
`pathname = redhat/configs/kernel-*-x86_64.config`
--> The output will be: `redhat/configs/kernel-[*]-x86_64.config`. 
I do not get `[*]` ? Could you help me clarify? Thanks

veruu:
@danghai check out the glob module documentation which explains the reasoning.

skt/__init__.py
```diff
@@ -15,12 +15,27 @@
...
 
 import requests
 
 
+def join_with_slash(base, *suffix):
+    """
+    Join parts of URL or path by slashes
+    Args:
+        base:    Base URL or path.
+        *suffix: Array of suffixes
```
spbnick: Could we name this argument `suffix_tuple`, perhaps? So it's clear it's a sequence, and the iteration below would look clearer? Also, could you please document the slash-stripping behavior, and keeping the trailing slash in particular? This is the main value of this function and it's not documented.

tests/test_kernelbuilder.py
```diff
@@ -249,6 +250,7 @@ def test_redhat_config_glob_failure(self, mock_check_call, mock_glob,
         # pylint: disable=W0212,E1101
         mock_check_call.return_value = ''
         mock_glob.return_value = []
+        self.kbuilder.rh_configs_glob = "redhat/configs/kernel-*-x86_64.config"
```
spbnick: Why are we adding these?

veruu: This was my idea. The tests use a minimal configuration of kbuilder which doesn't have the glob set, but it tries to test the `__make_redhat_config` too. If the glob is not set, the method tries to format `None` into the config path. The tests were working in the previous versions by sheer luck that `format` retypes the `None` object to string, but our new method doesn't retype things so it broke the tests (which shouldn't have been working in the first place, since we shouldn't be testing a method without setting the needed arguments).

So this addition ensures the `__make_redhat_config` is being tested properly, and fixes the travis failure since the test issue was discovered by this change.

skt/__init__.py
danghai: @veruu do you agree with his suggestion?

veruu: @danghai yes, sounds good. But if you do the renaming, please also change the docstring to say `Tuple of suffixes` instead of array, for consistency reasons

danghai:
@veruu I put some information, and rebase this pull. Could you review now?

PR #286, opened 2018-07-23
some separate stage outputs and fixes

veruu:
Originally wanted to add only the separate stage outputs as per meta/6 but found a bunch of issues on the way here and the pull grew.

Still missing the separate runner output but that one will follow as this shows to be working and merged

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -179,7 +179,7 @@ def cmd_merge(cfg):
             for mb in cfg.get('merge_ref'):
                 save_state(cfg, {'mergerepo_%02d' % idx: mb[0],
                                  'mergehead_%02d' % idx: bhead})
-                (retcode, _) = ktree.merge_git_ref(*mb)
+                (retcode, bhead) = ktree.merge_git_ref(*mb)
 
                 utypes.append("[git]")
                 if retcode:
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -304,10 +304,10 @@ def cmd_build(cfg):
             except OSError:
                 pass
 
-    report_string = ''
     build_result_path = os.path.join(cfg.get('workdir'), 'build.result')
     build_report_path = os.path.join(cfg.get('workdir'), 'build.report')
 
+    tgz = None
     builder = KernelBuilder(
         source_dir=cfg.get('workdir'),
         basecfg=cfg.get('baseconfig'),
@@ -328,17 +328,17 @@ def cmd_build(cfg):
         logging.error(exc)
         save_state(cfg, {'buildlog': builder.buildlog})
         retcode = 1
-        return
     except Exception:
         (exc, exc_type, trace) = sys.exc_info()
         raise exc, exc_type, trace
 
-    if cfg.get('buildhead'):
-        ttgz = "%s.tar.gz" % cfg.get('buildhead')
-    else:
-        ttgz = addtstamp(tgz, tstamp)
-    shutil.move(tgz, ttgz)
-    logging.info("tarball path: %s", ttgz)
+    if tgz:
+        if cfg.get('buildhead'):
+            ttgz = "%s.tar.gz" % cfg.get('buildhead')
+        else:
+            ttgz = addtstamp(tgz, tstamp)
+        shutil.move(tgz, ttgz)
+        logging.info("tarball path: %s", ttgz)
 
     tbuildinfo = None
     if cfg.get('buildinfo'):
@@ -363,6 +363,20 @@ def cmd_build(cfg):
     with os.fdopen(os.open(build_result_path, os.O_CREAT | os.O_WRONLY),
                    'w') as result_file:
         result_file.write('true' if not retcode else 'false')
+
+    report_string = '{} ({{{}}})'.format(
+        'The kernel was built with the attached configuration',
+        tconfig
+    )
+    if retcode:
+        report_string += '\n'.join([
+            '\nHowever, the build failed. We are attaching the build output '
+            'for',
+            'more information ({{{}}})'.format(
+                os.path.basename(builder.buildlog)
+            )
+        ])
+
     with os.fdopen(os.open(build_report_path, os.O_CREAT | os.O_WRONLY),
                    'w') as report_file:
         report_file.write(report_string)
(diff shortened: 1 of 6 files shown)


skt/executable.py
```diff
@@ -218,6 +256,18 @@ def cmd_build(cfg):
...
                                         "%Y%m%d%H%M%S")
 
+    # Remove existing results from previous runs (if any)
+    for filename in os.listdir(cfg.get('workdir')):
+        if filename.startswith('build.'):
+            try:
+                os.unlink(os.path.join(cfg.get('workdir'), filename))
+            except OSError:
+                pass
+
```
dzickusrh: I understand this is a safety thing, but have we seen or expect leftover files?  Doesn't each run have its own directory, meaning this can/should never happen?

skt/executable.py
```diff
@@ -334,6 +404,14 @@ def cmd_run(cfg):
 
     save_state(cfg, {'retcode': retcode})
 
+    if retcode != 2:
```
dzickusrh: Silly nitpick.  What is retcode 2?

skt/executable.py
```diff
@@ -328,17 +328,17 @@ def cmd_build(cfg):
...
-        ttgz = addtstamp(tgz, tstamp)
-    shutil.move(tgz, ttgz)
-    logging.info("tarball path: %s", ttgz)
+    if tgz:
+        if cfg.get('buildhead'):
+            ttgz = "%s.tar.gz" % cfg.get('buildhead')
+        else:
+            ttgz = addtstamp(tgz, tstamp)
+        shutil.move(tgz, ttgz)
+        logging.info("tarball path: %s", ttgz)
```
dzickusrh: The changelog is a little dry.  Is this shift for tgz to allow the failure path to receive the same comments as the success path?  In other words, what is the goal of this patch?

skt/executable.py
```diff
@@ -241,7 +241,8 @@ def cmd_merge(cfg):
         with open(ktree.mergelog, 'r') as mergelog:
             for line in mergelog:
                 # Skip the useless part of the 'git am' output
-                if "The copy of the patch" in line:
+                if ("The copy of the patch" in line) \
+                        or ('see the failed patch' in line):
```
dzickusrh: This seems fine, but in the future an example output change in the changelog would be a helpful reference.  Otherwise I have to reproduce this failure manually and double check what you are filtering is indeed useless.  As a reviewer I need to double check the stated opinion sometimes. :-)

skt/executable.py
veruu: This is more for running things by hand, not from pipeline. In the pipeline, new directories should be created for the runs. But if people (eg. the developers, or us when testing things out) would like to be able to reproduce the runs, they probably won't bother with creating new directories and deleting the old files. The result and report files would be overwritten anyways, this just makes it explicit that you shouldn't rely on the old output if you decide to use the same directory.

skt/executable.py
veruu: Just an agreement between us developers - 0 for success, 1 for failure and 2 for infrastructure errors.

skt/executable.py
```diff
@@ -179,7 +179,7 @@ def cmd_merge(cfg):
             for mb in cfg.get('merge_ref'):
                 save_state(cfg, {'mergerepo_%02d' % idx: mb[0],
                                  'mergehead_%02d' % idx: bhead})
-                (retcode, _) = ktree.merge_git_ref(*mb)
+                (retcode, bhead) = ktree.merge_git_ref(*mb)
```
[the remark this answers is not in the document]
veruu: That's not `skt merge` but `git merge`. Same as we allow to apply multiple patches at the same time, we allow multiple git merges into the tree. To be fair I'm not sure if the option was ever used but since it's there, it should at least be fixed.

skt/executable.py
dzickusrh: Ah, ok.  That makes sense then.  Thanks!

skt/executable.py
veruu: Yes. For both success and failure, we want to include the kernel config and similar report message. So for adding the full stage reporting, we can either duplicate a bunch of the code or just check if the kernel was built and deal with that part in the if clause, if that makes sense.

skt/executable.py
dzickusrh: Ok.  Maybe someday we can wrap that with something more informational..

skt/executable.py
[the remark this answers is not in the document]
dzickusrh: Hehe.  Fair enough.

skt/executable.py
veruu: Makes sense! I'll add examples of the new message found to the commit message

skt/executable.py
dzickusrh: No, it makes sense, and the code seems to point to that.  Just remember to add that to the changelog next time.

skt/executable.py
veruu: @dzickusrh added the example and reasoning in the commit message, thanks for the idea!

skt/executable.py
```diff
@@ -163,10 +175,21 @@ def cmd_merge(cfg):
             (retcode, _) = ktree.merge_git_ref(*mb)
 
             utypes.append("[git]")
-            idx += 1
             if retcode:
+                if retcode != 2:
```
danghai: @veruu Why dont' you use the string constant instead of these numbers. It is easier to read and makes more sense?

veruu: @danghai good idea! I'll open an issue for it, since it's not really a part of this series.

skt/executable.py
dzickusrh: Looks great thanks!

skt/executable.py
danghai: @veruu one more is why do not take both if condition into one. if `retcode !=0 and recode !=2`. Or you will add some thing in future?

veruu: Two reasons - this allows us to implement more return codes if we need in the future and not change too many things around; and we want the `return` that is below to execute if any failure occurs, not only infrastructure error, so we can't really merge the conditions anyways.

PR #289, opened 2018-07-23
Using skt.misc instead of constant number for retcode

### LINKED ISSUE #287: Use constants for success/fail/error instead of 0/1/2
As @danghai pointed out in one of the pulls, it would be more clear to understand than plain numbers.

Implement some simple constants like `SKT_SUCCESS = 0`, `SKT_FAIL = 1`, `SKT_ERROR = 2`, and then use these across the code instead of the direct numeric values where required.

danghai:
I make PR to use `skt.misc` instead of constant number for retcode.
Fix #287

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -36,10 +36,11 @@ import skt.reporter
 import skt.runner
 from skt.kernelbuilder import KernelBuilder, CommandTimeoutError, ParsingError
 from skt.kerneltree import KernelTree, PatchApplicationError
+from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR
 
 DEFAULTRC = "~/.sktrc"
 LOGGER = logging.getLogger()
-retcode = 0
+retcode = SKT_SUCCESS
 
 
 def full_path(path):
@@ -86,15 +87,16 @@ def junit(func):
     set to JSON representation of the configuration object after the function
     call has completed. The created test case is appended to the "_testcases"
     list in the configuration object after that. Sets the global "retcode" to
-    1 in case of test failure and 2 and above in case of infrastructure failure
-    or skt problem (eg. Beaker server is unreachable).
+    SKT_SUCCESS in case of test success, SKT_FAIL in case of test failure, and
+    SKT_ERROR and above in case of infrastructure failure or skt problem (eg.
+    Beaker server is unreachable).
 
     Args:
         func:   The function to call in the created function. Must accept
                 a configuration object as the argument. Return value would be
                 ignored. Can set the global "retcode" to indicate success
-                (zero), test failure (1) or infrastructure failure (2 and
-                above).
+                (SKT_SUCCESS), test failure (SKT_FAIL) or infrastructure
+                failure (SKT_ERROR and above).
 
     Return:
         The created function.
@@ -112,13 +114,13 @@ def junit(func):
                               "infrastructure failure or skt bug: %s",
                               traceback.format_exc())
                 testcase.add_error_info(traceback.format_exc())
-                retcode = 2
+                retcode = SKT_ERROR
 
-            if retcode == 1:
+            if retcode == SKT_FAIL:
                 # Tests failed
                 testcase.add_failure_info("Step finished with retcode: %d" %
                                           retcode)
-            elif retcode >= 2:
+            elif retcode >= SKT_ERROR:
                 testcase.add_error_info(
                     "Infrastructure issue or skt bug detected, retcode: %d" %
                     retcode
@@ -183,7 +185,7 @@ def cmd_merge(cfg):
 
                 utypes.append("[git]")
                 if retcode:
-                    if retcode != 2:
+                    if retcode != SKT_ERROR:
                         with os.fdopen(os.open(merge_result_path,
                                                os.O_CREAT | os.O_WRONLY),
                                        'w') as result_file:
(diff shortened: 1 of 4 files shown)


danghai:
@veruu So you want me to remove the class Result. Just declare the constants. It probably uses `skt.misc.SKT_` ?

veruu:
Yes, just declaring the constants would be much easier and understandable. And then do `from sktm.misc import <LIST OF CONSTANTS>` and use the constant name directly, instead of using the `skt.misc` prefix on each line (since the prefix itself is not very useful)

danghai:
@veruu I have fixed it.

skt/runner.py
```diff
@@ -19,6 +19,7 @@
 import subprocess
 import time
 import xml.etree.ElementTree as etree
+from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR
```
veruu: Can you add a newline before this line, to divide the import groups?

skt/runner.py
```diff
@@ -135,7 +136,7 @@ def dumpjunitresults(self, jobid, junit):
         with open(fname, 'w') as fileh:
             bkr = subprocess.Popen(args, stdout=fileh)
             bkr.communicate()
-        if bkr.returncode != 0:
+        if bkr.returncode != SKT_SUCCESS:
```
veruu: This one is a subprocess return code which shouldn't be compared with skt results.

skt/runner.py
```diff
@@ -292,30 +294,31 @@ def __getresults(self, jobid=None):
...
-            infrastructure failures.
+            SKT_SUCCESS if all jobs passed,
+            SKT_FAIL in case of failures, and
+            SKT_ERROR in case of infrastructure failures.
         """
-        ret = 0
+        ret = SKT_SUCCESS
         fhosts = set()
-        tfailures = 0
+        tfailures = SKT_SUCCESS
```
veruu: This variable represents number of test failures and not a return code

skt/runner.py
```diff
@@ -397,7 +400,7 @@ def __watchloop(self):
                                       ' rest of running jobs and aborting!')
                         self.failures[cid] = [[root.attrib.get('system')],
                                               set([(result, status)]),
-                                              1]
+                                              SKT_FAIL]
```
veruu: Again here, the last element of the list represents number of failed test runs and not the return code. Yes, we need to clean up the runner :)

skt/runner.py
```diff
@@ -432,7 +435,7 @@ def __watchloop(self):
                                 origin = cid
 
                             if origin not in self.failures:
-                                self.failures[origin] = [[], set(), 1]
+                                self.failures[origin] = [[], set(), SKT_FAIL]
```
veruu: Same as above

skt/runner.py
```diff
@@ -528,15 +531,15 @@ def run(self, url, release, wait=False, host=None, uid="",
...
+            SKT_FAIL if testing failed
+            SKT_ERROR in case of infrastructure error (exceptions are logged)
         """
-        ret = 0
+        ret = SKT_SUCCESS
         self.failures = {}
         self.recipes = set()
         self.watchlist = set()
-        self.aborted_count = 0
+        self.aborted_count = SKT_SUCCESS
```
veruu: As the variable name suggests, the variable contains a count of aborted jobs, not a return code

skt/kerneltree.py
```diff
@@ -334,7 +335,7 @@ def merge_patchwork_patch(self, uri):
         status, output = self.__git_cmd_pipe(patch_content,
                                              "am", "-",
                                              cwd=self.wdir)
-        if status != 0:
+        if status != SKT_SUCCESS:
```
veruu: Status here is the return code from subprocess call, so this one shouldn't be compared with the result codes

skt/executable.py
```diff
@@ -31,15 +31,17 @@
 import junit_xml
 
 import skt
+import skt.misc
```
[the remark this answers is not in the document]
danghai: I forget removing it.

danghai:
@veruu I fixed it!

skt/executable.py
```diff
@@ -34,12 +34,14 @@
 import skt.publisher
 import skt.reporter
 import skt.runner
+
```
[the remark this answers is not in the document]
danghai: Oh sorry, I think it is a import group. That is my mistake. I fix now

skt/kerneltree.py
```diff
@@ -20,6 +20,8 @@
 
 import skt
 
+from skt.misc import SKT_SUCCESS, SKT_FAIL
```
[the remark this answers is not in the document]
danghai: Do you mean no new line between `import skt` and `from skt.misc import SKT_SUCCESS, SKT_FAIL`?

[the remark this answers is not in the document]
veruu: Yes. In general, there are 3 import groups: standard modules, third party modules, and local modules. skt is a local module, so both skt imports should be together in the third group.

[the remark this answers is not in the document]
danghai: Oh thanks. I never know about this standard before. I fixed it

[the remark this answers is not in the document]
veruu: You removed the wrong newline :) the newline in runner should be there, the one that should have been removed is in this kerneltree file :)

danghai:
@veruu I fixed it!

danghai:
@veruu Thanks. I fixed it.

PR #293, opened 2018-07-25
Wrap the whole removal into try-except block

veruu:
There are some fun setups with workers not having access to the actual
workdir.


dzickusrh:
I always like to see more context in the changelog as I was scratching my head about why the change was needed.  It looked like random code cleanup.  But the PR had some of the context I was looking for (and sort of expected).

Trivial code movement.

veruu:
@dzickusrh you have to open the commit via the "commits" tab to see full message (which is in this case same as the pull request description), if that's what got you confused. I know I have similar problem sometimes

PR #294, opened 2018-07-26
BeakerRunner.dumpjunitresult() accept output filename

### LINKED ISSUE #273: BeakerRunner.dumpjunitresults() should accept output filename
At the moment, `BeakerRunner.dumpjunitresults()` accepts a directory to write the junit file to, and it just assumes that it can generate a sufficiently distinct filename to not clash with existing files. It cannot know that, since it's not actually creating the directory and filling it with files. Instead, the caller should decide how the output file should be named, and the function should accept a full name of the file to write to.

danghai:
I make PR for `BeakerRunner.dumpjunitresult()` accepting the output file name
Fixes #273

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -437,7 +437,9 @@ def cmd_run(cfg):
     for job in runner.jobs:
         if cfg.get('wait') and cfg.get('junit'):
             try:
-                runner.dumpjunitresults(job, cfg.get('junit'))
+                fname = "%s/%s.xml" % (cfg.get('junit'),
+                                       job.replace(":", "_").lower())
+                runner.dumpjunitresults(job, fname)
             except Exception as exc:
                 logging.error(exc)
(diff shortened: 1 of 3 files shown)


danghai:
@veruu yeah. I am not sure how to fix `flake8` violation when I use the new join function. For the test file, I see it only test the file create in this directory or not? And I do not know what I need to test when I pass the filename argument.

veruu:
I can't help you if you don't mention the error you get ;) For the test, I'd say you should create a filename inside the junit directory and pass that instead of the directory, and change the name of the expected file to match that.

danghai:
@veruu Thanks review. It is a good idea for me to write the test case. I think I am going to wait the opinion's @spbnick for this patch first.

danghai:
Because `dumpjunitresult()` method is removed, so I close this PR now.

PR #295, opened 2018-07-26
move out code from __init__.py to misc.py.

### LINKED ISSUE #291: Move out code from __init__.py
We now have the misc.py file which should be the home of the functions. init file's purpose is to mark the module, not to contain a bunch of code, and the file isn't even needed with the newer versions of Python so let's get rid of its contents.

danghai:
I make PR to move out code from `__init__.py` to `misc.py`.
Fix #291


danghai:
@veruu okay it makes sense. It is not useful for `misc` function.  It answers my question in my head that sometime people use module path while it is long. Sometime, it is not useful. I will fix it. Thanks

danghai:
@veruu It has some error, and I am not sure how to fix it. Could you help me figure out? 
The error:
```
  File "/home/travis/build/RH-FMK/skt/tests/test_kerneltree.py", line 235, in test_merge_pw_patch
    result = self.kerneltree.merge_patchwork_patch('uri')
  File "skt/kerneltree.py", line 312, in merge_patchwork_patch
    patch_content = get_patch_mbox(uri)
  File "skt/misc.py", line 99, in get_patch_mbox
    raise exc
MissingSchema: Invalid URL 'uri/mbox': No schema supplied. Perhaps you meant http://uri/mbox?
```

veruu:
@danghai since you are importing the methods with `from skt.misc import <method_name>`, you can't use the path of the method definition for mocking. So if you are testing methods from kerneltree module, and you have imported the method there, you have to use `skt.kerneltree.method_name> instead of `skt.misc.method_name`. Check out the [documentation](https://docs.python.org/3/library/unittest.mock.html#id5) for explanation of why it is needed

danghai:
@veruu Thanks for helping me figure out this. However, I am confused about this line: https://github.com/RH-FMK/skt/blob/master/tests/test_kerneltree.py#L227
why ` skt.get_patch_name` is inside the `merge_patchwork_patch`. I mean in the `merge_patchwork_patch` it does not call any `get_patch_name`.

veruu:
Looks like a residue of previous functionality. The code that was using the method was just recently removed. If you want, you can make a patch that removes the mock too

danghai:
Okay could I create another PR to remove this, and then merge first before update this PR?

veruu:
Yes, sounds good.

danghai:
@veruu I resolve the conflict. Could you review now?

skt/executable.py
```diff
@@ -36,6 +36,7 @@
 import skt.runner
 from skt.kernelbuilder import KernelBuilder, CommandTimeoutError, ParsingError
 from skt.kerneltree import KernelTree, PatchApplicationError
+from skt.misc import join_with_slash, get_patch_name, get_patch_mbox
```
veruu: Can you use a single import statement with parenthesis, as PEP328 suggests, instead of using two skt.misc import lines?

skt/kerneltree.py
```diff
@@ -18,7 +18,7 @@
 import shutil
 import subprocess
 
-import skt
+from skt.misc import join_with_slash, get_patch_mbox
```
veruu: Same here, if the change doesn't fit on a single line

skt/reporter.py
```diff
@@ -26,6 +26,7 @@
 
 import skt
 import skt.runner
+from skt.misc import join_with_slash, get_patch_name, get_patch_mbox
```
veruu: Is it possible to put this line above the runner import, to keep the alphabetical order?

skt/__init__.py
```diff
@@ -1,99 +0,0 @@
...
-# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
-# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
-# details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software Foundation, Inc.,
-# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-"""Functions used by multiple parts of skt."""
-from email.errors import HeaderParseError
-import email.header
```
veruu: The email.header should be moved with the `get_patch_name` function, not `get_patch_mbox` one. Can you move it to that commit please?

skt/misc.py
```diff
@@ -11,9 +11,95 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Functions used by multiple parts of skt."""
```
veruu: If you want to add a docstring, can it read "Functions and constants", since, well, it's not only functions in the module?

danghai:
@veruu Thanks review. I update the change.

skt/reporter.py
```diff
@@ -25,6 +25,7 @@
 import requests
 
 import skt
+from skt.misc import join_with_slash, get_patch_name, get_patch_mbox
```
veruu: If you check out the separate commits, the lines are still in the wrong order in the first two.

skt/kerneltree.py
```diff
@@ -18,8 +18,7 @@
 import shutil
 import subprocess
 
-import skt
-from skt.misc import SKT_SUCCESS, SKT_FAIL
+from skt.misc import join_with_slash, get_patch_mbox, SKT_SUCCESS, SKT_FAIL
```
veruu: If you check the join_with_slash commit, there are still two import lines kept.

skt/executable.py
```diff
@@ -36,7 +36,8 @@
 import skt.runner
 from skt.kernelbuilder import KernelBuilder, CommandTimeoutError, ParsingError
 from skt.kerneltree import KernelTree, PatchApplicationError
-from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR
+from skt.misc import (join_with_slash, get_patch_name, get_patch_mbox,
```
veruu: If you check the first two commits, there are still two import lines kept.

tests/test_misc.py
```diff
@@ -0,0 +1,133 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Test cases for misc.py."""
+from __future__ import division
```
veruu: This line is a residue of kernel builder being in the init file. Since it's not needed here, it shouldn't be copied over.

danghai: You mean `import division`?

skt/executable.py
danghai: @veruu right. I will fix now

tests/test_misc.py
veruu: Yes.

danghai:
@veruu I fix it. :)

PR #298, opened 2018-07-27
trace call extraction enhancements

veruu:


skt/reporter.py
```diff
@@ -222,38 +250,28 @@ def getfulllog(self):
...
+            if self.start_pattern.search(line):
+                tmpdata = [line]
+            elif tmpdata:
+                if self.end_pattern.search(line):
+                    tmpdata.append(line)
+                    result.append('\n'.join(tmpdata))
+                    tmpdata = []
+                    continue
+                if self.continue_pattern.search(line):
+                    tmpdata.append(line)
```
dzickusrh: I am trying to understand this logic.  If we see a start pattern, start recording.  Makes sense.  If we see an end pattern, append final line, record to results, and reset tmpdata.  Makes sense.

I get lost on the 'if continue pattern' part.  It seems like we would randomly drop lines between start and end pattern if they do not match 'continue pattern'?  Basically what happens when 'if continue_pattern' fails?

PR #299, opened 2018-07-28
Add sftp support for publish 

danghai:
I make PR to add sftp support for `publish` to transfer the kernel tarball, configuration, and build information to the specified location. I also add docstring.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -382,7 +382,7 @@ def cmd_publish(cfg):
     """
     Publish (copy) the kernel tarball and configuration to the specified
     location, generating their resulting URLs, using the specified "publisher".
-    Only "cp" and "scp" pusblishers are supported at the moment.
+    Only "cp", "scp" and "sftp" pusblishers are supported at the moment.
 
     Args:
         cfg:    A dictionary of skt configuration.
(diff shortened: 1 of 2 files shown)


danghai:
@veruu I apply the `sftp` being the same as `scp` command. 
`scp <source> <destination>`.
`sftp <source> <destination>`.
I think `sftp` is the same as `scp`. So I just write a simple `subprocess` call being the same `scp`.

danghai:
@veruu yeah I tried before. But I save it another computer. I will look at it soon. What about the docstrings?

veruu:
The docstrings are OK. I can merge that patch if you submit it via a separate pull request (can't merge only selected patches from the pull).

danghai:
@veruu Okay, I create another PR #310 . I will look at this PR more.

danghai:
@veruu right, it is actually different. I may have the mistake before I use `sftp` to get the data from host. It may be the same way as `scp`. However, uploading/ transferring data from host computer to server is different. I update my commit. Thanks

danghai:
@veruu . Thanks. I forget update the `README` and fix one type though.  :(

veruu:
That can be done in another pull, no problem :)

PR #305, opened 2018-08-01
Set default in parser

danghai:
I make PR to set default in parser. I think by explicitly setting the default value in parser, it reduces the
`if condition` case in source code.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -616,6 +616,7 @@ def setup_parser():
     parser_merge.add_argument(
         "--ref",
         type=str,
+        default="master",
         help="Base repo ref to which patches are applied (default: master)"
     )
     parser_merge.add_argument(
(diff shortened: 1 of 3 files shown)


veruu:
Setting the default in the parser means we will ignore the options passed via the configuration file. Question is whether we don't mind it -- I think these specific options should be passed via command line only and the change is OK, but this needs more input from others. @major, what do you think?

danghai:
@veruu oh I see.  The options in configuration file will not work in this case. I remember ( not sure) Nick told me he probably removes/changes configuration file in the future.

danghai:
It may require more input rather from command line rather than configuration file. What do you think @major ?

major:
If I remember correctly, @spbnick was heavily against this change (I proposed something similar a while back) because we still have a mess with the config file and command line argument precedence. Perhaps we should fix that mess first? :)

spbnick:
Yeah, we cannot have defaults in the parser, because we have configuration. We need to remove (most of) it, and then we should be able to do it. Closing for now. Thank you, Hai!

PR #307, opened 2018-08-02
Implement a blacklist for Beaker machines

veruu:
Sometimes, running on a specific machine is not desired. This patch
allows to list hostnames of those machines in a text file, which are
then included in the submitted Beaker XML, ensuring the tests are not
ran on these machines.


veruu:
Can't wait to not see the json either. Merging then, thanks for checking!

PR #308, opened 2018-08-03
Support ordered mixture of merge arguments

danghai:
I make PR to support ordered mixture of merge arguments. It is my idea and first draft. 
Because the optional argument for merge is not positional, so I introduce the `CustomAction` to get the order of merge option ( `--merge-ref`, `--patch` and `--pw`) . 
For example:  
if we use `--pw <PATCH1> --patch <PATCH2> --merge-ref <PATCH3> --pw <PATCH4>`
The `order_merge` stores the merge arguments by order. 
```
[('pw', 'PATCH1'), ('patch', 'PATCH2'), ('merge-ref', 'PATCH3'), ('pw', 'PATCH4')]

```
I declare the `task` in `cmd_merge` following:
```
task[0]: merge argument
task[1]: merge patch
The tasks will store: 
('pw', 'PATCH1')
('patch', 'PATCH2')
('merge-ref', 'PATCH3')
('pw', 'PATCH4')
```
It still needs to be enhanced by changing the `report_string` in order to get all information all patch. It also needs to fix the value `idx`. What do you think about this idea? Thanks

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -45,6 +45,15 @@ LOGGER = logging.getLogger()
 retcode = SKT_SUCCESS
 
 
+class AppendMergeArgument(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        if 'merge_queue' not in namespace:
+            setattr(namespace, 'merge_queue', [])
+        previous = namespace.merge_queue
+        previous.append((self.dest, values))
+        setattr(namespace, 'merge_queue', previous)
+
+
 def full_path(path):
     """Get an absolute path to a file."""
     return os.path.abspath(os.path.expanduser(path))
@@ -180,7 +189,12 @@ def cmd_merge(cfg):
         cfg:    A dictionary of skt configuration.
     """
     global retcode
-    idx = 0
+    # Counter merge patch following:
+    # id[0]:  counter of merge-ref option.
+    # idx[1]: counter of patch option.
+    # idx[2]: counter of pw option.
+    idx = [0, 0, 0]
+    previous_merge = ''
     ktree = KernelTree(
         cfg.get('baserepo'),
         ref=cfg.get('ref'),
@@ -203,15 +217,15 @@ def cmd_merge(cfg):
     merge_report_path = join_with_slash(cfg.get('output_dir'),
                                         'merge.report')
 
-    try:
-        if cfg.get('merge_ref'):
-            report_string = '\n{}\n{}'.format(
-                report_string,
-                'We merged the following references into the tree:'
-            )
-            for mb in cfg.get('merge_ref'):
-                save_state(cfg, {'mergerepo_%02d' % idx: mb[0],
-                                 'mergehead_%02d' % idx: bhead})
+    for thing_to_merge in cfg.get('merge_queue', []):
+        try:
+            if thing_to_merge[0] == 'merge_ref':
+                if previous_merge != 'merge_ref':
+                    report_string += '\nWe merged the following references ' \
+                        'into the tree:'
+                mb = thing_to_merge[1].split()
+                save_state(cfg, {'mergerepo_%02d' % idx[0]: mb[0],
+                                 'mergehead_%02d' % idx[0]: bhead})
                 (retcode, bhead) = ktree.merge_git_ref(*mb)
 
                 if retcode:
@@ -220,68 +234,65 @@ def cmd_merge(cfg):
                                        merge_report_path, report_string)
 
                     return
-
                 report_string += '\n  - %s\n    into commit %s' % (mb[0],
                                                                    bhead[:12])
-                idx += 1
+                previous_merge = thing_to_merge[0]
+                idx[0] += 1
 
-        elif cfg.get('patch') or cfg.get('pw'):
-            report_string = 'We applied the following patch(es):'
-
-            if cfg.get('patch'):
-                for patch in cfg.get('patch'):
-                    patch = os.path.abspath(patch)
-                    save_state(cfg, {'localpatch_%02d' % idx: patch})
-                    report_string += '\n  - %s\n' % patch
+            else:
+                if previous_merge not in ['patch', 'pw']:
+                    report_string += '\nWe applied the following patch(es):'
+                if thing_to_merge[0] == 'patch':
+                    patch = os.path.abspath(thing_to_merge[1])
+                    save_state(cfg, {'localpatch_%02d' % idx[1]: patch})
+                    report_string += '\n  - %s' % patch
 
                     ktree.merge_patch_file(patch)
-                    idx += 1
+                    idx[1] += 1
 
-            elif cfg.get('pw'):
-                for patch in cfg.get('pw'):
-                    save_state(cfg, {'patchwork_%02d' % idx: patch})
+                elif thing_to_merge[0] == 'pw':
+                    patch = thing_to_merge[1]
+                    save_state(cfg, {'patchwork_%02d' % idx[2]: patch})
                     report_string += '\n  - %s,\n' % get_patch_name(
                         get_patch_mbox(patch)
                     )
                     report_string += '    grabbed from %s' % patch
 
                     ktree.merge_patchwork_patch(patch)
-                    idx += 1
+                    idx[2] += 1
+
+                previous_merge = thing_to_merge[0]
+
+        except PatchApplicationError as patch_exc:
+            retcode = SKT_FAIL
+            logging.error(patch_exc)
+            save_state(cfg, {'mergelog': ktree.mergelog})
 
             report_string += '\n'.join([
-                'on top of commit %s from the repository at' % bhead[:12],
-                '  %s' % cfg.get('baserepo')
+                '\nHowever, the application of the last patch above failed '
+                ' with the',
+                'following output:\n\n'
             ])
-    except PatchApplicationError as patch_exc:
-        retcode = SKT_FAIL
-        logging.error(patch_exc)
-        save_state(cfg, {'mergelog': ktree.mergelog})
+            with open(ktree.mergelog, 'r') as mergelog:
+                for line in mergelog:
+                    # Skip the useless part of the 'git am' output
+                    if ("The copy of the patch" in line) \
+                            or ('see the failed patch' in line):
+                        break
+                    report_string += '    %s\n' % line.strip()
 
-        report_string += '\n'.join([
-            '\nHowever, the application of the last patch above failed with'
-            ' the',
-            'following output:\n\n'
-        ])
-        with open(ktree.mergelog, 'r') as mergelog:
-            for line in mergelog:
-                # Skip the useless part of the 'git am' output
-                if ("The copy of the patch" in line) \
-                        or ('see the failed patch' in line):
-                    break
-                report_string += '    %s\n' % line.strip()
+            report_string += '\n'.join([
+                '\nPlease note that if there are subsequent patches in the'
+                'series, they weren\'t',
+                'applied because of the error message stated above.\n'
+            ])
+            report_results(merge_result_path, 'false',
+                           merge_report_path, report_string)
 
-        report_string += '\n'.join([
-            '\nPlease note that if there are subsequent patches in the series,'
-            ' they weren\'t',
-            'applied because of the error message stated above.\n'
-        ])
-        report_results(merge_result_path, 'false',
-                       merge_report_path, report_string)
-
-        return
-    except Exception:
-        (exc, exc_type, trace) = sys.exc_info()
-        raise exc, exc_type, trace
+            return
+        except Exception:
+            (exc, exc_type, trace) = sys.exc_info()
+            raise exc, exc_type, trace
 
     kpath = ktree.getpath()
     buildhead = ktree.get_commit_hash()
@@ -662,21 +673,21 @@ def setup_parser():
     parser_merge.add_argument(
         "--patch",
         type=str,
-        action='append',
+        action=AppendMergeArgument,
         help="Path to a local patch to apply "
              + "(use multiple times for multiple patches)"
     )
     parser_merge.add_argument(
         "--pw",
         type=str,
-        action='append',
+        action=AppendMergeArgument,
         help="URL to Patchwork patch to apply "
              + "(use multiple times for multiple patches)"
     )
     parser_merge.add_argument(
         "-m",
         "--merge-ref",
-        action='append',
+        action=AppendMergeArgument,
         help="Merge ref format: 'url [ref]' "
              + "(use multiple times for multiple merge refs)"
     )
@@ -977,19 +988,6 @@ def load_config(args):
         # Use the reporter configuration from the configuration file
         cfg['reporter'] = config.items('reporter')
 
-    if not cfg.get('merge_ref'):
-        cfg['merge_ref'] = []
-    else:
-        for index, to_merge in enumerate(cfg['merge_ref']):
-            cfg['merge_ref'][index] = to_merge.split()
-
-    for section in config.sections():
-        if section.startswith("merge-"):
-            mdesc = [config.get(section, 'url')]
-            if config.has_option(section, 'ref'):
-                mdesc.append(config.get(section, 'ref'))
-            cfg['merge_ref'].append(mdesc)
-
     # Get an absolute path for the work directory
     if cfg.get('workdir'):
         cfg['workdir'] = full_path(cfg.get('workdir'))
@@ -1060,13 +1058,6 @@ def check_args(parser, args):
     if (args._name == 'build' and args.cfgtype == 'rh-configs'
             and not args.rh_configs_glob):
         parser.error("--cfgtype rh-configs requires --rh-configs-glob to set")
-
-    # Don't allow a mix of --merge-ref, --patch and --pw options
-    if args._name == 'merge':
-        if (args.merge_ref and any([args.patch, args.pw])) \
-                or (args.patch and args.pw):
-            parser.error('--merge-ref, --patch and --pw options are mutually '
-                         'exclusive!')
 
     # Check required arguments for 'report'
     if args._name == 'report':
--- a/tests/test_executable.py
+++ b/tests/test_executable.py
@@ -121,6 +121,25 @@ class TestExecutable(unittest.TestCase):
         args = ['report', '--reporter', 'stdio']
         self.check_args_tester(args, expected_fail=False)
 
+    def test_merge_option(self):
+        """Test parsing of multiple different merge options."""
+        args = ['--workdir', '/tmp/workdir', '--state', 'merge',
+                '--patch', 'patch.txt',
+                '--pw', 'http://patchwork.example.com/patch/1',
+                '--merge-ref', 'git://example.com/repo1']
+        parser = executable.setup_parser()
+        args = parser.parse_args(args)
+        cfg = executable.load_config(args)
+        # Check that ordered mixture of merge arguments
+        self.assertEqual('patch', cfg['merge_queue'][0][0])
+        self.assertEqual('pw', cfg['merge_queue'][1][0])
+        self.assertEqual('merge_ref', cfg['merge_queue'][2][0])
+        # Check the content patch of ordered mixture merge arguments
+        self.assertEqual('patch.txt', cfg['merge_queue'][0][1])
+        self.assertEqual('http://patchwork.example.com/patch/1',
+                         cfg['merge_queue'][1][1])
+        self.assertEqual('git://example.com/repo1', cfg['merge_queue'][2][1])
+
     def test_load_config(self):
         """Test load_config() with some arguments."""
         config_file = [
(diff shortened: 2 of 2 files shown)


skt/executable.py
```diff
@@ -194,57 +203,61 @@ def cmd_merge(cfg):
...
-                    ktree.merge_patch_file(patch)
-                    idx += 1
-
-            elif cfg.get('pw'):
-                utypes.append("[patchwork]")
-                for patch in cfg.get('pw'):
-                    save_state(cfg, {'patchwork_%02d' % idx: patch})
-                    report_string += '  - %s,\n' % get_patch_name(
-                        get_patch_mbox(patch)
+        if cfg.get('order_merge'):
```
veruu: It would be easier to read if we skipped the whole try/except block if no options are passed, instead of indenting everything. Especially since we are now iterating over a single attribute.

skt/executable.py
```diff
@@ -44,6 +44,15 @@
 retcode = SKT_SUCCESS
 
 
+class CustomAction(argparse.Action):
```
veruu: I'd use a more descriptive class name, like `AppendMergeParameter` (or something shorter, but the idea is to have a descriptive name)

skt/executable.py
danghai: I do not get it `skip the whole try/except block`. Could you elaborate more? 
`Especially since we are now iterating over a single attribute.` the option `--merge-ref` still iterates the multiple attribute ( It has the for loop inside). Could you give me the example command for option `merge-ref`? Because it is not mentioned in `README`

danghai:
@veruu  Thanks review. I fix the `idx`, and `report string`. Could you review now?

skt/executable.py
danghai: What about the name `AppendMergeArgument` ?

veruu: Sure, that works too

skt/executable.py
veruu: I mean that we are iterating over the `order_merge` attribute (can we name it something else so it doesn't sound like a command?), instead of multiple attributes. So instead of

```
try:
    if cfg.get('order_merge'):
        for task in cfg['order_merge']
            ....
```

we could have

```
for thing_to_merge in cfg.get('order_merge', []):
    try:
        ....
```

Granted, we need to make sure that the option isn't initialized to `None` but `[]` but this should be handled by the custom action.

The `merge-ref` option parsed by command line is slightly broken right now and only grabbing it from config file works (I'll take a look at fixing that later), but the format is available in the merge help. After parsing, the format of the parsed option should be a list of `[url, reference]` sublists (similarly as the `patch` is a list of patch paths).

Does this answer your questions?

skt/executable.py
```diff
@@ -194,52 +202,61 @@ def cmd_merge(cfg):
...
-                    save_state(cfg, {'patchwork_%02d' % idx: patch})
-                    report_string += '  - %s,\n' % get_patch_name(
-                        get_patch_mbox(patch)
-                    )
-                    report_string += '    grabbed from %s\n' % patch
-
-                    ktree.merge_patchwork_patch(patch)
-                    idx += 1
+        if cfg.get('order_merge'):
+            if cfg['order_merge'][0] == 'merge_ref':
```
veruu: Based on your explanation, this seems wrong. Wouldn't `cfg['order_merge'][0]` be `('option name', 'option value')` so the condition would never go through? And what if we first want to apply a patch and then merge the git repo and the `merge_ref` won't be the first one?

skt/executable.py
danghai: Awesome! I know what you mean.

veruu:
Have you tried the format in the merge help? `skt merge <all the options you want> --merge-ref 'repo-url reference'`, where you substitute the `repo-url` and `reference` by the real values (if you want to use master, you don't have to include `reference`). If it doesn't work, can you file the issue describing what happens?

If we had the idea for the same format for the options, we'd already implement it since it would simplify things a lot :) Having the information about what repository and commit was cloned at the beginning of the report, and then mentioning what was merged/patched based on the specific option still works, it just needs a bit tweaking.

danghai:
I think I put the wrong `repo-url`. I put `repo-url`:
```
git://git.kernel.org/pub/scm/linux/kernel/git/davem/net-next.git
```
The error:
```
  File "/home/danghai/skt/skt/executable.py", line 215, in cmd_merge
    (retcode, bhead) = ktree.merge_git_ref(*mb)
  File "/home/danghai/skt/skt/kerneltree.py", line 277, in merge_git_ref
    stderr=subprocess.PIPE)
  File "/home/danghai/skt/skt/kerneltree.py", line 110, in __git_cmd
    **kwargs)
  File "/home/danghai/skt/skt/kerneltree.py", line 93, in __git_cmd_call
    **kwargs
TypeError: check_output() got multiple values for keyword argument 'stderr'
```

veruu:
Can you please open the issue and list the full command you used (there shouldn't be any sensitive information)? Pull requests are not a place for debugging

danghai:
@veruu I have update it. It still needs to enhance the string `report`, as well as `idx`. Could you review now?

veruu:
Functionality-wise this seems to apply the patches just fine (super cool!). I tested it with my clone of the kernel tree and some random local and remote patches (didn't try the merge yet). However, there are some things that need further tweaking:

This is the copy of the text that was created:
```
We cloned the git tree and checked out fedb8da96355 from the repository at
  https://github.com/veruu/linuxWe applied the following patch(es):
  - [PATCH] tools/power turbostat: Do not display an error on systems without a cpufreq driver,
    grabbed from https://patchwork.kernel.org/patch/10560441/on top of commit fedb8da96355 from the repository at
  https://github.com/veruu/linuxWe applied the following patch(es):
  - /home/vkabatova/git/skt/0001-test-MAINTAINERS.patch
on top of commit fedb8da96355 from the repository at
  https://github.com/veruu/linux
```
As you can see, there are some newlines missing and the repository path with "we applied the following... ... on top of..." part is present multiple times.

Another thing is that for baseline testing (without applying any patches; this is used to test current status of the repository), there are no result or report files created. You can check the current master to see what is the expected content of the files.

If you want to apply multiple patches of the same type, they later ones aren't saved in the state file. I tried applying two local patches and while they applied just fine, the state file contained only the first one. While we will eventually get rid of the state file, it is still used now and this functionality needs to work.

I'm really happy to see the functionality working! Can you resolve the details I mentioned here?

danghai:
@veruu I am very glad to get these review.  I fix some issue that you mention above:

> Another thing is that for baseline testing (without applying any patches; this is used to test current status of the repository), there are no result or report files created. You can check the current master to see what is the expected content of the files.

I fix it.

> If you want to apply multiple patches of the same type, they later ones aren't saved in the state file. I tried applying two local patches and while they applied just fine, the state file contained only the first one. While we will eventually get rid of the state file, it is still used now and this functionality needs to work.
I think the reason is `idx` ( the counter patch). It still applies the patch, but the idx reset back to 0, and it overwrite it in `state file`. I create the `idx = [0, 0 , 0] following idx[0]: merge-ref, idx[1]: pw, and idx[2]: patch. Could you recheck it for me?

> As you can see, there are some newlines missing and the repository path with "we applied the following... ... on top of..." part is present multiple times.
I add a new line in it. But it still need to be enhanced because the `merge-ref` does not have the same format with others. 
Could you review it? Thanks

skt/executable.py
```diff
@@ -221,70 +231,69 @@ def cmd_merge(cfg):
...
-        elif cfg.get('patch') or cfg.get('pw'):
-            report_string = 'We applied the following patch(es):'
+            elif thing_to_merge[0] == 'patch' or thing_to_merge[0] == 'pw':
+                report_string += '\nWe applied the following patch(es):'
 
-            if cfg.get('patch'):
-                utypes.append("[local patch]")
-                for patch in cfg.get('patch'):
+                if thing_to_merge[0] == 'patch':
+                    utypes.append("[local patch]")
```
veruu: Putting it here means the type will be added with each patch applied. So if we grab series with 50 patches, the type will be present in the state file 50 times, which is not exactly wanted. Since you went with the array of indices, it may be better to iterate over it at the end and add the type for each non-zero one, what do you think?

veruu:
Also, the duplicate report is still an issue. Here is the output of two local patch applications:
```
We cloned the git tree and checked out fedb8da96355 from the repository at
  https://github.com/veruu/linux
We applied the following patch(es):
  - /home/vkabatova/git/skt/mbox
on top of commit fedb8da96355 from the repository at
  https://github.com/veruu/linux
We applied the following patch(es):
  - /home/vkabatova/git/skt/0001-test-MAINTAINERS.patch
on top of commit fedb8da96355 from the repository at
  https://github.com/veruu/linux
```

Again, I have not tested the merges yet.

skt/executable.py
```diff
@@ -180,7 +189,7 @@ def cmd_merge(cfg):
     """
     global retcode
     utypes = []
-    idx = 0
+    idx = [0, 0, 0]
```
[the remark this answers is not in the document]
danghai: Yeah I will note it in commit message when I am done.

skt/executable.py
danghai: @veruu yeah it is right. I can add it at the end. Or I can add it at beginning (outside the loop). I think the way to fix the string at the beginning and, add the type at the same time.

skt/executable.py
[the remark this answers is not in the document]
veruu: Do note it in a comment as well because it's not an obvious piece of code, so people don't have to dig through the logs to figure out what it does.

danghai:
@veruu I resolve all the above issues. As I said above, I add the `utypes` at beginning (outside the loop),  I resolve  the report string at the same time. For report string, my algorithm is:
Outside the loop, I create the template report string depending on condition of merge (`merge-ref`, or `patch` , `pw` is the same format as `patch`). Inside the loop, for each iteration, I fill a `merge patch` or `merge ref` base on searching keyword. In particular, for `merge-ref` I look the index of keyword`tree:`, and for `pw` and `patch` I look the index of keyword `patch/es:`. And then I make the string concatenation. The information of `merge-ref` is always above `pw` or `patch`. The order of patch in `pw` or `patch` from high to low is `high is recent merge `. 
I am not sure it is a good way or not. If you have another algorithm, please let me know. 

I pass the test case:
1. Every single option merge
2. Double option merge:  (I only test ref = master)
<pw> <pw>
<patch> <patch>
<pw> <patch>
<patch> <pw>
<pw> <merge-ref>
<merge-ref> <pw> 
3) Triple option merge with all random order.
<pw> <patch> <merge-ref>
<pw> <merge-ref> <patch>
...
The report string looks fine with me. 

The Travis fail because I actually modify the argument load `merge-ref`. We cannot take the value of `cfg[merge-ref]` = `repo` + `ref`. I do this in loop of `cmd_merge` and remove this split.

veruu:
```
We cloned the git tree and checked out fedb8da96355 from the repository at
  https://github.com/veruu/linux
We applied the following patch(es):
  - /home/vkabatova/git/skt/0001-test-MAINTAINERS.patch
  - /home/vkabatova/git/skt/mbox
on top of commit fedb8da96355 from the repository at
  https://github.com/veruu/linux
```
The last two lines seem redundant since they are already at the beginning.

For the option order of `<local patch>, <local patch>, <merge>`, the report is the following:
```
We cloned the git tree and checked out fedb8da96355 from the repository at
  https://github.com/veruu/linux
We merged the following references into the tree:
  - https://github.com/veruu/linux
    into commit 95be1d332c29
We applied the following patch(es):
  - /home/vkabatova/git/skt/0001-test-MAINTAINERS.patch
  - /home/vkabatova/git/skt/mbox
on top of commit fedb8da96355 from the repository at
  https://github.com/veruu/linux
```
Which is fairly confusing because the report order doesn't match the options applied. If you are the receiver of the report, what would you think was done? And again, the first two and last two lines contain the same information.

The report string searching and overwriting is a really bad idea. Repeating what I said in the beginning,
>  Having the information about what repository and commit was cloned at the beginning of the report, and then mentioning what was merged/patched based on the specific option

The only thing that needs to be done for this to work is to slightly tweak the strings - the text itself that's created now would be fine if it kept the right order and didn't repeat itself.

> The Travis is fail in tests because I actually modify the option merge-ref. We cannot take the value of cfg[merge_ref] = repo + ref. I do this in loop of cmd_merge and remove this split at below.

If the tests start failing because the functionality is changed in the patch you submit, you have to fix the tests to reflect the change.

danghai:
@veruu Thanks review. I am not sure about what do you mean `tweak the strings `? Could you elaborate it more?
However, I update the PR following enhance something:
1. Removing the search string and then concatenation. I use `merge_string` ( it stores string of `merge-ref`), and `patch_string` ( it stores string of `pw` and `patch`). Fill information for each iteration, and then concatenation at the end. 
2. I fix the tests issue for `merge-ref`. 

Could you review? Thanks

veruu:
@danghai can you rebase the pull first?

danghai:
@veruu oh you remove the `uid` and `utype` at previous commits. Let me read your patches first.

danghai:
@veruu I rebase the pull

veruu:
Thanks for the changes! The duplicated report lines are now solved, but the order of the report lines still seems wrong.

Running `skt -vv --rc skt-rc --state --workdir ../linux merge -b https://github.com/veruu/linux --patch mbox --patch 0001-test-MAINTAINERS.patch --merge-ref 'https://github.com/veruu/linux testbranch'` (which is my testing run with two local patches and a merge) results in
```
We cloned the git tree and checked out 10f3e23f07cb from the repository at
  https://github.com/veruu/linux
We merged the following references into the tree:
  - https://github.com/veruu/linux
    into commit 95be1d332c29
We applied the following patch(es):
  - /home/vkabatova/git/skt/mbox
  - /home/vkabatova/git/skt/0001-test-MAINTAINERS.patch
```

In this case, the merge should be reported last, since it was applied last.

danghai:
@veruu thanks review. I resolve the order of `merge-ref` and `pw` or `patch`. In my opinion, I think it is not important. For example, if I want to merge following order:
```
<pw>    <merge-ref>   <patch>   <merge-ref>
```
The output probably is:
```
We applied the following patch(es):
<pw>
<patch>
We merged the following references into the tree:
<merge-ref>  (1)
<merge-ref>  (2)
```
It is going to fill information in 2 types. We still do not know the order between `merge-ref` and `pw` , `patch`. However the merge process is order. What do you think?

veruu:
No, in that case the output should be
```
We applied the following patch(es):
<pw>
We merged the following references into the tree:
<merge-ref>  (1)
We applied the following patch(es):
<patch>
We merged the following references into the tree:
<merge-ref>  (2)
```
The order of the output should reflect the order of arguments. The output you showed would imply `<pw>, <patch> <merge-ref>, <merge-ref>` The output is not important to the caller since they know what command they used. It's important for the developers who don't see the command, only the report that is sent to them, and it should be accurate.

danghai:
So, I think I miss understand what you want me to do at beginnning. For clear, what the suppose output of this order:
```
<pw>   <patch>    <merge-ref>   <merge-ref>
```

veruu:
Yes, in case of the `<pw>   <patch>    <merge-ref>   <merge-ref>` arguments, the output you show is correct.

danghai:
@veruu I have updated it, but the logic changes a little bit. I use the `previous_merge` to keep the information about which option merge is. From this, I can handle the `report_string` better.

veruu:
Seems like the reports are missing the intro line about applied patches now:

```
We cloned the git tree and checked out 10f3e23f07cb from the repository at
  https://github.com/veruu/linux
  - /home/vkabatova/git/skt/0001-test-MAINTAINERS.patch        <---- missing intro about patch being applied
We merged the following references into the tree:
  - https://github.com/veruu/linux
    into commit 95be1d332c29
We applied the following patch(es):
  - /home/vkabatova/git/skt/mbox
```

```
We cloned the git tree and checked out 10f3e23f07cb from the repository at
  https://github.com/veruu/linux
  - /home/vkabatova/git/skt/mbox    <------ again, what's being done with the patches?
  - /home/vkabatova/git/skt/0001-test-MAINTAINERS.patch
We merged the following references into the tree:
  - https://github.com/veruu/linux
    into commit 95be1d332c29
```

danghai:
@veruu Oh sorry, I miss one condition in `if`. I fix it. Could you review now? Thanks

danghai:
Hi @veruu what about this patch?

skt/executable.py
```diff
@@ -203,15 +218,15 @@ def cmd_merge(cfg):
...
-                report_string,
-                'We merged the following references into the tree:'
-            )
-            for mb in cfg.get('merge_ref'):
-                save_state(cfg, {'mergerepo_%02d' % idx: mb[0],
-                                 'mergehead_%02d' % idx: bhead})
+    for thing_to_merge in cfg.get('merge_queue', []):
+        try:
+            if thing_to_merge[0] == 'merge_ref':
+                if not previous_merge[0] == 'merge_ref':
```
veruu: Can we use a less confusing syntax of `if previous_merge[0] != 'merge_ref':` ?

skt/executable.py
```diff
@@ -180,7 +190,12 @@ def cmd_merge(cfg):
         cfg:    A dictionary of skt configuration.
     """
     global retcode
-    idx = 0
+    # Counter merge patch following:
+    # id[0]:  counter of merge-ref option.
+    # idx[1]: counter of patch option.
+    # idx[2]: counter of pw option.
+    idx = [0, 0, 0]
+    previous_merge = ['', '']
```
veruu: Why use a list of two strings here when only the first one (the name of the option) is used in the comparisons and the second one is never touched? Wouldn't it be easier to use a single string, to which the name of the applied option is assigned instead?

skt/executable.py
```diff
@@ -220,68 +235,66 @@ def cmd_merge(cfg):
...
-
-        elif cfg.get('patch') or cfg.get('pw'):
-            report_string = 'We applied the following patch(es):'
-
-            if cfg.get('patch'):
-                for patch in cfg.get('patch'):
+                previous_merge = thing_to_merge
+                idx[0] += 1
+
+            elif thing_to_merge[0] == 'patch' or thing_to_merge[0] == 'pw':
```
veruu: There's no other option that can be applied besides `merge_ref`, `patch` and `pw`. If should be enough to use a siple `else` instead of the condition here.

skt/executable.py
```diff
@@ -220,68 +235,66 @@ def cmd_merge(cfg):
...
+                if previous_merge[0] == 'merge_ref' or not previous_merge[0]:
```
veruu: Similar reason as above (and readability), it would be more clear to see what's going on with a modified condition of `if previous_merge[0] not in ['patch', 'pw']`, what do you think? Without that index if we go with the suggestion above about the `previous_merge` variable

skt/executable.py
```diff
@@ -220,68 +235,66 @@ def cmd_merge(cfg):
...
+                    report_string += '\nWe applied the following patch(es):'
+                if thing_to_merge[0] == 'patch':
+                    patch = thing_to_merge[1]
```
veruu: This can be joined into a single line with the line below, using `thing_to_merge[1]` as the argument

veruu:
Thanks for the patience! I had some higher-priority problems to resolve and didn't want to rush the review in between so I don't miss anything.

The functionality works well now, at least based on my testing. However, I noticed a residue of the previously used options in the configuration. Can those be removed, since there is no use for them now? Those options used to be lists and now are only a string that gets overwritten with each use of the same option so even if we were to use them, the data wouldn't be correct.

I left some more comments to clean up the code inline

skt/executable.py
danghai: I fix it.

skt/executable.py
danghai: Right. It is redundant

skt/executable.py
danghai: yeah it is better

skt/executable.py
danghai: I fix it

skt/executable.py
danghai: I fix it

danghai:
@veruu thanks. I rebase and update the patch based on your review. I also put some information in commit message. Let me know if I need to change anything else.

Take your time, I just ping you because just in case you do not receive the notification.

veruu:
I mean the thing you eg. changed in the tests - these old options are a) not a list of strings anymore and b) not used because we are iterating over `merge_queue` so it would be better to just not set them anymore and modify the tests that used them to check `merge_queue`. Is this explanation more clear?

skt/executable.py
```diff
@@ -220,68 +234,65 @@ def cmd_merge(cfg):
...
 
                     return
-
                 report_string += '\n  - %s\n    into commit %s' % (mb[0],
                                                                    bhead[:12])
-                idx += 1
-
-        elif cfg.get('patch') or cfg.get('pw'):
-            report_string = 'We applied the following patch(es):'
+                previous_merge = thing_to_merge
```
veruu: This should use `thing_to_merge[0]` too

danghai:
@veruu I understand what you mean. But I do not know how to check `merge_queue`  in `tests`. For clear, I remove something in next commit. Do you want to remove these configuration, and replace them by check `merge_queue` in test? Could you give me the hint to write the this test case?

veruu:
> Do you want to remove these configuration, and replace them by check merge_queue in test?

Yes. You don't write a new test case, but replace the `pw`, `patch` and `merge_ref` checks in tests by the `merge_queue` (with the correct index and merge option)

danghai:
@veruu Sorry I am not quite sure how to get value of `merge_queue` in tests :( . Do I need to set anything in `load_config` relating `merge_queue` because it is a way to get value in tests? Thank you

veruu:
@danghai `load_config` is for reading configuration files, not command line arguments. Use the parser to parse arguments, as it is done in the test helpers `check_args_tester` and `load_config_tester`.

danghai:
@veruu Thanks give me the hint. I follow the wrong direction, so it takes a lot of time. I add the the test for ordered mixture of merge arguments.

veruu:
Thanks! Can you merge the commits into a single one, since they are all related to a single change? I don't think the testing should be done via a config tester but the arg parser, as I mentioned before - since the options are not supposed to be read from the configuration it's confusing. So I'd just use the parser and rename the method accordingly (eg. `test_merge_options`)

danghai:
@veruu I squash all commits into 1 commit. I remove the confused `config` inside, and change the name too.

danghai:
Oh it still has 2 more commit. Sorry, let me remove it

tests/test_executable.py
```diff
@@ -121,6 +121,25 @@ def test_check_args_stdio_valid(self):
         args = ['report', '--reporter', 'stdio']
         self.check_args_tester(args, expected_fail=False)
 
+    def test_merge_option(self):
+        """Test test_merge_option() with some arguments."""
```
veruu: The docstring should say what is being tested, not the name of the test, so in this case it should say something like `Test parsing of multiple different merge options`

danghai:
@veruu I fix it. I am glad to see this patch is merged. Thanks review and give me the advice,

PR #311, opened 2018-08-03
Don't raise exception if can't publish tarball

veruu:
Allow publisher to be ran anytime, even if the build fails. It makes it
easier to publish what is available instead of raising an exception,
especially since the build command returns non-zero code if it failed so
the caller knows whether the build succeeded or not.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -393,12 +393,12 @@ def cmd_publish(cfg):
         cfgurl = publisher.publish(cfg.get('buildconf'))
         save_state(cfg, {'cfgurl': cfgurl})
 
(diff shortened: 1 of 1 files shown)


danghai:
I think it should be updated the example of command in README too

veruu:
@danghai where do you see it? I can't find an example with `--tarpkg` in the README

danghai:
I mean this command:
```
    publish -p cp /srv/builds http://skt-server
```

veruu:
That command doesn't have to be changed since the information about both config and kernel is grabbed from the statefile

danghai:
Ah I see. :)

PR #314, opened 2018-08-05
 Executable.py: Substitute removing the existing results by function

danghai:
I make PR to substitute remove the existing results by function.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -67,6 +67,22 @@ def report_results(result_path, result_string, report_path, report_string):
         report_file.write(report_string)
 
 
+def remove_oldresult(output_dir, prefix_filename):
+    """
+    Remove existing results from previous runs.
+
+    Args:
+        output_dir:      Source directory stores existing result.
+        prefix_filename: Prefix of the existing result filename.
+    """
+    try:
+        for filename in os.listdir(output_dir):
+            if filename.startswith(prefix_filename):
+                os.unlink(join_with_slash(output_dir, filename))
+    except OSError:
+        pass
+
+
 def save_state(cfg, state):
     """
     Merge state to cfg, and then save cfg.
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -59,12 +59,19 @@ def report_results(result_path, result_string, report_path, report_string):
         report_path:    A full path of report log.
         report_string:  The content of string report.
     """
-    with os.fdopen(os.open(result_path, os.O_CREAT | os.O_WRONLY),
-                   'w') as result_file:
-        result_file.write(result_string)
-    with os.fdopen(os.open(report_path, os.O_CREAT | os.O_WRONLY),
-                   'w') as report_file:
-        report_file.write(report_string)
+    try:
```
veruu: What's the idea behind the try/except block? If the result files can't be created, it a problem in user's configuration and we should hide it.

danghai: @veruu I am little bit confused. Could you ask you about idea behind the try/except block here? https://github.com/RH-FMK/skt/blob/master/skt/kerneltree.py#L54. Thanks

danghai: Sorry I make you confused. Could you elaborate `user's configuration` ? I point out the link of the code because I think try/except for 2 situations are the same. I think the result file cannot be created being the same the the workdir cannot create.

veruu: The workdir can't be created if the user created it manually before running skt. We are always removing any result files so not being able to create a result file because it already exists would never happen.

danghai: Ah I see. the user has to resolve and it cannot pass in exception. But the workdir is different because it could probably pass if user created it manually before. 
What about the next commit?

veruu: Functionality-wise, that one is OK. I'd like to see the "source_dir" argument renamed to "output_dir" or something similar (to not be confused with kernel source directory), and rename the function itself ("exresult" is not that clear - spelling out the full word or using "old" would work better, for example)

danghai:
@veruu ah I have the idea about return value. I think it could return the path for both report, and result. What do you think? I reduce a little bit code in cmd_merge.

veruu:
Uhm... why would a return value be needed?

danghai:
I mean I can put these in function too. For example:
```
    merge_result_path = join_with_slash(cfg.get('output_dir'),
                                        'merge.result')
    merge_report_path = join_with_slash(cfg.get('output_dir'),
                                        'merge.report')
```

It returns the both path

veruu:
If I understand correctly what you want to do, I don't think that would improve the code.

danghai:
It just reduces the code in cmd_merge. It dot not improve the code. I mean the function remove the old value and return 2 path. 
```
merge_result_path, merge_report_path = write_oldresult(...)
```

veruu:
Returning a totally unrelated values doesn't look like a good idea. A single more line of code is better than a code noone understands, and we are actively trying to get rid of the code that is confusing.

danghai:
@veruu I fixed it. Thanks

Issue #296, opened 2018-07-26
Add support for input and output directories for results


veruu:
See meta/6 for full discussion. I thought we can get by with using the workdir in the meanwhile (and that should stay the default), but we have some cute setups with runner not having access to the real workdir.

veruu:
Closing this one since we are going with a different design decisions

PR #321, opened 2018-08-09
Add support output directory for result

danghai:
I make PR to add support output directory for result. If the command has `--output_dir` or `-o` option, it is going to create the output directory to store result. And the default is working directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,18 +131,18 @@ written to the `<JUNIT_DIR>` directory.
 
 To checkout a kernel tree run:
 
-    skt --rc <SKTRC> --state --workdir <WORKDIR> -vv \
+    skt --rc <SKTRC> --state --workdir <WORKDIR> --output-dir <OUTDIR> -vv \
         merge --baserepo <REPO_URL> --ref <REPO_REF>
 
 Here `<SKTRC>` would be the configuration file to retrieve the configuration
 and the state from, and store the updated state in. `<WORKDIR>` would be the
-directory to clone and checkout the kernel repo to, `<REPO_URL>` would be the
-source kernel Git repo URL, and `<REPO_REF>` would be the reference to
-checkout.
+directory to clone and checkout the kernel repo to, `<OUTDIR>` would be
+directory to store output files, `<REPO_URL>` would be the source
+kernel Git repo URL, and `<REPO_REF>` would be the reference to checkout.
 
 E.g. to checkout "master" branch of the "net-next" repo:
 
-    skt --rc skt-rc --state --workdir skt-workdir -vv \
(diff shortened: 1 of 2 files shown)


skt/executable.py
```diff
@@ -564,6 +564,12 @@ def setup_parser():
         type=str,
         help="Path to work dir"
     )
+    parser.add_argument(
+        "-o",
+        "--output_dir",
```
veruu: The command line options should use `-` instead of `_`

skt/executable.py
```diff
@@ -564,6 +564,12 @@ def setup_parser():
...
+        type=str,
+        help="Path to output dir"
```
veruu: Can you use the full "directory" word here?

skt/executable.py
```diff
@@ -966,8 +972,13 @@ def load_config(args):
             cfg['runner'][1]['blacklist']
         )
 
-    # Preparation for output directory option
-    if os.access(cfg.get('workdir'), os.W_OK | os.X_OK):
+    # Create and get an absolute path for the output directory
+    if cfg.get('output_dir'):
+        try:
+            os.mkdir(full_path(cfg.get('output_dir')))
```
veruu: Can you assign the full path to the output directory first and work with that? That way we don't need to convert it with every use.

danghai:
@veruu thanks review. I have updated it.  Could you review it, and my commit message? It relates to issue #296 But I am not sure about what `input` for directory result is? Thanks

README.md
```diff
@@ -131,18 +131,18 @@ written to the `<JUNIT_DIR>` directory.
...
+    skt --rc <SKTRC> --state --workdir <WORKDIR> --output-dir <OUTDIR> -vv \
         merge --baserepo <REPO_URL> --ref <REPO_REF>
 
 Here `<SKTRC>` would be the configuration file to retrieve the configuration
 and the state from, and store the updated state in. `<WORKDIR>` would be the
-directory to clone and checkout the kernel repo to, `<REPO_URL>` would be the
-source kernel Git repo URL, and `<REPO_REF>` would be the reference to
-checkout.
+directory to clone and checkout the kernel repo to, `<OUTDIR>` would be
+directory to store output files creation, `<REPO_URL>` would be the source
```
veruu: the `creation` word doesn't belong here

veruu:
@danghai don't worry about mentioning the issue. It's a bit more complicated and you'd have to read the full discussion in the meta/pull6. We'd like to make skt more modular and get rid of workdir and potentially state as well.

danghai:
@veruu okay. I am going to read it.

danghai:
@veruu do I need to fix anything in this patch?

veruu:
No, I think the pull is fine now that it's rebased. Will merge it soon, thanks!

PR #323, opened 2018-08-12
Executable.py: Add removing output directory in cmd_cleanup()

danghai:
I make PR to remove the whole output directory, if the global `--wipe` option was specified.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -492,8 +492,11 @@ def cmd_cleanup(cfg):
         except OSError:
             pass
 
-    if cfg.get('wipe') and cfg.get('workdir'):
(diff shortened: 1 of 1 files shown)


veruu:
The output directory can be different for each command, and can contain anything besides the result files created. If you want the result files dropped into your home directory, this would then delete it, despite the fact that you probably only wanted to delete skt-related files.

I wouldn't worry about removal of output directories.

danghai:
Oh I thought we have the `output_dir`, and then we put all output result of all stages (merge, build, run,...) in it. `cleanup` function will clean it if it is specified.

veruu:
I believe this pull should be closed now, please reopen if anything changes

PR #325, opened 2018-08-13
add console checking stage

veruu:


skt/executable.py
```diff
@@ -460,6 +461,50 @@ def cmd_report(cfg):
...
+        raise Exception('<krelease> or <console-url> parameter missing!')
+
+    console_logs = []
+    trace_list = []
+
+    for console_path_or_url in cfg.get('console'):
+        console_log = skt.console.ConsoleLog(cfg.get('krelease'),
+                                             console_path_or_url)
+
+        console_logs.append(console_log)
```
spbnick: Why do we keep those around?

skt/executable.py
```diff
@@ -460,6 +461,50 @@ def cmd_report(cfg):
...
+        trace_list.append(console_log.gettraces())
+
+    if any(trace_list):
+        report_string = ''
+
+        for traces in trace_list:
```
spbnick: Shouldn't this be called `trace`, not `traces`? We already have `trace_list`, and its name suggest that each of its elements is a trace.

skt/executable.py
```diff
@@ -460,6 +461,50 @@ def cmd_report(cfg):
...
+            if traces:
```
spbnick: Why do we check this?

skt/executable.py
```diff
@@ -460,6 +461,50 @@ def cmd_report(cfg):
...
+                report_string += '{}\n{}:\n\n{}\n\n'.format(
+                    'This is the first trace we found in ',
+                    cfg.get('console')[trace_list.index(traces)],
+                    traces[0]
```
spbnick: Sorry, I'm lost, what are we outputting here?

skt/console.py
```diff
@@ -149,14 +150,22 @@ def __fetchdata(self):
...
             List of console log lines related to tested kernel
         """
-        if not self.url:
+        if not self.url_or_path:
             return []
 
-        response = requests.get(self.url)
+        try:
+            console_text = requests.get(self.url_or_path).text
+        except Exception:  # We got a file path
```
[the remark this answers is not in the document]
veruu: good idea!

skt/executable.py
veruu: I wanted to be able to attach the parsed out log in this stage if someone was to run the stage separately, outside of skt. Since we don't have the input/output directory schema fully working yet, that part is not done. If you think that's a bad idea, I can remove this preparation.

skt/executable.py
veruu: `traces` is a list of traces and `trace_list` is a list of them. If you have better names for this kind of nesting please do share it (this was the best I could come up with) but I feel that calling it `trace` would be misleading in this case

skt/executable.py
veruu: Because the method for extracting the traces returns an empty list if there are no traces found, and we don't want to say "we found an empty trace"

skt/executable.py
veruu: The path/URL of the console log in which the trace was found and the related first trace

skt/executable.py
spbnick: I think it is better not to add unused code, in general. Every time somebody reads this they would wonder what's going on, until the code using this is added. It's not a big deal here, so leave it if you feel you'll add something to use it soon, but in general I think it's best to avoid it.

skt/executable.py
spbnick: I see. A list of lists is a bit of an awkward situation for any human language, I expect, but it's still better to be clear. Please see if you could use something like `trace_list` and `trace_list_list`, or `traces` and `traces_list`, or something of that sort. Personally I would use the former.

skt/executable.py
spbnick: I see, thank you.

This is one of the situations, I think, where complex logical checks Python calls for are making for a difficult read. I think an explicit check for list emptiness/length would have been better, but once you make it clear that you're operating on a list (e.g. by renaming `traces` to `trace_list` or similar) it will be easier to understand.

skt/executable.py
spbnick: I see, thank you. Do you think you can explain how you get the path/URL in a comment above this line:

    cfg.get('console')[trace_list.index(traces)],

spbnick:
Thank you, Veronika! I have just two more style requests left :)

veruu:
@spbnick resolved

spbnick:
Looks good, thank you :)

PR #326, opened 2018-08-14
Runner improvements

veruu:
~~- [ ]  depends on pull #325~~ rebased on master


skt/runner.py
```diff
@@ -582,7 +582,7 @@ def __jobsubmit(self, xml):
 
         return jobid
 
-    def run(self, url, max_aborted, release, wait=False, host=None,
```
spbnick: Could you please remove the `test_run_host` test as well?

veruu: Didn't notice that one, thanks for finding it!

PR #328, opened 2018-08-16
Creates a file in place of non-existent directory 

### LINKED ISSUE #327: `skt publish -p cp` creates a file in place of non-existent directory
If the destination directory supplied to `skt publish -p cp` doesn't exist, skt creates a file with the supplied name. Instead, it should abort with an error about non-existent directory. The easiest way to do that would probably be appending a slash (`/`) to the end of the destination path, when invoking `shutil.copy()`.

danghai:
I make PR to create a file in place of non-existent directory.
Fix #327


danghai:
@spbnick Thanks review. I reword the commit message.

PR #330, opened 2018-08-19
Enhance merge patch by applying the wildcard, or a series of patches

danghai:
- [ ] #308 merge first

This patch allows user to merge a series of patches, and to use the command with wildcard to merge. Because developers are more likely to merge a series of patches, and the name of each patch is very
long, merging a series of patches is convenient for this purpose. By using `nargs="*"`, the patches are stored in a list under multiple optional arguments. In the class `AppendMergeArgument`, each `patch_file` in values is going to add to the `namespace` by order. At the previously old patch only is applied under each optional arguments one time.

For example:
```
    --patch PATCH1.patch --patch PATCH2.patch --patch PATCH3.patch
```
Now, it is enhanced by applied a series of patch:
```
    --patch PATCH1.patch PATCH2.patch PATCH2.patch
```
Or it probably is merged by using wildcard command:
```
    --patch *.patch
```
What do you think?

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -44,6 +44,16 @@ LOGGER = logging.getLogger()
 retcode = SKT_SUCCESS
 
 
+class AppendMergeArgument(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        if 'merge_queue' not in namespace:
+            setattr(namespace, 'merge_queue', [])
+        previous = namespace.merge_queue
+        previous.append((self.dest, values))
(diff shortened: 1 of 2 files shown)


veruu:
This patch removes all the work done to issue #227 which @spbnick advocated for. skt is meant to be automated and running it by hand is not a typical use case so I don't think it's worth to undo those changes.

danghai:
@veruu I see. Thanks. I close PR now. Feel free to open it when you think it is necessary.

PR #331, opened 2018-08-22
runner rework - final huge pull

veruu:


skt/runner.py
```diff
@@ -145,14 +145,14 @@ def getresultstree(self, jobid, logs=False):
 
     def __forget_cid(self, cid):
         """
-        Remove a job or recipe from the current list
+        Remove a job or recipe set from self.job_to_recipe_map.
```
spbnick: I would prefer a human-readable name of what's in `self.job_to_recipe_map` used here, and also as a comment over it in `__init__()`. E.g. something like "the job->recipe IDs map". I won't hold my ACK over this, though.

skt/runner.py
```diff
@@ -121,23 +121,23 @@ def __getxml(self, replacements):
...
 
-    def getresultstree(self, jobid, logs=False):
+    def getresultstree(self, cid, logs=False):
         """
-        Retrieve Beaker job results in Beaker's native XML format.
+        Retrieve Beaker results for cid in Beaker's native XML format.
 
         Args:
-            jobid: The Beaker job ID.
+            cid:   ID of the job, recipe or recipe set.
```
spbnick: I believe the Beaker's official term for this is "taskspec", as described in https://beaker-project.org/docs/man/bkr.html#specifying-tasks
I might be wrong (e.g. that's just a placeholder used in that manpage only), but if it is, then we should use it here, and I was already using it through the code.
What do you think?

skt/executable.py
```diff
@@ -421,10 +421,15 @@ def cmd_run(cfg):
                                         cfg.get('wait'),
                                         arch=cfg.get("kernel_arch"))
 
-    for index, job in enumerate(runner.job_to_recipe_map.keys()):
+    recipe_set_index = 0
```
spbnick: I get that these changes are hard to split, I know how that happens. I'm willing to accept it as is, to speed you up, but I would not be able to say I did *any* meaningful review of this, because I just don't have the energy nor time to understand it as is. Perhaps @major would be able to do it :)

skt/runner.py
veruu: TBH I don't know what the official term is and I chose one I saw used in the code already. I'm totally for changing all the `cid`s to `taskspec`s (probably in a separate pull) if that's the right term to use.

major:
I don't see anything within this patchset that worries me a lot. Like @spbnick, this is a big one to thoroughly review. I'll test this out in UpShift and see if I find any issues.

PR #334, opened 2018-08-26
Reporter.py: Add try except in email report

danghai:
I make PR to add try except in email report.


veruu:
I don't think catching the exception is a good idea at all. If something is wrong with sending the email, we should know about it and get the correct error message, not have a "something went wrong" logged without any indication of what happened.

PR #335, opened 2018-08-27
Improve email footer

major:


skt/reporter.py
```diff
@@ -382,11 +382,18 @@ def _getreport(self):
...
+        msg += """
+We welcome any feedback you have about these test results. Please feel free to
+reply to this email with:
+
+    * Questions about the testing process or test results
+    * Suggestions for improvements
+    * Any other comments
+
+Sincerely,
+Kernel CI Team
```
spbnick: I like how you're giving examples of what to write us with. However, we need to let people know they can stop receiving those messages. We must also not be condescending, or appear as an authority.

The "we welcome any feedback", and "please feel free" can be interpreted like we're giving a permission for people to write to us, and that we may even consider answering or taking their messages into account. Instead we should just let people know they can contact us at this address, or even *ask* to contact us. Lastly, we should be sparing with our writing. Our reports already look like a novel (and we should fix that), let's try not to make it longer.

Could you just add the examples to the existing sentence/paragraph, using as few words as possible?

Also, could you keep the indentation intact and just concatenate the strings, instead of using the triple-quoted string starting at column zero?

PR #337, opened 2018-08-29
input/output directory support for merge/build

veruu:


danghai:
@veruu I think `input dir` should be mentioned in `README.md` too?

veruu:
@danghai this is not a full feature, just a preparation for the modularity (runner changes, getting rid of the state and reporting integration in sktm are still pending). I'd prefer to switch up the docs from workdir to input/output directories only after the full functionality is in place because a partial one isn't really usable in practice.

skt/executable.py
```diff
@@ -77,20 +77,21 @@ def report_results(result_path, result_string, report_path, report_string):
...
-    try:
-        for filename in os.listdir(output_dir):
-            if filename.startswith(prefix_filename):
+    for filename in os.listdir(output_dir):
+        if filename.startswith(prefix):
+            try:
                 os.unlink(join_with_slash(output_dir, filename))
-    except OSError:
-        pass
+            except OSError:
```
spbnick: It will probably be OK, but it might be better to check for `os.errno.EISDIR` and re-raise everything else. What do you think?

skt/executable.py
```diff
@@ -298,6 +298,10 @@ def cmd_merge(cfg):
     save_state(cfg, {'workdir': cfg.get('input_dir'),
                      'buildhead': buildhead})
 
+    if cfg.get('input_dir') != cfg.get('output_dir'):
+        shutil.copytree(cfg.get('input_dir'),
+                        join_with_slash(cfg.get('ouptut_dir'), 'merge.source'))
```
spbnick: Typo: `ouptut_dir`.

skt/executable.py
veruu: Works for me!

skt/executable.py
veruu: Aaarggh! Fixing!

veruu:
@spbnick both fixed!

spbnick:
Sorry, I failed at doing this today, will do first thing in the morning.

veruu:
>  I'd like to say that we should never change anything in the input directory, and all intermediate files and outputs should be created in the output directory. If they're the same, it's the caller's decision, but we should only operate on the output directory.

I went with the "do whatever in the input directory with the files you got and only output what is expected by the next command" idea. Never thought about doing it the way you suggested but it does make sense, I'll take a look on implementing it.

spbnick:
Thank you, Veronika! Nag me on the IRC, if anything.

spbnick:
Having temporary files somewhere else would also be fine, we should just keep the input directory intact. If only because multiple invocations could use it, and possibly in parallel.

veruu:
Agreed with the points you make :+1:

veruu:
@spbnick reimplemented! I'm not running into the issue you show, so it probably got resolved in the process.

spbnick:
OK, first issue, re-running a failed "build" command produces this:

```
    load_entry_point('skt==1', 'console_scripts', 'skt')()
  File "/home/nkondras/.local/lib/python2.7/site-packages/skt/executable.py", line 1120, in main
    args.func(cfg)
  File "/home/nkondras/.local/lib/python2.7/site-packages/skt/executable.py", line 182, in wrapper
    func(cfg)
  File "/home/nkondras/.local/lib/python2.7/site-packages/skt/executable.py", line 331, in cmd_build
    join_with_slash(cfg.get('output_dir'), 'merge.source'))
  File "/usr/lib/python2.7/shutil.py", line 191, in copytree
    os.makedirs(dst)
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/home/nkondras/tmp/skt_test/build/merge.source'
```

spbnick:
I would say the copied `merge.source` shouldn't be called `merge.source`, but `build.source` and probably copied unconditionally, since we're likely to change it due to the kernel build process.

spbnick:
I think I also saw `None.csv.config` created in the `build` input directory, not in the output.

veruu:
@spbnick I resolved the issues. The config presence in the input directory looks like a residue of the first test run which might have done that, can you check with the clean directory structure? I can't reproduce it with the current version.

spbnick:
Thank you, Veronika. I tried it out again, and I got this in the build output dir:
```
-rw-r--r-- 1 nkondras nkondras 78648774 Sep 11 20:00 20180911193703-linux-4.19.0-rc2.skt+-x86.tar.gz
-rw-r--r-- 1 nkondras nkondras   998419 Sep 11 20:00 build.log
-rwxr-xr-x 1 nkondras nkondras       72 Sep 11 20:00 build.report
-rwxr-xr-x 1 nkondras nkondras        4 Sep 11 20:00 build.result
-rw-r--r-- 1 nkondras nkondras   205906 Sep 11 20:00 None.csv.config
```

The None.csv.config doesn't seem to be in our design, nor does it comply with the file naming scheme. I think the tarball should be called "build.kernel.tar.gz" according to the design too.
So we need to either change the code or the design. What do you think?

veruu:
The names are probably the defaults if the state file is not used, but I agree that they should be changed. Will do it tomorrow and let you know!

veruu:
@spbnick I renamed the tarball and the kernel config name. Let me know if this works for you!

veruu:
> Thank you, @veruu. This looks better, but build.kernel.tar.gz got created outside the -o directory of the build stage.

Oh. Not sure how that got there, but it's fixed now.

> Also the .report files get created with executable bits for some reason, which they shouldn't.

That's because of the default mask of the file creation used, and unrelated to the code changed in this pull. I'll submit an issue to track it.

spbnick:
Thank you, Veronika, this looks better. However, I get this when running this command in the merge output directory:

```
nkondras@bard:~/tmp/skt_test/merge$ skt --state -vv build -c ../config-4.17.0-3-amd64 
    load_entry_point('skt==1', 'console_scripts', 'skt')()
  File "/home/nkondras/.local/lib/python2.7/site-packages/skt/executable.py", line 1065, in main
    args.func(cfg)
  File "/home/nkondras/.local/lib/python2.7/site-packages/skt/executable.py", line 182, in wrapper
    func(cfg)
  File "/home/nkondras/.local/lib/python2.7/site-packages/skt/executable.py", line 328, in cmd_build
    join_with_slash(cfg.get('output_dir'), 'build.source')
  File "/usr/lib/python2.7/distutils/dir_util.py", line 128, in copy_tree
    "cannot copy tree '%s': not a directory" % src
distutils.errors.DistutilsFileError: cannot copy tree '/tmp/tmpj7jFNo/merge.source/': not a directory
```

But `skt --state -vv -i . -o . build -c ../config-4.17.0-3-amd64` works. According to the design, the input directory should be the current directory by default and the output directory should be the input directory by default, IIRC. Can you fix this?

Also, how would we need to use skt with this merged? Do we need to update the usage in the pipeline code, or will the old way work until the rest of the stages is transitioned to the new interface?

spbnick:
Aargh. So, do I understand right that we're keeping the notion of "workdir" until we're fully through with implementing the input/output directory support, or do you want to keep it?

spbnick:
I took another quick look through the code and couldn't see any immediate issues. I have one question, however: how would publishing work with the renamed tarballs and configs?

Right now the pipeline code relies on skt producing tarball and config names (BTW, why do we need to publish configs?), which are "unique" (are using hashes), and simply copies them over to the same directory. I assume that that is going to get broken with this change. Am I right?

Anyway, I think that approach is wrong, since skt cannot know where the independent runs are going to be published, and it's only the pipeline who knows. Therefore pipeline should tell skt which filenames (or prefixes thereof) to use when publishing. Do you think we can implement something like that?

veruu:
> why do we need to publish configs?

Because the currently-in-use old reporter needs access to the config in order to report it, which, at least in our pipeline, it doesn't have.

Yes, if all the kernels are being published into the same directory, then a prefix passed to publisher command would make sense (and use the original name if the option is not used). I agree with you that skt shouldn't decide by itself how the names are chosen (which is what it did up until now anyways).

I'll update the pull with the added option (hopefully) tomorrow.

spbnick:
Perhaps we can simply change the "destination" parameter to the publisher to "prefix", or such. Then prefix could include the destination directory. Not sure how that would work with *all* the publishers, though.

veruu:
If I understand how you mean it, then it sounds pretty confusing from the UX standpoint. And there's no guarantee it works with all publisher we have or might include in the future, as you say.

spbnick:
Yeah, it's just something I thought could work and could make it simpler. It's what e.g. `git archive` and `git checkout-index` do, as well as some other tools. I would leave it up to you to find out what's better, though.

veruu:
@spbnick option added and publishers updated.

danghai:
@veruu Does the new commit resolve the issue #350 ?

veruu:
@danghai no. Please check the commit message and the actual code if you want to see what changed.

skt/executable.py
```diff
@@ -928,6 +934,12 @@ def load_config(args):
         cfg['publisher'] = [config.get('publisher', 'type'),
                             config.get('publisher', 'destination'),
                             config.get('publisher', 'baseurl')]
+        if config.has_option('publisher', 'prefix'):
+            cfg['publisher'].append(config.get('publisher', 'prefix'))
+        else:
+            cfg['publisher'].append('')
+    elif cfg.get('publisher'):
+        cfg['publisher'].append(cfg['prefix'])
```
spbnick: It's hard to figure out what's going on here. Do you think we can rename these variables eventually?

spbnick:
BTW, @veruu, do you think you can add some documentation regarding how the new -i/-o options work and interact with e.g. --workdir? Because I'm getting confused about that all the time and I'm sure others would too :) Maybe a separate PR?

skt/executable.py
veruu: see the new issue #350 which I submitted when I saw how the configuration is parsed

veruu:
Yes, I do plan to add the docs once everything is in place. Runner still needs to be integrated into the system and this one is a bit trickier than the other ones (didn't have enough time to think about it properly yet).

spbnick:
Thank you. Could you perhaps explain briefly here, how --workdir should work with -i/-o options at the moment, so I could test it, and check if it makes sense?

skt/executable.py
spbnick: I think #350 is spot on, but I was talking about the `cfg` vs. `config` variable names here, and how difficult it is to make sense of which means what.

veruu: Ah, ok. I agree that it's confusing. Maybe submit another issue for clear variable naming if there isn't one?

veruu:
Sure. If -i isn't used, workdir is assigned to it by default. If -o isn't used, -i is assigned by default. So if neither of -i/-o options are present, workdir is used for both. This can lead to following scenarios:

no -i/-o options -> workdir is used (tmpdir is created in case workdir is not specified)
-i, -o are specified -> workdir is ignored for merge and build
-i -> -o is the same as -i
-o -> workdir is used for -i

This goes with your specification from here: https://github.com/RH-FMK/meta/pull/6/files#diff-38e228d3cd2e344915a60edf44b233f6R20
After the rest of stages will work with the -i/-o, current directory will be used instead of workdir, as outlined in the section I linked (workdir seems like a good default to use for now, to avoid breaking existing setups before the whole system is in place while still verifying what happens on the background works).

Let me know if you have other questions!

skt/executable.py
spbnick: Done: #351 :)

PR #341, opened 2018-09-03
Use defusedxml for XML parsing

veruu:
In theory, a user can pass a maliciously crafted XML, exploiting the
server/testing side. While currently it means they would exploit their
own infrastructure, if in the future there is a publicly available
service allowing people to run eg. custom tests or pass their own
Beaker XMLs, the possibility to expoit the organization running the
service grows exponentially.

xml.ElementTree is safe against external entity expansion, DTD retrieval
and decompression bomb, but not against exponential entity expansion or
quadratic blowup entity expansion. defusedxml provides a modified
fromstring method for parsing untrusted XMLs, to prevent the exploits
original ElementTree is not protected against.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,6 +17,7 @@ include_package_data = True
 install_requires = flake8
                    junit_xml
                    mock
+                   defusedxml
                    responses
                    pycodestyle!=2.4.0
                    pylint
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -20,6 +20,7 @@ import subprocess
 import time
 import xml.etree.ElementTree as etree
 
+from defusedxml.ElementTree import fromstring
 import requests
 
 from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -17,6 +17,7 @@ include_package_data = True
 install_requires = flake8
                    junit_xml
                    mock
+		   defusedxml
```
spbnick: Am I seeing this indent right? Is there a tab in there?

veruu: /me checks vimrc copied over from the old laptop and fixes the issue both there and in the commit

veruu:
The XML is parsed in skt, so it would be the place skt is executed that gets most of the mess if something is wrong, and only then Beaker. It was more or less directly said that skt will not be a regular supported package so I think the RPM dependency restriction is lifted. In any case, the `defusedxml` module is available as RPM on Fedora 28 so there's a chance it's in some extra repos for RHEL as well (can't check that as I don't have the list of packages nor a RHEL machine)

spbnick:
Looks good, thank you, Veronika! Merging this now, although I thought (a bit too late) that just unqualified `fromstring()` method might be confusing, something like `ElementTree.fromstring()` could've been better, IMO.

veruu:
@danghai "available package" is not the same as "installed by default". Of course you need to install it, the package name is `python2-defusedxml` (since you installed it via `pip` you don't have to install the RPM)

PR #345, opened 2018-09-05
Remove cmd_cleanup

### LINKED ISSUE #342: Remove cmd_cleanup?
This needs some thought, but there are some hints that the command doesn't make much sense with our plans:

1. The `--wipe` option is pretty drastic, especially if the working directory is a cached git repo or contains additional files/directories that the user forgot about; and hence is not used in practice
2. We are slowly migrating to input/output directory structure and workdir usage will be deprecated. Since each stage can use different directories, it will become impossible to clean up without specifying each of them on the command line of the cleanup command. At this point, the command won't be any easier than manual removal of the residue.
3. Related to the above, the state will not need to be used anymore either, and hence won't need to be removed in this command.

@spbnick, @major - do you have any comments on why this is a good/bad idea?

danghai:
I make PR to remove `cmd_cleanup`.
Fixes #342

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,18 +109,13 @@ The following commands are supported by `skt`:
 * `console-check`
     - Check the specified console logs for any traces, and report the first
       trace found for each console log.
-* `cleanup`
-    - Remove the build information file, kernel tarball. Remove state
-      information from the configuration file, if saving state was enabled
-      with the global `--state` option, and remove the whole working directory,
-      if the global `--wipe` option was specified.
 * `all`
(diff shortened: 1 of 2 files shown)


danghai:
@veruu I update the `--wipe`

veruu:
Functionality looks good, thanks for the changes! However, can you provide more information in the commit message, so it's easy to get information on *why* the changes happened from it?

danghai:
@veruu I add your comment in commit message.

PR #347, opened 2018-09-06
Minor fix

danghai:
I make PR fix some minor.

### DIFF
--- a/skt/console.py
+++ b/skt/console.py
@@ -11,6 +11,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Class for managing ConsoleLog."""
 import gzip
 import re
 import StringIO
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -190,7 +190,7 @@ def cmd_merge(cfg):
     """
     global retcode
     # Counter merge patch following:
-    # id[0]:  counter of merge-ref option.
+    # idx[0]: counter of merge-ref option.
     # idx[1]: counter of patch option.
     # idx[2]: counter of pw option.
     idx = [0, 0, 0]
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -150,12 +150,12 @@ class KernelBuilder(object):
             self.__prepare_kernel_config()
 
         args = self.make_argv_base + ["kernelrelease"]
-        mk = subprocess.Popen(args, stdout=subprocess.PIPE)
-        (stdout, _) = mk.communicate()
+        make = subprocess.Popen(args, stdout=subprocess.PIPE)
+        (stdout, _) = make.communicate()
         for line in stdout.split("\n"):
-            m = re.match(r'^\d+\.\d+\.\d+.*$', line)
-            if m:
-                krelease = m.group()
+            match = re.match(r'^\d+\.\d+\.\d+.*$', line)
+            if match:
+                krelease = match.group()
                 break
 
         if krelease is None:
--- a/skt/kerneltree.py
+++ b/skt/kerneltree.py
@@ -259,6 +259,7 @@ class KernelTree(object):
             match = re.match('Fetch URL: (.*)', line)
             if match:
                 return match.group(1)
+        return None
 
     def __get_remote_name(self, uri):
         remote_name = (uri.split('/')[-1].replace('.git', '')
--- a/skt/kerneltree.py
+++ b/skt/kerneltree.py
@@ -284,6 +284,17 @@ class KernelTree(object):
         return remote_name
 
     def merge_git_ref(self, uri, ref="master"):
+        """
+        Merge a git branch/reference into the tree.
+
+        Args:
+            uri: URL points to a repo containing the branch/reference to merge
+            ref: Reference to checkout, default is master.
+
+        Return:
+            A tuple (SKT_SUCCESS, reference to commit) on success.
+            A tuple (SKT_FAIL, None)                   on failure.
+        """
         remote_name = self.__get_remote_name(uri)
         head = None
 
@@ -317,6 +328,15 @@ class KernelTree(object):
         return (SKT_SUCCESS, head)
 
     def merge_patchwork_patch(self, uri):
+        """
+        Apply a patch from Patchwork (using git am)
+
+        Args:
+            uri: URL of patch on a Patchwork instance.
+
+        Raises:
+            PatchApplicationError in case the patch failed to apply.
+        """
         patch_content = get_patch_mbox(uri)
 
         logging.info("Applying %s", uri)
@@ -339,6 +359,15 @@ class KernelTree(object):
             )
 
     def merge_patch_file(self, path):
+        """
+        Apply a particular local patch run.
+
+        Args:
+            path: A full path of local patch.
+
+        Raises:
+            PatchApplicationError in case the patch failed to apply.
+        """
         if not os.path.exists(path):
             raise Exception("Patch %s not found" % path)
 
--- a/skt/kerneltree.py
+++ b/skt/kerneltree.py
@@ -143,11 +143,11 @@ class KernelTree(object):
         # Does the repo have an origin set?
         if 'origin' in self.__git_cmd("remote").split('\n'):
             # Ensure the origin is set to the correct URL
-            logging.debug("Setting URL for remote 'origin': %s" % self.uri)
+            logging.debug("Setting URL for remote 'origin': %s", self.uri)
             self.__git_cmd("remote", "set-url", "origin", self.uri)
         else:
             # Add the origin remote
-            logging.debug("Adding missing remote 'origin': %s" % self.uri)
+            logging.debug("Adding missing remote 'origin': %s", self.uri)
             self.__git_cmd("remote", "add", "origin", self.uri)
 
     def get_commit_date(self, ref=None):
--- a/skt/kerneltree.py
+++ b/skt/kerneltree.py
@@ -262,6 +262,15 @@ class KernelTree(object):
         return None
 
     def __get_remote_name(self, uri):
+        """
+        Get the remote name of a specific URI.
+
+        Args:
+            uri: The URI of the repository's remote.
+
+        Returns:
+            A string representing the remote name.
+        """
         remote_name = (uri.split('/')[-1].replace('.git', '')
                        if not uri.endswith('/')
                        else uri.split('/')[-2].replace('.git', ''))
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -11,7 +11,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-
+"""Class for managing Runner."""
 import logging
 import os
 import platform
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -87,12 +87,12 @@ class BeakerRunner(Runner):
                     if line:
                         hostnames.append(line)
         except (IOError, OSError) as exc:
-            logging.error('Can\'t access {}!'.format(filepath))
+            logging.error('Can\'t access %s!', filepath)
             raise exc
         except TypeError:
             logging.info('No hostname blacklist file passed')
 
-        logging.info('Blacklisted hostnames: {}'.format(hostnames))
+        logging.info('Blacklisted hostnames: %s', hostnames)
         return hostnames
 
     def __getxml(self, replacements):
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -381,9 +381,9 @@ class BeakerRunner(Runner):
         (stdout, _) = bkr.communicate(xml)
 
         for line in stdout.split("\n"):
-            m = re.match(r"^Submitted: \['([^']+)'\]$", line)
-            if m:
-                jobid = m.group(1)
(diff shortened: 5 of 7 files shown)


skt/kernelbuilder.py
```diff
@@ -150,12 +150,12 @@ def getrelease(self):
             self.__prepare_kernel_config()
 
         args = self.make_argv_base + ["kernelrelease"]
-        mk = subprocess.Popen(args, stdout=subprocess.PIPE)
-        (stdout, _) = mk.communicate()
+        proc = subprocess.Popen(args, stdout=subprocess.PIPE)
```
veruu: I'd prefer the name to be more descriptive/specific - while the `mk` name is definitely not either of them, the first thing some of us will think of when reading `proc` would be the `/proc` directory and not make process.

skt/runner.py
```diff
@@ -87,12 +87,12 @@ def __load_blacklist(self, filepath):
                     if line:
                         hostnames.append(line)
         except (IOError, OSError) as exc:
-            logging.error('Can\'t access {}!'.format(filepath))
+            logging.error('Can\'t access %s!', filepath)
```
veruu: Can you please explain in the commit message *why* this change is needed?

skt/runner.py
```diff
@@ -173,7 +173,7 @@ def __getresults(self):
             logging.error('All test sets aborted or were cancelled!')
             return SKT_ERROR
 
-        for job, recipe_sets in self.job_to_recipe_set_map.items():
+        for recipe_sets in self.job_to_recipe_set_map.itervalues():
```
veruu: I'd prefer to use unnamed variable (`_`) instead of this change - there is no `itervalues()` method in Python3 so while this change solves the unused variable issue, it complicates the issue about Python3 compatibility.

skt/kerneltree.py
```diff
@@ -248,6 +248,7 @@ def __get_remote_url(self, remote):
         Returns:
             URL of the remote, None if the remote doesn't exist.
         """
+        url = None
```
veruu: This is a revert of commit cce8dc23. Why complicate the code to achieve the same functionality?

danghai: The `pylint` complains about it. No return value if `for` loop does not execute, or no match in `if` condition

skt/kernelbuilder.py
danghai: @veruu I am not sure about what `mk` means?  It implies `make kernel` ?

skt/kerneltree.py
danghai: I see it look the same logic here: https://github.com/RH-FMK/skt/blob/master/skt/runner.py#L383

skt/kernelbuilder.py
veruu: It probably means `make` (kernel people have a tendency to ignore vowels when naming things), the name was there before I got involved with the project so I'm not 100% sure.

skt/kerneltree.py
veruu: What pylint version / configuration are you using? I have seen no such complaints when running it. The code you are linking to should be simplified too, but we can't do everything at the same time :)

danghai: @veruu my pylint version `1.9.2` :). What about you?

veruu: Thanks for the help, this warning seems like a newer addition. Okay, let's see how to resolve it without complicating the code too much. What about adding `return None` to the end of the method?

danghai: That sounds good!

skt/runner.py
danghai: I fix it

skt/runner.py
danghai: I add commit message

skt/kernelbuilder.py
danghai: I rename it to `make`

skt/kerneltree.py
danghai: For `runner.py`. It has raise exception if not jobid

veruu: @danghai can you please also clean up the commit message? "Return None if no URL exists", with a mention of the pylint warning it resolves should be enough

skt/runner.py
```diff
@@ -173,7 +173,7 @@ def __getresults(self):
             logging.error('All test sets aborted or were cancelled!')
             return SKT_ERROR
 
-        for job, recipe_sets in self.job_to_recipe_set_map.items():
+        for (_, recipe_sets) in self.job_to_recipe_set_map.items():
```
veruu: Is the addition of parens really needed here?

skt/runner.py
veruu: The commit message doesn't explain *why* the change was needed. It quotes the documentation, but seeing only the message, noone would know what kind of change it describes. Again, mentioning that it resolves a pylint warning (and which one) would provide more useful information.

danghai:
@veruu I have updated, and add 1 commit for docstring.

skt/kerneltree.py
```diff
@@ -143,11 +143,11 @@ def __setup_repository(self):
         # Does the repo have an origin set?
         if 'origin' in self.__git_cmd("remote").split('\n'):
             # Ensure the origin is set to the correct URL
-            logging.debug("Setting URL for remote 'origin': %s" % self.uri)
```
veruu: Can you also mention the warning string that is being resolved in the commit message? It would provide more useful information.

skt/kerneltree.py
```diff
@@ -259,8 +259,18 @@ def __get_remote_url(self, remote):
             match = re.match('Fetch URL: (.*)', line)
             if match:
                 return match.group(1)
+        return None
```
veruu: Same in this commit, mentioning the name of the specific warning is more useful than saying "something about the return value" :)

skt/runner.py
```diff
@@ -87,12 +87,12 @@ def __load_blacklist(self, filepath):
                     if line:
                         hostnames.append(line)
         except (IOError, OSError) as exc:
-            logging.error('Can\'t access {}!'.format(filepath))
```
veruu: Here as well, can you mention the specific warning in the commit message that is being solved?

skt/console.py
```diff
@@ -11,6 +11,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Class for managing ConsoleLog."""
```
veruu: This commit definitely needs to be split. If nothing else, then into one adding the module docstrings, one adding the docstrings for the `merge_*` methods and another one for the `__get_remote_name`, because all of them are unrelated.

skt/kerneltree.py
```diff
@@ -259,8 +259,18 @@ def __get_remote_url(self, remote):
...
 
     def __get_remote_name(self, uri):
+        """
+        Get the remote name of a specific URI.
+
+        Args:
+            uri: The URI of the repository's remote.
+
+        Returns:
+            Remote name's string, None if the uri doesn't exist.
```
veruu: The wording is a bit awkward. The first part would sound better if it was something like `A string representing the name of the remote`, what do you think?

skt/kerneltree.py
```diff
@@ -274,6 +284,16 @@ def __get_remote_name(self, uri):
         return remote_name
 
     def merge_git_ref(self, uri, ref="master"):
+        """
+        Merge the patch from source kernel Git reference.
```
veruu: This merges a git branch/reference, not a patch

skt/kerneltree.py
```diff
@@ -274,6 +284,16 @@ def __get_remote_name(self, uri):
...
+
+        Args:
+            uri: The source kernel Git repo URL.
```
veruu: This should mention that the URL points to the repo containing the branch/reference to merge, to distinguish form the repo which it is being merged into

skt/kerneltree.py
```diff
@@ -274,6 +284,16 @@ def __get_remote_name(self, uri):
...
+            ref: Reference to checkout, default is master.
+
+        Raises:
+            Exception in case the failed to apply patch
```
veruu: There is no patch to apply, and no exception being explicitly raised. OTOH, there are return values not mentioned in the docstring, and they should be.

skt/kerneltree.py
```diff
@@ -307,6 +327,15 @@ def merge_git_ref(self, uri, ref="master"):
         return (SKT_SUCCESS, head)
 
     def merge_patchwork_patch(self, uri):
+        """
+        Merge the patch from Patchwork run.
```
veruu: What does "patchwork run" mean? Do you only mean that the patch to be applied lives on Patchwork? In that case, something like "Apply a patch from Patchwork (using `git am`)" would be a better description of the method. If you mean something else, please do elaborate and I can help you figure out the right wording :)

skt/kerneltree.py
```diff
@@ -307,6 +327,15 @@ def merge_git_ref(self, uri, ref="master"):
...
+
+        Args:
+            uri: URL of patch on a Patchwork instance.
+
+        Raises:
+            Exception in case the failed to apply patch
```
veruu: This should mention the specific `PatchApplicationError` instead of generic `Exception`

skt/kerneltree.py
```diff
@@ -329,6 +358,15 @@ def merge_patchwork_patch(self, uri):
             )
 
     def merge_patch_file(self, path):
+        """
+        Merge the particular local patch run.
```
veruu: Same comments as with the `merge_patchwork_patch` apply here

skt/kerneltree.py
danghai: What about `A string representing the remote name`?

skt/console.py
danghai: I agree.

danghai:
@veruu I updated it based on your review. Thank you

skt/kerneltree.py
veruu: Yeah, I agree that the name of the method is slightly confusing. I think the merge prefix comes from the "merge" command, instead of git merge. And now we are back to the unclear naming which you mentioned before :)

danghai: Ah I see.

skt/kerneltree.py
```diff
@@ -259,8 +259,19 @@ def __get_remote_url(self, remote):
...
     def __get_remote_name(self, uri):
+        """
+        Get the remote name of a specific URI.
+
+        Args:
+            uri: The URI of the repository's remote.
+
+        Returns:
+            A string representing the remote name
+            None if the uri doesn't exist.
```
veruu: Checking the code, I don't think this method will ever return `None`. Invalid URLs would throw an exception on the parsing and the valid ones will have a value assigned. Do you have an example where `None` can be returned?

skt/kerneltree.py
```diff
@@ -274,6 +285,17 @@ def __get_remote_name(self, uri):
...
     def merge_git_ref(self, uri, ref="master"):
+        """
+        Merge a git branch/reference from source kernel Git reference.
+
+        Args:
+            uri: URL points to a repo containing the branch/reference to merge
+            ref: Reference to checkout, default is master.
+
+        Return:
+            SKT_SUCCESS on success merging a git branch/reference
```
veruu: A tuple is being returned, not a single constant.

skt/kerneltree.py
```diff
@@ -274,6 +285,17 @@ def __get_remote_name(self, uri):
         return remote_name
 
     def merge_git_ref(self, uri, ref="master"):
+        """
+        Merge a git branch/reference from source kernel Git reference.
```
veruu: The source mentioned here is still a bit confusing. It's great that the argument contains the description (and please keep it there!) but I'd simplify this sentence to just say "Merge a git branch/reference into the tree.", to make it less confusing. What do you say?

skt/kerneltree.py
```diff
@@ -307,6 +329,15 @@ def merge_git_ref(self, uri, ref="master"):
...
 
     def merge_patchwork_patch(self, uri):
+        """
+        Apply a patch from Patchwork (using git am)
+
+        Args:
+            uri: URL of patch on a Patchwork instance.
+
+        Raises:
+            PatchApplicationError in case the failed to apply patch
```
veruu: Can we adjust the working a bit here and with `merge_patch_file` too? "... in case the patch failed to apply" or "... in case of patch application failure" would sound a lot better

skt/kerneltree.py
danghai: Yeah. I agree

danghai:
@veruu I resolved the docstrings.

skt/kerneltree.py
```diff
@@ -274,6 +284,17 @@ def __get_remote_name(self, uri):
...
     def merge_git_ref(self, uri, ref="master"):
+        """
+        Merge a git branch/reference into the tree.
+
+        Args:
+            uri: URL points to a repo containing the branch/reference to merge
+            ref: Reference to checkout, default is master.
+
+        Return:
+            A tuple (SKT_SUCCESS, reference to commit) on success merging.
```
veruu: I'd remove the "merging" from the end (also on the line below). Or use different wording, but this one is easier :)

skt/executable.py
```diff
@@ -190,7 +190,7 @@ def cmd_merge(cfg):
     """
     global retcode
     # Counter merge patch following:
-    # id[0]:  counter of merge-ref option.
+    # idx[0]: counter of merge-ref option.
```
veruu: "Fix typo in the comment" would be better to use in this commit message, what do you say?

danghai:
@veruu I fix it.

PR #352, opened 2018-10-03
reporter - task urls and make command

veruu:
Maintain the changes in both stage and full reports until we figure out which one (if any) is useful with the factory concept

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -477,14 +477,20 @@ class BeakerRunner(Runner):
                     recipe_tests = self.get_recipe_test_list(recipe)
                     report_string += 'We ran the following tests:\n'
                     for test_name in recipe_tests:
-                        test_result = recipe.find(
+                        task_node = recipe.find(
                             "task[@name='{}']".format(test_name)
-                        ).attrib.get('result')
+                        )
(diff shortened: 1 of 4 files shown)


skt/runner.py
```diff
@@ -477,15 +477,21 @@ def run(self, url, max_aborted, release, wait=False,
...
                         report_string += '  - {}: {}\n'.format(
                             test_name, test_result.upper()
                         )
                         if test_result != 'Pass':
                             failed_tasks.append(test_name)
 
+                        if task_node.find('fetch') is not None:
+                            report_string += '    - Test URL: {}\n'.format(
+                                task_node.find('fetch').attrib.get('url')
+                            )
```
spbnick: Those URLs we have there don't make much sense to people, when they have that "fragment" part, after `#`, but it's something to start with.

veruu: The fragment only points to a specific branch/test and so if people choose to browse the repo instead of downloading the test they can still do it based on the information. I agree that the direct URL to the repo would be better in the long run, but I don't want to go on URL parsing spree in the report, especially if someone is running skt with their own tests with different URL formats. So we would need to figure out a better way to get the canonical URL (does Beaker eat those?)

PR #353, opened 2018-10-05
Don't report skipped tests

veruu:
Some tests require a specific architecture or hardware to run on. This
is handled by a test reporting "Skip" result and exiting. Detect skipped
tests and exclude them from the list of ran tests (since they didn't
actually run) and the report, to avoid confusion.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -293,21 +293,23 @@ class Reporter(object):
                     failed_tasks.append('/distribution/kpkginstall')
                 else:
                     recipe_tests = runner.get_recipe_test_list(recipe)
-                    result += ['We ran the following tests:']
-                    for test_name in recipe_tests:
-                        task_node = recipe.find(
-                            "task[@name='{}']".format(test_name)
-                        )
-                        test_result = task_node.attrib.get('result')
-                        result += ['  - {}: {}'.format(test_name,
-                                                       test_result.upper())]
-                        if test_result != 'Pass':
-                            failed_tasks.append(test_name)
+                    if recipe_tests:
+                        result += ['We ran the following tests:']
+                        for test_name in recipe_tests:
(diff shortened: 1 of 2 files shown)


skt/reporter.py
```diff
@@ -293,21 +293,23 @@ def __getjobresults(self):
...
-                        result += ['  - {}: {}'.format(test_name,
-                                                       test_result.upper())]
-                        if test_result != 'Pass':
-                            failed_tasks.append(test_name)
-
-                        if task_node.find('fetch') is not None:
-                            result.append('    - Test URL: {}'.format(
-                                task_node.find('fetch').attrib.get('url')
-                            ))
+                    if recipe_tests:
```
spbnick: Is adding this condition necessary for the main change? Or is this just fixing something related?

skt/runner.py
```diff
@@ -475,22 +476,23 @@ def run(self, url, max_aborted, release, wait=False,
...
-                        report_string += '  - {}: {}\n'.format(
-                            test_name, test_result.upper()
-                        )
-                        if test_result != 'Pass':
-                            failed_tasks.append(test_name)
-
-                        if task_node.find('fetch') is not None:
-                            report_string += '    - Test URL: {}\n'.format(
-                                task_node.find('fetch').attrib.get('url')
+                    if recipe_tests:
```
spbnick: And same here?

skt/reporter.py
veruu: There's a possibility all the tests that were supposed to be ran had to be skipped, so there are no actual tests that were run. The condition makes sure we don't say "We ran the following tests" followed by an empty line in the report.

spbnick: Makes sense, thank you, Veronika :)

PR #359, opened 2018-10-19
[WIP] Report jinja2 conversion

major:
This is still a work in progress. All of the tests are broken (YAY!) and that will require a little more work.


spbnick:
Hi Major, I like that you're getting the whole report into templates. I have
just a few comments/questions from a quick glance at the code, but please
shout if you'd like a proper review.

* What is your general idea/design/plans for this approach?
* Do you think making the summary more compact would be good?
  I.e. instead of writing this:
  ```
      +----------------+---------------+
      | OVERALL RESULT | PASSED        |
      +================+===============+
      | Patch merge    | PASSED        |
      +----------------+---------------+
      | Kernel compile | PASSED        |
      +----------------+---------------+
      | Hardware test  | PASSED        |
      +----------------+---------------+
  ```
  Write this:

  ```
           OVERALL: PASSED
       Patch merge: PASSED
    Kernel compile: PASSED
     Hardware test: PASSED
  ```
  I think this way there's less visual noise and it takes fewer lines.

  Maybe go as far as shortening the stage names, so they're faster to read and
  identify? I.e.:
  ```
        OVERALL: PASSED
          Merge: PASSED
          Build: PASSED
           Test: PASSED
  ```
  I think it will be OK as is, though.

* Do you think you could pass the `multireport_failed` values as strings to
  the template, so the conditions are easier to understand?

skt/templates/report.j2
```diff
@@ -0,0 +1,95 @@
...
+We ran automated tests on a patchset that was proposed for merging into this
+kernel tree. The results of these automated tests are provided below.
+{% else %}
+{# We ran baseline tests against a kernel tree. #}
+We ran automated tests on a recent commit from this kernel tree. The results
+of these automated tests are provided below.
+{% endif %}
+
+{# What is the overall job result? #}
+{% if multireport_failed == 1 %}
```
spbnick: I think this would be more readable, if we used the status name, instead of value. E.g.:

    {% if multireport_failed == "MERGE" %}

or:

    {% if multireport_failed == "MULTI_MERGE" %}

or even, if we want a better protection from typos, we could export those numbers as variables and instead write:

    {% if multireport_failed == MULTI_MERGE %}

That would match the Python code! Oh, and I see that we also have greater/lesser comparison, and they would work with this too.

major:
Thanks for the comments, folks. I plan to split this into more than one commit, but I'm trying to get things working. The templates are working fine (but I'd like to make them more terse) but the tests all expect the old email format. I also wrote some fairly hacky tests in the past (oops) and that's coming back to bite me now.

Using the strings for `multireport_failed` makes sense and I'll do that.

major:
This is over in #362 now.

PR #363, opened 2018-10-24
Remove version constraint for pycodestyle (new release of flake8)

agustinhenze:
On the last release of flake8 the dependency version for pycodestyle was bumped to >= 2.4.0 and there is no newer version for pycodestyle yet i.e. skt is not installable.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,14 +12,12 @@ packages = find:
 # Parse the MANIFEST.in file and include those files, too.
 include_package_data = True
 # Let pip install dependencies automatically.
-# NOTE(mhayden): pycodestyle 2.4.0 has a bug that causes flake8 to fail.
-# Remove the pycodestyle constraint once upstream is fixed.
 install_requires = flake8
                    junit_xml
                    mock
                    defusedxml
                    responses
-                   pycodestyle!=2.4.0
+                   pycodestyle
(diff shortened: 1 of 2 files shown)


spbnick:
The change is good, of course, but why do we get Travis CI failures?

agustinhenze:
> The change is good, of course, but why do we get Travis CI failures?

Because the new release of `flake8` :man_facepalming:.
```
flake8 runtests: commands[0] | flake8 --show-source .
./skt/kernelbuilder.py:79:11: W605 invalid escape sequence '\g'
        return re.sub(r"[]*?[]", "[\g<0>]", pathname)
          ^
ERROR: InvocationError for command '/home/travis/build/RH-FMK/skt/.tox/flake8/bin/flake8 --show-source .' (exited with code 1)
```
And it seems a false positive...

veruu:
Marking the second string as raw should fix it, unless anything changed since I ran into the error

agustinhenze:
Oh, you're right I saw the `r` on the beginning and got confused... Pushing the fix

agustinhenze:
This PR will be needed for `production` and for `staging`  otherwise all pipelines will fail.

spbnick:
And they're failing, thank you, Agustin!

PR #368, opened 2018-10-24
Make CONFIG_LOCALVERSION configurable

major:
Allow users to specify the value for `CONFIG_LOCALVERSION` with the
`--localversion` argument.

### DIFF
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -32,7 +32,7 @@ from skt.misc import join_with_slash
 class KernelBuilder(object):
     def __init__(self, source_dir, basecfg, cfgtype=None,
                  extra_make_args=None, enable_debuginfo=False,
-                 rh_configs_glob=None):
+                 rh_configs_glob=None, localversion=None):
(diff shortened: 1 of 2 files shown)


skt/kernelbuilder.py
```diff
@@ -102,8 +103,12 @@ def __prepare_kernel_config(self):
         if not self.enable_debuginfo:
             self.__adjust_config_option('disable', 'debug_info')
 
-        # Set the CONFIG_LOCALVERSION to '.skt'
-        self.__adjust_config_option('set-str', 'LOCALVERSION', '.skt')
+        # Set CONFIG_LOCALVERSION
+        self.__adjust_config_option(
+            'set-str',
+            'LOCALVERSION',
+            '.{}'.format(self.localversion)
```
spbnick: Please be aware, that with the code as is skt can be made to execute arbitrary code passed via the `--localversion` option, and we need to take additional measures if we want to prevent that. Perhaps we need an issue for a security review of these things.

PR #374, opened 2018-10-26
Don't get runner results if tests didn't run

major:
Skip the runner report gathering if the runner never ran.


spbnick:
This looks right, but it would help if you could describe the fault you're fixing, and conditions that bring it.

PR #380, opened 2018-11-02
Be specific about kernel architectures in reports

major:
The reports from tests use `powerpc` for both big and little endian systems. This PR captures the cross compiler (if it is specified) and then uses it for reports.

### DIFF
--- a/skt/templates/partials/build_details.j2
+++ b/skt/templates/partials/build_details.j2
@@ -4,7 +4,8 @@
 We compiled the kernel for {{ jobs_run }} {{ arch_plural }}:
 
   {% for job in report_jobs %}
-  {{ job.kernel_arch }}:
+  {% set kernel_arch = job.kernel_compiler.split('-')[0] if job.kernel_compiler else job.kernel_arch %}
+  {{ kernel_arch }}:
     make options: {{ job.make_opts }}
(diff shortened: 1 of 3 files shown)


spbnick:
Thank you for wonderfully split and described commits, Major! Nicely done, and easy to review :)

I have only one concern with the code: despite its name, the `__get_build_compiler()` function returns cross compiler only. Could you please make the name match the function? I.e. either name it something like `__get_build_cross_compiler()`, or change the function to always return the compiler name (e.g. by capturing the (stripped) output of `${CC:-${CROSS_COMPILE}gcc} -dumpmachine` shell command). I would prefer the latter, if you could pull it off, but the former would work fine too :)

Thank you!

spbnick:
Actually, as is, the function should probably be called either get `__get_cross_compile()` (after the environment variable name), or something like `__get_build_cross_compiler_prefix` (since it's not returning the full compiler name).

spbnick:
Oh, and if you change the function name, the template variable name would need to be changed too, I think.

major:
@spbnick Okay, I've changed up the names. Adding the `${CC:-${CROSS_COMPILE}gcc} -dumpmachine` sounds like a good idea but that might be a good enhancement for later. ;)

PR #382, opened 2018-11-05
Add testing details to all reports

major:
This PR adds details to the report about the lists of hardware tests that ran whether they passed or not. This is especially useful for success reports since it shows developers what was tested.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -273,6 +273,9 @@ class Reporter(object):
         for recipe_set_result in recipe_set_results:
             for recipe in recipe_set_result.findall('recipe'):
 
+                # Set up a list of all the tasks that passed in this job.
+                passed_tasks = []
+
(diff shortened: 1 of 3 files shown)


skt/reporter.py
```diff
@@ -308,6 +315,9 @@ def __getjobresults(self):
                     # tasks.
                     if task_result != 'Pass':
```
spbnick: Not asking to change this now, but I think it is better to streamline conditions when you can: e.g. remove extra negations, so they're easier to read and reason about. In this case, since you added the second branch, it was a good opportunity to do an equals comparison instead, and switch the branches. I.e.:

    if task_result == 'Pass':
        test_name = task_node.attrib.get('name')
        passed_tasks.append(test_name)
    else:
        failed_tasks.append(task_node)

PR #383, opened 2018-11-06
Get cross compiler prefix

spbnick:
Use cross compiler prefix in report if available

The cross compiler prefix is usually more descriptive than the kernel
arch used to build the kernel. As an example, 'powerpc' appears for both
big and little endian but the compiler names are different.

This patch uses the cross compiler prefix in the report if it is
available.

This is a remake of #380.

### DIFF
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -152,6 +152,18 @@ class KernelBuilder(object):
 
         return platform.machine()
 
+    def __get_cross_compiler_prefix(self):
+        """
+        Determine the cross compiler prefix for the kernel build.
+
+        Returns:
+            The cross compiler prefix, if defined in the environment.
+        """
+        if 'CROSS_COMPILE' in os.environ:
+            return os.environ['CROSS_COMPILE']
+
+        return None
+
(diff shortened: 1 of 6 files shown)


skt/kernelbuilder.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, source_dir, basecfg, cfgtype=None,
         self.make_argv_base = ["make", "-C", self.source_dir]
         self.enable_debuginfo = enable_debuginfo
         self.build_arch = self.__get_build_arch()
+        self.cross_compiler_prefix = self.__get_cross_compiler_prefix()
```
agustinhenze: `os.environ.get('CROSS_COMPILE', None)` :man_shrugging:

spbnick: True. Do you mean that it would have been better to write that instead of having that function?

agustinhenze: Exactly, less code to maintain, less documentation, etc hehehe. Lazy programmer :man_shrugging:

spbnick: Agreed. Would you care to submit a fix?

PR #386, opened 2018-11-06
Clarify 'hardware tests' in report

major:
The wording for the testing details is a little confusing. We're testing
the kernel on physical hardware, not performing testing of the hardware
itself. This patch clarifies the wording.

### DIFF
--- a/skt/templates/partials/test_details.j2
+++ b/skt/templates/partials/test_details.j2
@@ -1,5 +1,5 @@
 
-We booted the compiled kernel and ran the following hardware tests:
+We booted each kernel and ran the following tests:
 
 {% for job in report_jobs %}
   {{ job.kernel_arch }}:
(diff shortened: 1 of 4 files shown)


skt/templates/partials/test_details.j2
```diff
@@ -1,5 +1,6 @@
 
-We booted the compiled kernel and ran the following hardware tests:
+We booted the compiled kernel and ran the following tests on physical
+hardware:
```
spbnick: This is better, thank you, Major.
However, we often run tests in VMs. Would this sentence still make sense to people then?
Also, we have "hardware tests" in other places of the report, IIRC. Could you fix that as well?

agustinhenze: In the summary still saying `     Hardware test:`

spbnick: All I'm saying, is could we call it just "tests"?

major:
@spbnick @agustinhenze So how should we word this so that it makes the point that we're doing real tests with the kernel booted on real hardware? Sure, that hardware might be an LPAR on a POWER or s390 node, but it's still real hardware.

spbnick:
Looks good, thank you, Major!

PR #389, opened 2018-11-12
report fixes

veruu:


skt/templates/partials/test_failure.j2
```diff
@@ -2,14 +2,18 @@ One or more kernel tests failed:
...
-  {{ kernel_arch }}: FAILED
+  {# Check the failed_tasks for each recipe to figure out if anything failed #}
+  {% set result = namespace(res='PASSED') %}
+  {% for test_result in job.test_results %}
+    {% if test_result.failed_tasks %}
+      {% set result.res = 'FAILED'  %}
+    {% endif %}
+  {% endfor%}
+  {{ kernel_arch }}: {{ result.res }}
+  {% if result == 'FAILED' %}
```
spbnick: I know little to nothing about Jinja. Is it OK that we assign to `result.res`, but then check just `result`?

skt/reporter.py
```diff
@@ -297,27 +297,26 @@ def __getjobresults(self):
                 task_node = self.__get_task(recipe, test_name)
                 if task_node.attrib['result'] != 'Pass':
                     failed_tasks.append(task_node)
-                    continue
```
spbnick: I thought this `continue` was weird while reviewing the original patch, but didn't have the capacity to review this properly. Sigh...

skt/templates/partials/test_failure.j2
veruu: Good comment, ha! I was changing the previous version to the namespacing because jinja takes variable scope really seriously and didn't notice the missing jobs since I was looking at the limited template. Fixed now

spbnick:
circleci has been hanging on us for a while. @major do you know what's going on?

skt/reporter.py
```diff
@@ -329,6 +328,15 @@ def __getjobresults(self):
                         'name': task_node.attrib.get('name'),
                         'logs': logs
                     }
+                    # Add 'boot test' to the kpkginstall task if it failed to
+                    # show that it includes booting the kernel, not just
+                    # installing it. We can't do it earlier because we need the
+                    # original task to work with.
+                    kpkginstall_name = '/distribution/kpkginstall'
+                    if failed_task_detail['name'] == kpkginstall_name:
+                        failed_task_detail['name'] = '{} (boot test)'.format(
```
agustinhenze: I think we should move this logic to the jinja template and do the renaming of the task just once. Here we are renaming it twice, on the success case and here on the failed case.

PR #391, opened 2018-11-14
Stop processing the report on infrastructure error

agustinhenze:
skt uses the `exit code == 2` when it finds errors wasn't caused by the patch or kernel being tested.

### DIFF
--- a/skt/reporter.py
+++ b/skt/reporter.py
@@ -43,6 +43,10 @@ JINJA_ENV = Environment(
 )
 
 
+class InfrastuctureError(Exception):
+    """Raised when an infrastucture error was found"""
+
+
 class MultiReportFailure(enum.IntEnum):
     """IntEnum to track multireport failure statuses."""
 
@@ -424,6 +428,10 @@ class Reporter(object):
                 # If the tests failed, mark the result as a test failure.
                 if self.cfg.get('retcode') != '0':
                     self.multireport_failed = MultiReportFailure.TEST
+                if self.cfg.get('retcode') == '2':
+                    raise InfrastuctureError(
+                        "Stop processing, an InfrastuctureError was detected"
+                    )
 
                 # Collect the tests results and append them to our list.
                 job_data['test_results'] = self.__getjobresults()
--- a/tests/test_reporter.py
+++ b/tests/test_reporter.py
@@ -442,6 +442,30 @@ class TestStdioReporter(unittest.TestCase):
     @mock.patch('skt.reporter.load_state_cfg')
     @mock.patch('skt.runner.BeakerRunner.getresultstree')
     @responses.activate
+    def test_infrastructure_error(self, mock_grt, mock_load):
+        """Verify InfrastuctureError is raised when exit code == 2."""
+        self.mock_patchwork_responses()
+        mock_grt.return_value = self.beaker_fail_results
+
+        self.basecfg['result'] = ['state1', 'state2']
+
+        # Create our two mocked state files for two different arches
+        state1 = self.basecfg.copy()
+        state1['kernel_arch'] = 's390x'
+        state1['retcode'] = '1'
+        state2 = self.basecfg.copy()
+        state2['kernel_arch'] = 'x86_64'
+        state2['retcode'] = '2'
+
+        # Mock the loading of these state files
+        mock_load.side_effect = [state1, state2]
+
+        rptclass = reporter.StdioReporter(self.basecfg)
+        self.assertRaises(reporter.InfrastuctureError, rptclass.report)
+
+    @mock.patch('skt.reporter.load_state_cfg')
+    @mock.patch('skt.runner.BeakerRunner.getresultstree')
+    @responses.activate
     def test_multireport_partial_success(self, mock_grt, mock_load):
         """Verify multireport partial success output."""
         # pylint: disable=invalid-name
--- a/tests/test_reporter.py
+++ b/tests/test_reporter.py
@@ -152,12 +152,9 @@ class TestStdioReporter(unittest.TestCase):
 
         return tempfilename
 
-    @responses.activate
-    def test_merge_failure(self):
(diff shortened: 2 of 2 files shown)


agustinhenze:
> @agustinhenze, I appreciate the work you put into this, refactoring, separate commits, and tests. However, I wouldn't ACK it, unless you explain your need for it, and it turns out to be a reasonable need.
> 

Yeah sure, no problem. I will try to explain it

> With the current interface, whoever invokes skt is responsible for checking what the exit status was, and not executing the next stage if it was anything but zero, with the exception of "skt report" which can also be executed when the exit status was one. Therefore this is a redundant feature.

I believe that skt shouldn't know anything about where it is running. You can run it on your machine with a fake rc or maybe you can resume an older run. skt only has to be aware of the rc state and nothing else. We can't rely not sending our reports on not calling `skt report`, when all the information needed should be on the rc file.

spbnick:
We do that for test failures already. We make sure that the state is valid when the tests have failed (or passed, of course). And that's the only thing we need. We don't need to do anything else, "skt report" is only about sending reports when there *is* something to report. In case of infrastructure failure there's nothing to report.

I see it like this: somebody is handing a paper to a professor at a university, and is asking him/her to just ignore the work, if they find that the printer shredded it, or a dog chewed on it. I don't mean that "skt report" would be offended if you hand it an sktrc with "ret == 2" (why does it even need to be there, BTW?), but that it's simply not its job.

You need to abort your pipeline if you failed to download something, or if you ran out of space, or things like that, anyway. Don't make "skt report" responsible for a small subset of the problems, which skt can encounter and somehow report itself, but which it didn't create. We simply cannot put all the possible failures into sktrc, most of them are out of skt's scope, and we *have* to not run "skt report" for those anyway. Just check skt exit code and abort the pipeline if it's not zero or one.

You cannot resume a run from the point where it aborted, it entered an invalid state. It's gone. Kaput. Just assume that skt barfed into its sktrc. If you'd like to resume, then back-off to a last-known good state, and e.g. re-run the GitLab "job" which aborted, if that's supported.

If you still disagree, I'm willing to consider a more specific use case and perhaps I'll see what you're driving at, or the problem that you consider and we can't solve in the above way. Thank you.

agustinhenze:
Yes, I'll summarize with. It is needed to avoid sending wrong reports on the new architecture we are planing to implement. The report builder is not part of the pipeline and skt neither. It's just a tool and you can execute it on every step without taking into account any previous state.

spbnick:
I'm sorry, this is not specific enough for me to see the benefit. I see that you still disagree, and I'm still holding my opinion. However, there's not enough harm from this change so far, for me to justify our opinion lock-up, and if nobody else objects, I wouldn't oppose merging this further. I still woudln't ACK it, though.

@veruu, are you OK with merging this?

agustinhenze:
@spbnick here you can see a report sent wrongly https://paste.debian.net/1051681/. The job for s390x was aborted with exit-code == 2 and skt report sent the report anyway.

spbnick:
Well, just don't let it :) You can detect when skt returns 2 and abort the pipeline, can't you? We added that exit code **specifically** for that: aborting the pipeline.

spbnick:
If speaking in terms of your current design, then make your webhook handler ignore aborted pipelines and not run "skt report".

agustinhenze:
If you take a look to the diagram, skt report is not part of the pipeline. We can't rely on not calling it to avoid sending wrong reports, it doesn't make any sense.

spbnick:
You know I disagree and that it makes sense to me :) Do you still run `skt report` if e.g. `pip install` of skt fails for a test job?

veruu:
TBH I'm kind of on @spbnick 's side in here. If there's an infrastructure error, the pipeline should abort and we should go there and figure out what happened.

However, if that won't be easy to do in the new pipeline, I think this pull has a more serious problem - it only handles infrastructure errors from Beaker. If there's an infrastructure error during merge or build stage, it stays ignored -- the `retcode` in the state file is *not* skt's generic return code after each stage but the one that gets returned by `runner.run()`.

If there's an infra error during merge or build stage, you still don't want to send the report, and this only handles 1/3 of the cases - where the other 2/3 can be only detected by the return code.

agustinhenze:
The responsibility of the pipeline is running commands and just that. When skt exits with an exit code == 2, it's not the pipeline responsibility to decide what to do with that. Otherwise running skt in this last step is just for doing retries on beaker. We have to define clearly the responsibilities  for every component and here we are mixing the responsibility because the pipeline is not aborted.
An aborted pipeline means that the command was running couldn't finish for any reason. It was suddenly aborted.

agustinhenze:
> TBH I'm kind of on @spbnick 's side in here. If there's an infrastructure error, the pipeline should abort and we should go there and figure out what happened.
> 

Totally agree, if the pipeline is aborted. skt has nothing to do there.

> However, if that won't be easy to do in the new pipeline, I think this pull has a more serious problem - it only handles infrastructure errors from Beaker. If there's an infrastructure error during merge or build stage, it stays ignored -- the `retcode` in the state file is _not_ skt's generic return code after each stage but the one that gets returned by `runner.run()`.
> 
> If there's an infra error during merge or build stage, you still don't want to send the report, and this only handles 1/3 of the cases - where the other 2/3 can be only detected by the return code.

I think that you're absolutely right, we have to handle the 2 cases missing as well. Can I do that later?

spbnick:
I'm not buying your logic. I think there's a gap in your reasoning, and you don't understand the existing pipeline fully, judging by the comments you made here and elsewhere. I feel you have a design in mind and you're trying to bend the logic towards it. Not blaming you or anything, been there done that myself. I think we can survive that code in skt as we did with plenty of other issues, and I will get on with things and stop answering. I trust we'll figure it out in the end, one way or another :)

agustinhenze:
OK, I'm going to close it for now. Let's try to figure out together what is the best way to keep the report working on the new pipeline :).

PR #392, opened 2018-11-15
reporter thingies

veruu:


skt/reporter.py
```diff
@@ -311,14 +311,19 @@ def __getjobresults(self):
...
-                        if task_result != 'Pass':
-                            failed_tasks.append(task_node)
-                        else:
+                        if task_result == 'Pass':
                             test_name = task_node.attrib.get('name')
                             passed_tasks.append(test_name)
+                        elif (task_result == 'Warn'
+                              and task_status == 'Aborted'):
+                            # Don't add tasks that aborted to the lists
+                            continue
```
spbnick: Wouldn't this result in the task that e.g. caused the lock-up to be missing from the list of failed tests? Should we maybe only skip the ones *after* the first aborted task?

veruu: What do you mean by lock up? If it caused a panic, the result would be 'Panic' and not 'Warn' so it would be added to the list. Only the tasks that run into infrastructure/watchdog problems should end with this combination of result/status, and we shouldn't report those because they are not test failures. Is that what you had in mind?

spbnick: Um, I was thinking, what if a test caused the machine to lock up completely, so that the external watchdog had to come and kill it? Wouldn't it be a Warn/Aborted also? If so, shouldn't we report that? I mean wouldn't a machine lockup after the tests have started point to a potential kernel issue? Or do we have infrastructure issues like that regularly and so they're difficult to separate?

veruu: I don't remember seeing a kernel problem manifesting that way, but I do remember lab controller loosing the connection with the machine after reboot, resulting in complains by the external watchdog.

spbnick: Hmm. OK, I think I saw something

spbnick: Ha-ha, sorry, finger slip :grin:
I mean, I think I saw some test failure like that, but I might be mistaken and don't want to look for that right now. Let's merge this as is and see. Thank you :)

spbnick:
The circleci tests are still getting stuck. @major is not returning for another week. @agustinhenze do you know anything about circleci, could you take a look at what's going on there? I'm merging this as is.

PR #400, opened 2018-12-03
Always cancel beaker jobs on exit

wizardofoz123:
This implements FASTMOVING-357.

At the moment, if skt is terminated, it doesn't always cancel its Beaker
jobs.

This patch wraps runner.run() with cleanup handler using atexit and
signal. When skt terminates, runner's cancel_pending_jobs() is called.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -16,12 +16,14 @@ from __future__ import print_function
 import ConfigParser
 import argparse
 import ast
+import atexit
 import datetime
 import importlib
 import json
 import logging
 import os
 import shutil
+import signal
 import subprocess
 import sys
 import tempfile
@@ -443,6 +445,10 @@ def cmd_run(cfg):
     run_report_path = join_with_slash(cfg.get('output_dir'), 'run.report')
 
     runner = skt.runner.getrunner(*cfg.get('runner'))
+
+    atexit.register(runner.cleanup_handler)
+    signal.signal(signal.SIGINT, runner.signal_handler)
+    signal.signal(signal.SIGTERM, runner.signal_handler)
     retcode, report_string = runner.run(cfg.get('buildurl'),
                                         cfg.get('max_aborted_count'),
                                         cfg.get('krelease'),
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -17,6 +17,7 @@ import os
 import platform
 import re
 import subprocess
+import sys
 import time
 import xml.etree.ElementTree as etree
 
@@ -72,6 +73,9 @@ class BeakerRunner(Runner):
         self.aborted_count = 0
         # Set up the default, allowing for overrides with each run
         self.max_aborted = 3
+
+        # determines if termination cleanup was done and all jobs terminated
+        self.cleanup_done = False
 
         logging.info("runner type: %s", self.TYPE)
         logging.info("beaker template: %s", self.template)
@@ -243,6 +247,21 @@ class BeakerRunner(Runner):
         newroot.append(tmp)
 
         return newroot
+
+    def cleanup_handler(self):
+        # don't run cleanup handler twice by accident
+        if self.cleanup_done:
+            return
+
+        # skt is being terminated, cancel its jobs
+        self.cancel_pending_jobs()
+
+        self.cleanup_done = True
+
+    def signal_handler(self, signal, frame):
+        self.cleanup_handler()
+
+        sys.exit(1)
 
     def cancel_pending_jobs(self):
         """


skt/executable.py
```diff
@@ -27,6 +27,8 @@
 import tempfile
 import time
 import traceback
+import atexit
```
veruu: Can you please keep the import order of standard packages - third party packages - local packages and the alphabetical order withing the groups, so it's easier to identify what is and what isn't imported?

skt/runner.py
```diff
@@ -19,6 +19,7 @@
 import subprocess
 import time
 import xml.etree.ElementTree as etree
+import sys
```
veruu: Same here

skt/executable.py
```diff
@@ -443,11 +445,19 @@ def cmd_run(cfg):
...
 
     runner = skt.runner.getrunner(*cfg.get('runner'))
-    retcode, report_string = runner.run(cfg.get('buildurl'),
-                                        cfg.get('max_aborted_count'),
-                                        cfg.get('krelease'),
-                                        cfg.get('wait'),
-                                        arch=cfg.get("kernel_arch"))
+
+    atexit.register(runner.cleanup_handler)
+    signal.signal(signal.SIGINT, runner.signal_handler)
```
veruu: This would only hit with `SIGINT` but not with `SIGTERM`, `SIGKILL` or any other signals. Not sure what GitLab does to kill skt (maybe @agustinhenze can help here?) but it would be better to use something more generic here than `^C`

skt/executable.py
```diff
@@ -443,11 +445,19 @@ def cmd_run(cfg):
...
+    try:
```
veruu: All of the `runner.run` code that's dealing with Beaker (from job submission to waiting for completion) is already wrapped in a `try/except` block: https://github.com/CKI-project/skt/blob/master/skt/runner.py#L468

In theory, it is possible that the assignments and result interpretation that aren't in the block would cause an exception, but we don't want to catch them because the issues would be interpreted as a completed runs. Can you elaborate what advantage the addition of the block adds here?

skt/executable.py
```diff
@@ -443,11 +445,19 @@ def cmd_run(cfg):
...
+        retcode, report_string = runner.run(cfg.get('buildurl'),
+                                            cfg.get('max_aborted_count'),
+                                            cfg.get('krelease'),
+                                            cfg.get('wait'),
+                                            arch=cfg.get("kernel_arch"))
+    except Exception:
+        exc_info = sys.exc_info()
+        print('\n'.join(traceback.format_tb(exc_info[2])))
```
veruu: Please use the initialized `logging` module instead of prints. Since this part of code is hiding infrastructure issues, please also either raise the caught exception or set the `retcode` to `SKT_ERROR`.

skt/executable.py
wizardofoz123: Sure, that's reasonable. 

Also, do you have a code-style guide for SKT somewhere? I'd prefer not to learn guidelines one by one.

skt/executable.py
wizardofoz123: We can add SIGTERM, but SIGKILL by its very nature cannot be trapped. 

http://en.wikipedia.org/wiki/Unix_signal#SIGKILL

skt/executable.py
wizardofoz123: OK, probably not the best way to handle this part. There's quite a bit of code inside that isn't wrapped in any exception, but we can address that later on.

skt/executable.py
veruu: The guidelines are here, but we haven't touched them in a long time so there might be some changes in the meanwhile (eg. in the linked documents): https://github.com/CKI-project/meta/blob/master/CODING.md

skt/executable.py
veruu: Right, forgot about that detail for a while. Real fun will be if that's what GitLab/UpShift use :D

spbnick: From what I've seen so far, we have no way of handling termination on GitLab/OpenShift, we're likely getting a SIGKILL there. At least GitLab has an issue asking to switch to a softer shutdown: SIGTERM with timeout to SIGKILL.

wizardofoz123: I don't see much else we can do from this end. So how about merging this issue, leaving FM-357 open or creating a follow-up ticket and asking Don to put a bit of pressure on that GitLab/OpenShift issue?

spbnick: I would say that skt should still handle SIGTERM in addition to SIGINT. After that, and provided @veruu is OK with all the rest, I think we can merge it and close FASTMOVING-357. We need to open a separate issue regarding our GitLab pipeline not cleaning up, and mention the [upstream issue](https://gitlab.com/gitlab-org/gitlab-runner/issues/3031) there.

We cannot really put a pressure on GitLab, but we can suggest an implementation, or at least give the issue a thumbs-up in hope that it would raise the priority. @agustinhenze, is there anything else we can do there?

wizardofoz123: Look underneath, V2 has SIGTERM and SIGINT.

spbnick: Ah, great, thank you, Jakub :)

wizardofoz123:
@veruu done

wizardofoz123:
@veruu done

Issue #349, opened 2018-09-12
.report files shouldn't be created with an executable bit


veruu:

veruu:
Closing this as we'll just be getting rid of the files

PR #401, opened 2018-12-04
Fix GitHub issue #349

wizardofoz123:
We don't want report files to have an executable flag.
Let's not create report files with os.fdopen(os.open(..)), the intuitive way
sets the flags correctly by default.


veruu:
Based on the yesterday's conversation, would it be better to get rid of the stage reports altogether and close the issue?

PR #402, opened 2018-12-05
tests: add blacklist tests for BeakerRunner

wizardofoz123:


veruu:
Travis seems to be complaining. @wizardofoz123 could you deal with that first?

wizardofoz123:
@veruu : yaa, needed to add pylint to my IDE. It should pass now.

PR #403, opened 2018-12-05
runner: blacklisting: fix adding to hostrequires

wizardofoz123:
The tests I wrote revealed that appending a new node to blacklist a
hostname to hostrequires doesn't work. This patch fixes it.


spbnick:
@wizardofoz123, @veruu, should we maybe just move blacklisting from skt to kpet-db, as described in FASTMOVING-343?

wizardofoz123:
Ugh, lxml behaves differently than defusedxml even though syntax is pretty similar. This patch needn't be merged.

PR #407, opened 2018-12-07
Tests runner cleanup

wizardofoz123:
Here are 2 tests for cleanup handler and a small fix to make sure the handler gets called.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -261,7 +261,7 @@ class BeakerRunner(Runner):
     def signal_handler(self, signal, frame):
         self.cleanup_handler()
 
-        sys.exit(1)
+        sys.exit(SKT_ERROR)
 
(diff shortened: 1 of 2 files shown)


veruu:
@wizardofoz123 could you please fix the Travis failure?

wizardofoz123:
@veruu Thanks, but there's usually no need to notify me. I saw travis fail, just wasn't sure how to fix the test at first.

skt/runner.py
```diff
@@ -507,6 +507,9 @@ def run(self, url, max_aborted, release, wait=False,
                 ret = self.__getresults()
         except Exception as exc:
             logging.error(exc)
+            if exc is SystemExit:
```
agustinhenze: Why are you using `is` for the comparison? And why it's working? :thinking:

PR #408, opened 2018-12-10
Add tests for skt/runner.py:run

wizardofoz123:
This adds a couple of tests for skt/runner.py:run. The tests feed some XMLs to run() and change one of the recipes to state like Completed or Aborted to hit different code branches.

### DIFF
--- a/tests/misc.py
+++ b/tests/misc.py
@@ -14,6 +14,8 @@
 """Miscellaneous for tests."""
 import os
 
+import mock
+from defusedxml.ElementTree import fromstring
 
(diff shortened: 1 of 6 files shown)


tests/misc.py
```diff
@@ -13,7 +13,8 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 """Miscellaneous for tests."""
 import os
-
```
veruu: Can you keep the newline here to keep the import groups separate?

PR #409, opened 2018-12-10
drop separate stage reports

wizardofoz123:
Drop result and report files created by each stage.
For now we keep methods:
* report_results
* remove_oldresult

Because console module has no other ways to report.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -39,8 +39,7 @@ import skt.reporter
 import skt.runner
 from skt.kernelbuilder import KernelBuilder, CommandTimeoutError, ParsingError
 from skt.kerneltree import KernelTree, PatchApplicationError
-from skt.misc import (join_with_slash, get_patch_name, get_patch_mbox,
-                      SKT_SUCCESS, SKT_FAIL, SKT_ERROR)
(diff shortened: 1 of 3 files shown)


skt/executable.py
```diff
@@ -467,10 +383,6 @@ def cmd_run(cfg):
 
     save_state(cfg, {'retcode': retcode})
 
-    if retcode != SKT_ERROR:
-        report_results(run_result_path, 'true' if retcode else 'false',
```
veruu: The `report_string` is still being created in the runner. Is that planned to be a separate change?

wizardofoz123: @veruu Nope, just missed that.

skt/executable.py
```diff
@@ -39,8 +39,7 @@
 import skt.runner
 from skt.kernelbuilder import KernelBuilder, CommandTimeoutError, ParsingError
 from skt.kerneltree import KernelTree, PatchApplicationError
-from skt.misc import (join_with_slash, get_patch_name, get_patch_mbox,
-                      SKT_SUCCESS, SKT_FAIL, SKT_ERROR)
+from skt.misc import (join_with_slash, SKT_SUCCESS, SKT_FAIL, SKT_ERROR)
```
veruu: We shouldn't need the `()` since the import is a single line now.

PR #410, opened 2018-12-10
add tests for skt/executable

wizardofoz123:


tests/test_executable.py
```diff
@@ -238,3 +247,118 @@ def test_load_config_with_state_arg(self):
...
+            result = dict1.copy()
+            result.update(dict2)
+            return result
+
+        # cfg has to have 'state', otherwise it returns None
+        self.assertIsNone(executable.save_state({}, {}))
+
+        config_file = []
+        args = ['--rc', '/tmp/testing.ini', '--workdir', '/tmp/workdir',
+                '--state', 'run', '--runner', 'myrunner', '[\'value\']']
```
veruu: Runer still takes a json, so instead of `'[\'value\']'`, there should be something like `'{"key": "value"}'`so the test is more accurate.

tests/test_executable.py
```diff
@@ -15,6 +15,7 @@
 from StringIO import StringIO
 from io import BytesIO
 import os
+import logging
```
veruu: Can we keep the import order here? Oh wait, there's the `StringIO` import at the start, so nevermind, there's no order :D

veruu:
Also, the test coverage is getting really nice here! Thanks!

tests/test_executable.py
wizardofoz123: I find this obsession with import ordering strange. It's supposed to make it easier to find unused imports or something?  Any decent IDE does that automatically. I'd just separate custom modules, standard libraries and pypi packages.

wizardofoz123:
@veruu Can you merge this now?

veruu:
@wizardofoz123 yup! GitHub doesn't send notifications on pull request changes unless it's a new pushed commit or a comment, so I didn't see them

PR #412, opened 2018-12-11
drop dead code

wizardofoz123:


skt/executable.py
```diff
@@ -1004,27 +1004,3 @@ def check_args(parser, args):
                 'the stdio reporter was selected but arguments for the mail '
                 'reporter were provided'
             )
-
-
-def main():
```
wizardofoz123: Where is this being used? Is it being used? Looks like a dead code to me.

veruu: https://github.com/CKI-project/skt/blob/master/setup.cfg#L33

PR #414, opened 2018-12-18
pep8 fixes

wizardofoz123:
Handle or silence most of skt/skt code pep8 warnings. Now we should see more clearly what parts deserve some kindof cleanup/rewrite, or if it's not worth it.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -21,6 +21,7 @@ import sys
 import time
 import xml.etree.ElementTree as etree
 
+from abc import ABCMeta, abstractmethod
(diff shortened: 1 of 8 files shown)


skt/runner.py
```diff
@@ -249,6 +279,12 @@ def __recipe_set_to_job(self, recipe_set, samehost=False):
         return newroot
 
     def cleanup_handler(self):
+        """
+        Call cancel_pending_jobs() to cancel all pending jobs
+
+        Returns:
+             None
```
veruu: This probably doesn't have to be here, but let's merge it

PR #418, opened 2019-01-16
Pass args to cmd_merge(), not cfg

major:
Pass the command line arguments directly to `cmd_merge` and update
the state file directly without passing around the entire `cfg`
object.

The goal is to treat the state file like a state file and not a
config file. We should be doing just-in-time reads from and writes
to the state file (treating it more like a database).

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -40,6 +40,7 @@ import skt.runner
 from skt.kernelbuilder import KernelBuilder, CommandTimeoutError, ParsingError
 from skt.kerneltree import KernelTree, PatchApplicationError
 from skt.misc import join_with_slash, SKT_SUCCESS, SKT_FAIL, SKT_ERROR
+from skt.state_file import update_state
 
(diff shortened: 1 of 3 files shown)


skt/executable.py
```diff
@@ -97,6 +97,33 @@ def remove_oldresult(output_dir, prefix_filename):
         pass
 
 
+def update_state(state_file, state_dict):
```
veruu: I have actually not checked the changes, but do you want to leverage `skt/state_file.py` for this?

major: @veruu We could. I can move that if it makes more sense.

veruu: Well, we already have that code and it isn't used anywhere, so changing it to fit the functionality we want instead of adding new stuff would be better, IMHO.

major: @veruu  Done! I removed the unused functions/tests, too.

PR #419, opened 2019-01-16
Pass arguments to cmd_build, not cfg

major:
This PR continues the series where we move towards using the state file as a true state file with just-in-time reads and writes.

This PR also includes a new method in `KernelBuilder` that always assembles make options the same way. It can be re-used outside the class as well.


major:
The travis-ci/push failed because I originally pushed this branch to cki-project/skt and then quickly deleted it and pushed it here. :/

skt/executable.py
```diff
@@ -312,72 +312,104 @@ def cmd_merge(args):
...
     tgz = None
     builder = KernelBuilder(
-        source_dir=cfg.get('workdir'),
-        basecfg=cfg.get('baseconfig'),
-        cfgtype=cfg.get('cfgtype'),
-        extra_make_args=cfg.get('makeopts'),
-        enable_debuginfo=cfg.get('enable_debuginfo'),
-        rh_configs_glob=cfg.get('rh_configs_glob'),
-        localversion=cfg.get('localversion')
+        source_dir=full_path(args.get('workdir')),
```
agustinhenze: Actually we are using relative paths on the reporter :smirk: . Maybe you want to take advantage of you're touching this part of the code to simplify the reporter code later :).

skt/state_file.py
```diff
@@ -16,6 +16,39 @@
...
+    # Read the state file
+    config.read(state_file)
+
+    # Check if the state file has a 'state' section.
+    if not config.has_section("state"):
+        return None
+
+    # Get the value from the state file.
+    try:
+        return config.get('state', state_key)
```
agustinhenze: Are you sure this piece of code doesn't catch the previous if?

skt/kernelbuilder.py
```diff
@@ -217,6 +217,15 @@ def getrelease(self):
 
         return krelease
 
+    def assemble_make_options(self):
+        """Assemble all of the make options into a list."""
```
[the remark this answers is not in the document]
major: That's meant to be a list so it can be passed right into `subprocess` methods without modifications.

skt/kernelbuilder.py
```diff
@@ -236,19 +245,15 @@ def mktgz(self, timeout=60 * 60 * 12):
...
-            + self.extra_make_args
-        )
-
-        logging.info("building kernel: %s", kernel_build_argv)
-
         with io.open(self.buildlog, 'wb') as writer, \
                 io.open(self.buildlog, 'rb') as reader:
             self.__prepare_kernel_config(stdout=writer,
                                          stderr=subprocess.STDOUT)
+            # Get the kerel build options.
```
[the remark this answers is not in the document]
major: Indeed. Fixed. ;)

skt/kernelbuilder.py
```diff
@@ -236,19 +245,15 @@ def mktgz(self, timeout=60 * 60 * 12):
...
+            kernel_build_argv = self.assemble_make_options()
+            logging.info("building kernel: %s", kernel_build_argv)
```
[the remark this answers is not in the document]
major: Fixed.

skt/executable.py
major: Valid point. I'll remove it.

skt/state_file.py
major: So if the `[state]` section exists, but the key we are looking for does not, a `NoOptionError` exception is raised. I caught the exception in that `try:` block to make it easier to handle elsewhere in the package.

PR #421, opened 2019-01-17
Refactor tarball searching into its own method

major:
Move the tarball searching logic into its own method, `find_tarball()`.
This allows for more log searching capabilities later, such as searching
for RPM files.

Also, adjust append_and_log2stdout to just write to stdout. This allows
us to read the buildlog instead of storing a large list of log lines
while the build runs.

Add a test for the situation where a kernel was built, it appears in the
log, but the file doesn't actually exist on the filesystem.


skt/kernelbuilder.py
```diff
@@ -284,33 +315,28 @@ def mktgz(self, timeout=60 * 60 * 12):
...
+        # Does the tarball mentioned in the buidlog actually exist on the
+        # filesystem?
         if not os.path.isfile(fpath):
             raise IOError("Built kernel tarball {} not found".format(fpath))
 
         return fpath
 
     @staticmethod
-    def append_and_log2stdout(lines, full_log):
+    def log2stdout(lines):
```
agustinhenze: Maybe we can use the logging module instead :thinking:

PR #422, opened 2019-01-17
Use loggers for file/stdout output

major:
Add a new `run_multipipe` function that uses Python's logging module
to simultaneously write to a log file and stdout. All uses of
check_output/check_call are replaced with the new method.

This patch also includes updated tests.

### DIFF
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -63,6 +63,9 @@ class KernelBuilder(object):
         else:
             self.extra_make_args = []
 
+        # Truncate the buildlog, if it exists.
+        self.__reset_build_log()
+
         logging.info("basecfg: %s", self.basecfg)
         logging.info("cfgtype: %s", self.cfgtype)
 
@@ -170,6 +173,12 @@ class KernelBuilder(object):
             return os.environ['ARCH']
 
         return platform.machine()
+
+    def __reset_build_log(self):
+        """Truncate the build log."""
+        if os.path.isfile(self.buildlog):
+            with open(self.buildlog, 'w') as fileh:
+                fileh.write('')
 
     @classmethod
     def __get_cross_compiler_prefix(cls):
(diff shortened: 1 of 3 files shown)


skt/kernelbuilder.py
```diff
@@ -171,6 +174,12 @@ def __get_build_arch(self):
 
         return platform.machine()
 
+    def __reset_build_log(self):
+        """Truncate the build log."""
+        if os.path.isfile(self.buildlog):
```
agustinhenze:
```python
os.truncate(self.buildlog, 0)
```

skt/kernelbuilder.py
```diff
@@ -338,16 +329,71 @@ def mktgz(self, timeout=60 * 60 * 12):
...
+            bufsize=1
+        )
+
+        # Loop over the buffered stdout/stderr from the command and write the
+        # output to our logger.
+        for line in iter(process.stdout.readline, ''):
+            root.info(line.strip())
+
+        # Ensure the process has exited.
+        process.wait()
```
agustinhenze: Here you should collect the exit code, otherwise you'll get the cached one.
```python
exit_code = process.wait()
```
then return this variable at the bottom of the function.

skt/kernelbuilder.py
```diff
@@ -338,16 +329,71 @@ def mktgz(self, timeout=60 * 60 * 12):
...
+        root.addHandler(file_handler)
+        root.addHandler(stdout_handler)
+
+        # Run the command.
+        logging.debug("Running multipipe command: %s" % ' '.join(args))
+        root.info("$ %s" % ' '.join(args))
+        process = subprocess.Popen(
+            args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
```
agustinhenze: I think you wanted to put
```python
stderr=subprocess.STDOUT
```
here. Otherwise you'll lost the stderr error reading just stdout below.

skt/kernelbuilder.py
major: Ah, no `truncate()` in python2 :(

skt/kernelbuilder.py
major: Fixed.

skt/kernelbuilder.py
major: Thanks for catching that.

PR #423, opened 2019-01-18
Add RPM support, 100% kernelbuilder test coverage

major:
This PR includes several improvements for `kernelbuilder`:

* Refactors kernel compiles to use a common method
* kernelbuilder now supports `targz-pkg` as well as `binrpm-pkg`

Kernel compiles start with `compile_kernel()`. Config is prepared and the kernel is compiled. Upon completion, `handle_tarball()` or `handle_rpm()` can be called to find kernel packages and prepare them for publishing.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -289,7 +289,7 @@ def cmd_build(args):
 
     # Attempt to compile the kernel.
     try:
-        tgz = builder.mktgz()
+        package_path = builder.compile_kernel()
     # Handle a failure if the build times out, fails, or if the build
     # artifacts can't be found.
     except (CommandTimeoutError, subprocess.CalledProcessError, ParsingError,
@@ -310,7 +310,8 @@ def cmd_build(args):
     # Get the SHA of the commit from the repo that we just compiled.
     buildhead = get_state(args['rc'], 'buildhead')
 
-    if tgz:
+    # Handle any built tarballs.
+    if package_path and package_path.endswith('tar.gz'):
         if buildhead:
             # Replace the filename with the SHA of the last commit in the repo.
             ttgz = "{}.tar.gz".format(buildhead)
@@ -319,11 +320,16 @@ def cmd_build(args):
             ttgz = addtstamp(tgz, tstamp)
 
         # Rename the kernel tarball.
-        shutil.move(tgz, ttgz)
+        shutil.move(package_path, ttgz)
         logging.info("tarball path: %s", ttgz)
 
         # Save our tarball path to the state file.
         state = {'tarpkg': ttgz}
+        update_state(args['rc'], state)
+
+    # Handle any RPM repositories.
+    if package_path and 'rpm_repo' in package_path:
+        state = {'rpm_repo': package_path}
         update_state(args['rc'], state)
 
     # Set a filename for the kernel config file based on the SHA of the last
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -44,18 +44,33 @@ class KernelBuilder(object):
         self.make_argv_base = [
             "make", "-C", self.source_dir
         ]
+        self.make_target_args = {
+            'targz-pkg': [
+                "INSTALL_MOD_STRIP=1",
+                "-j%d" % multiprocessing.cpu_count(),
+            ],
+            'binrpm-pkg': [
+                "-j%d" % multiprocessing.cpu_count()
+            ]
+        }
         self.enable_debuginfo = enable_debuginfo
         self.build_arch = self.__get_build_arch()
         self.cross_compiler_prefix = self.__get_cross_compiler_prefix()
         self.rh_configs_glob = rh_configs_glob
         self.localversion = localversion
-        self.make_target = None
 
-        self.targz_pkg_argv = [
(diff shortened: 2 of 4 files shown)


major:
Currently splitting this one into pieces.

major:
This should be good for review now. @agustinhenze? :)

skt/kernelbuilder.py
```diff
@@ -234,11 +249,73 @@ def assemble_make_options(self):
         """Assemble all of the make options into a list."""
         kernel_build_argv = (
             self.make_argv_base
-            + self.targz_pkg_argv
+            + self.compile_args
             + self.extra_make_args
         )
         return kernel_build_argv
 
+    def find_rpm(self):
```
agustinhenze: Find rpm returns an RPM repository :thinking:. Maybe `get_rpm_repo`? Then the docstring could be more explicit I guess as well.

skt/kernelbuilder.py
```diff
@@ -234,11 +249,73 @@ def assemble_make_options(self):
...
+        Args:
+            rpm_files:  A list of RPM paths.
+
+        Returns:
+            Path to the repo directory.
+
+        """
+        # Create a directory for the RPM repo.
+        repo_dir = join_with_slash(self.source_dir, 'rpm_repo/')
+        os.mkdir(repo_dir)
```
agustinhenze: repo_dir must not exist every time you're trying to build rpms using skt. Not saying that it's wrong but is that the behavior we want/expect?

major: I could put a check in there to see if it exists and skip that step if so.

skt/kernelbuilder.py
major: I'll separate the logic there.

skt/kernelbuilder.py
```diff
@@ -234,11 +249,73 @@ def assemble_make_options(self):
...
+
+        # Move our current RPMs to that directory.
+        for rpm_file in rpm_files:
+            shutil.move(rpm_file, repo_dir)
+
+        # Create an RPM repository in the repo directory.
+        args = ["createrepo", repo_dir]
+        self.run_multipipe(args)
```
agustinhenze: If something goes wrong you're ignoring the exit code here :thinking:.

major: Good catch. I'll add a check there.

agustinhenze:
> -1cry CKI pipeline failed [#423 (comment)](https://github.com/CKI-project/skt/pull/423#issuecomment-455677557)
> 
> https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/1532
> https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/1531
> https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/1530
> https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/1529

Very close @cki-bot, but noup. There still pipelines running

skt/executable.py
```diff
@@ -310,7 +310,8 @@ def cmd_build(args):
     # Get the SHA of the commit from the repo that we just compiled.
     buildhead = get_state(args['rc'], 'buildhead')
 
-    if tgz:
+    # Handle any built tarballs.
+    if package_path and package_path.endswith('tar.gz'):
```
agustinhenze: I wonder what happens if package_path is None :stuck_out_tongue:

skt/kernelbuilder.py
```diff
@@ -234,11 +249,80 @@ def assemble_make_options(self):
...
+
+        # Create an RPM repository in the repo directory.
+        args = ["createrepo", repo_dir]
+        exit_code = self.run_multipipe(args)
+
+        if exit_code != 0:
+            raise(
+                IOError,
+                "Unable to create RPM repository at {}".format(repo_dir)
+            )
```
agustinhenze: Maybe
```python
            raise subprocess.CalledProcessError(
                exit_code,
                ' '.join(args)
            )
```
Should be better in this case

skt/kernelbuilder.py
```diff
@@ -317,6 +401,49 @@ def mktgz(self, timeout=60 * 60 * 12):
...
+            package_path = self.handle_tarball()
+
+        if 'rpm' in self.make_target:
+            package_path = self.handle_rpm()
+
+        return package_path
+
+    def handle_rpm(self):
+        """
+        Finds the kernel tarball and verifies that it is present.
```
agustinhenze: the kernel tarball? :thinking:

skt/executable.py
major: The tarball + rpm_repo don't get written to the state file in that case.

skt/kernelbuilder.py
major: That does make more sense. I'll change it.

skt/executable.py
agustinhenze: OK, but do we want to keep going if that happens?

skt/kernelbuilder.py
major: Of... course. I'll fix that.

PR #426, opened 2019-01-18
Only set cross compiler in state if set

major:
Fix a bug where the cross_compiler_prefix was set to `None` in the
state file. It should not be present at all in this situation.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -341,14 +341,19 @@ def cmd_build(args):
     # Gather additional details about the build and save them to the state
     # file.
     kernel_arch = builder.build_arch
-    cross_compiler_prefix = builder.cross_compiler_prefix
     make_opts = builder.assemble_make_options()
     state = {
         'kernel_arch': kernel_arch,
-        'cross_compiler_prefix': cross_compiler_prefix,
         'make_opts': make_opts
     }
     update_state(args['rc'], state)
+
+    # Write the cross compiler prefix to the state file only if the
+    # environment variable is set:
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -341,15 +341,20 @@ def cmd_build(args):
...
-        'cross_compiler_prefix': cross_compiler_prefix,
         'make_opts': make_opts
     }
     update_state(args['rc'], state)
 
+    # Write the cross compiler prefix to the state file only if the
+    # environment variable is set:
+    cross_compiler_prefix = builder.cross_compiler_prefix
+    if cross_compiler_prefix:
+        state = {'cross_compiler_prefix': cross_compiler_prefix}
```
spbnick: Maybe we could put this before the "update_state..." line above, and avoid a second state write, but it will work this way too.

skt/executable.py
```diff
@@ -341,15 +341,20 @@ def cmd_build(args):
     # Gather additional details about the build and save them to the state
     # file.
     kernel_arch = builder.build_arch
-    cross_compiler_prefix = builder.cross_compiler_prefix
```
agustinhenze: Mmm it wouldn't be enough something like this?
```python
     cross_compiler_prefix = builder.cross_compiler_prefix or ''
```

spbnick: Strictly speaking, it won't become `None` after reading, so won't be equivalent, and can result in confusion either in computers or in humans.

PR #428, opened 2019-01-18
Add --make-target argument

major:
This argument allows a user to specify a custom make target. Only
targz-pkg is supported for now.


skt/executable.py
```diff
@@ -719,6 +720,14 @@ def setup_parser():
         type=str,
         help="Additional options to pass to make"
     )
+    parser_build.add_argument(
+        "--make-target",
+        dest="make_target",
+        type=str,
+        default='targz-pkg',
+        choices=('targz-pkg', 'binrpm-pkg'),
```
agustinhenze: The commit says `Only targz-pkg is supported for now.` hehe

major: Oops. :) That's coming in the next patch.

PR #432, opened 2019-01-18
Refactor/simplify publisher; RPM repo support

major:
This PR includes:

- Testing for `skt publish` in CircleCI
- A refactor and simplification of the `Publisher` class
- Support for publishing directories, such as RPM repositories recursively
- Just-in-time state file writes/reads for `cmd_publish()`
- Removal of the unused `sftp` publisher
- Full test coverage for `Publisher` class

### DIFF
--- a/skt/publisher.py
+++ b/skt/publisher.py
@@ -14,36 +14,51 @@
 """Class for managing Publisher."""
 import logging
 import os
-import shutil
 import subprocess
-
-from abc import ABCMeta, abstractmethod
 
 from skt.misc import join_with_slash
 
 
 class Publisher(object):
-    """An abstract result publisher."""
-    __metaclass__ = ABCMeta
+    """Publish artifacts to destinations.."""
 
-    TYPE = 'default'
-
-    def __init__(self, dest, url):
+    def __init__(self, pub_type=None, source=None, dest=None, baseurl=None):
         """
         Initialize an abstract result publisher.
 
         Args:
-            dest:   Type-specific destination string.
-            url:    Base URL prefix of the published result,
-                    without '/' on the end.
+            pub_type:   A type of publisher found in this class.
+            source:     Current location of the artifact to be published.
+            dest:       A destination for the artifact.
+            baseurl:    Base URL of the published result without a
+                        trailing slash.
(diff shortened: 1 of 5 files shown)


skt/publisher.py
```diff
@@ -17,134 +17,101 @@
...
+                os.path.basename(self.source)
+            )
+
+            # copytree will not copy to a directory that exists.
+            if os.path.isdir(destination):
+                shutil.rmtree(destination)
+
+            shutil.copytree(self.source, destination)
+
+    def scp(self):
```
agustinhenze: You need to pass `-r` here for handling folders. Actually, if you add `-r` scp can handle local copies as well which means you can remove `def cp` and just use `scp` for local and remote copies :man_shrugging:

major: Nice find. I've updated the patch.

skt/publisher.py
```diff
@@ -14,137 +14,74 @@
...
-        if cls.TYPE == ptype:
-            return cls(parg, pburl)
-    raise ValueError("Unknown publisher type: %s" % ptype)
+        return join_with_slash(self.baseurl, os.path.basename(self.source))
+
+    def scp(self):
+        """Use scp to copy the source to the destination."""
+        # Create the destination directory.
+        if not os.path.isdir(self.destination):
+            os.makedirs(self.destination)
```
agustinhenze: This doesn't make sense for remote copies :/. Actually `scp` creates the destination if it doesn't exist.

major: I thought so as well, but if I do this:

```
scp -r /tmp/tmpHdZUbM/kernel.tar.gz /tmp/tmpHdZUbM/published/
cp: cannot create regular file '/tmp/tmpHdZUbM/published/': Not a directory
```

There's sadness. ;(

agustinhenze: Oh I see, it creates just depth=1 :smirk:

agustinhenze: Well, after all if you remove makedirs, it will have the same behavior when copy remotely or locally

wizardofoz123:
Publisher is gone, closing.

PR #440, opened 2019-01-29
Fix broken dependency on pycodestyle

agustinhenze:
Remove development dependencies from production creating a new extra option
called [dev].


agustinhenze:
@cki-bot test it now!

agustinhenze:
ok, thanks. @major you can merge it now

PR #443, opened 2019-01-30
add more duct tape

wizardofoz123:

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -655,7 +655,7 @@ class BeakerRunner(Runner):
                 logging.debug(
                     "Got return code when gathering results: %s", ret
                 )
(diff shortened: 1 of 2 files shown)


skt/runner.py
```diff
@@ -662,11 +662,6 @@ def run(self, url, max_aborted, release, wait=False,
                 self.cleanup_handler()
             ret = SKT_ERROR
 
-        if ret == SKT_ERROR:
-            return ret
-        if not wait:
-            return ret
-
```
spbnick: Ha-ha, this is a good one :smile: 
Only, this is rather a "NOOP" or "useless" code, but it does get executed, so it's not "dead".

PR #444, opened 2019-01-30
Revert redis

wizardofoz123:
Drop redis usage in favor of grafana and rename soaking to waiving.


spbnick:
BTW, maybe we don't actually need the `--waiving` option now, BTW.

PR #450, opened 2019-02-07
Send test failures to maintainers

spbnick:


wizardofoz123:
@spbnick: This is sending the test report without any modifications to test maintainers, right?

spbnick:
Not sure what you're exactly asking, but this is supposed to simply send an unmodified copy of the report to maintainers of any failed test.

spbnick:
Thank you for the review, Jakub :) Yes, it would be best to send a separate report to maintainers explaining why they're getting it and highlighting the issue, I agree. That would require more changes in other parts of the pipeline, though, so let's start with this.

I'd like to wait for a pipeline with a test failure and maintainer address in Beaker results, so I could write a reporter test to test this properly before merging. Plus we need a corresponding change to trigger config and retriggers. So I'm keeping this in WIP for now.

spbnick:
OK, I reworked this a bit to support specifying where to add maintainer addresses to (TO, CC, or BCC).
@wizardofoz123, could you take a look again? It should be ready now, and we can merge it if you don't find anything bad.

PR #451, opened 2019-02-13
Show waived tests in reports, plus bonus

spbnick:
Here's my attempt to add waived tests to reports.

They will only appear at the end of the message under the "Tests" heading, like this:

```
Tests
-----

We booted each kernel and ran the following tests:

  x86_64:
      PASSED:         /distribution/kpkginstall
      - URL: https://github.com/CKI-project/tests-beaker/archive/master.zip#distribution/kpkginstall
      WAIVED: FAILED: /a/failed/waived/test
      - URL: https://github.com/CKI-project/tests-beaker/archive/master.zip#a/failed/waived/test
      - Logs:
          http://example.com/foo.log
          http://example.com/bar.log
      WAIVED: PASSED: /a/passed/waived/test
      FAILED:         /a/test/with/no/waive
```

This PR also adds some report cleanup, but we can pick-and-choose that :)


spbnick:
Merging as is and we'll see if anyone complains about the report.

PR #453, opened 2019-02-20
fixup catching panics

wizardofoz123:
There's a missing case in the code. Panic result is always SKT_FAIL,
regardless of status.


agustinhenze:
This is kind of messy... We have to build a dict to map the beaker results and add some tests...

PR #454, opened 2019-03-01
allow permissive makefile target

wizardofoz123:
This is necessary when we want to bypass config checks during special
occasion in per-patch testing.


wizardofoz123:
@cki-bot: Okay, test this now cki-bot! Fetch!

PR #456, opened 2019-03-11
Making ppcs distiguishable

Toaster192:
Adding a simple "(le)" to the name if the ARCH_CONFIG variable ends with "le".

### DIFF
--- a/skt/kernelbuilder.py
+++ b/skt/kernelbuilder.py
@@ -193,8 +193,8 @@ class KernelBuilder(object):
         """Determine the build architecture for the kernel build."""
         # pylint: disable=no-self-use
         # Detect cross-compiling via the ARCH= environment variable
-        if 'ARCH' in os.environ:
-            return os.environ['ARCH']
+        if 'ARCH_CONFIG' in os.environ:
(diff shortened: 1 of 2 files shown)


skt/kernelbuilder.py
```diff
@@ -194,6 +194,9 @@ def __get_build_arch(self):
         # pylint: disable=no-self-use
         # Detect cross-compiling via the ARCH= environment variable
         if 'ARCH' in os.environ:
+            if ('ARCH_CONFIG' in os.environ and
+                    os.environ['ARCH_CONFIG'][-2:] == "le"):
+                return os.environ['ARCH'] + "(le)"
```
agustinhenze: what about returning always `ARCH_CONFIG`?

Toaster192: Yeah that sounds like a cleaner solution, given ARCH_CONFIG should be set everytime ARCH is

The names will just be slightly less readible (but more informative)

veruu:
Changes are looking OK to me (cc @agustinhenze in case he has more feedback), but can we get rid of the merge commit please?

PR #458, opened 2019-03-25
Add tag to krelease variable (reporter headline)

Toaster192:
Adding more information about the state before the merges to krelease variable.

The tag is also added to the state file now (could be used by itself or to 'restore' the old krelease for whatever reason)

The krelease now looks like this: `3.10.0-1030.el7.skt` instead of `3.10.0.skt`
(tag contains either the 'r' from 'n-v-r' in the RHEL tag or the first 7 characters of the commit hash for upstream)

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -34,7 +34,7 @@ import skt.reporter
 import skt.runner
 from skt.kernelbuilder import KernelBuilder, CommandTimeoutError, ParsingError
 from skt.kerneltree import KernelTree, PatchApplicationError
-from skt.misc import join_with_slash, SKT_SUCCESS, SKT_FAIL
+from skt.misc import join_with_slash, SKT_SUCCESS, SKT_FAIL, get_tag
 from skt.state_file import get_state, update_state
 
 DEFAULTRC = "~/.sktrc"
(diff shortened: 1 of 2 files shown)


Toaster192:
@veruu

veruu:
@cki-bot can you test this please? Up till Beaker submission should be enough so we can check the whiteboard

Toaster192:
Also it's probably arguable what format  hould the krelease with tag be

The krelease is `3.10.0.1030.el7.skt` while the commit msg uses `kernel-3.10.0-1030.el7`

veruu:
@Toaster192 the second one (with the `.skt` added at the end) would be ideal to follow the N-V-R format

Toaster192:
Yeah I agree

Toaster192:
Should now work with every LOCALVERSION (except maybe those that contain '.')

veruu:
@cki-bot can you test the new changes again please? :)

Toaster192:
Also the format looks like this now `3.10.0-1030.el7.foo`

Toaster192:
Um, maybe put the dash symbol into the krelease + tag apendation so the tag in config doesn't start with it?

PR #463, opened 2019-04-05
Add release tag to localversion

Toaster192:
Adding the tag to config with localversion so that 'make kernelrelease' now has the tag as well

The krelease variable is taken from the end of the build which prints the same value so there's no longer a need to add it later.


veruu:
@cki-bot test this out please!

PR #467, opened 2019-05-31
Rewrite simplify py3

wizardofoz123:
This is a continuation of an effort to remove skt from test stage.

This patchset drops unused code and most of related documentation and ports the project to Python3.
After this is merged, the project will be simplified further. But for now, we're happy with determining what code we actually need or not.


wizardofoz123:
@cki-bot test this

wizardofoz123:
This needs to be tested with py3 in pipeline...

skt/executable.py
```diff
@@ -13,86 +13,27 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from __future__ import print_function
```
veruu: Is this line really needed if we want to run with py3 only?

tests/assets/testing_state.cfg
```diff
@@ -3,10 +3,6 @@ foo=bar
 
 [state]
 jobid_01=J:123456
-mergerepo_01=git://git.kernel.org/pub/scm/linux/kernel/git/
-mergehead_01=master
-localpatch_01=/tmp/patch.txt
-patchwork_01=http://example.com/patch/1
```
veruu: Maybe keep some of the data in the testing files and verify they haven't been changed in a test? We use a state file with a ton of data skt runner shouldn't be touching so this could be useful. Not sure though what your next steps are though so this might not be needed based on that.

skt/executable.py
```diff
@@ -1026,61 +311,25 @@ def load_config(args):
...
+        cfg = load_config(args)
+        retcode = args.func(cfg)
 
         sys.exit(retcode)
     except KeyboardInterrupt:
         print("\nExited at user request.")
         sys.exit(130)
+
+
+if __name__ == '__main__':
```
veruu: Is this addition needed since the main function is set as entry point in the installation setup?

wizardofoz123: I can remove this with next respin.

PR #468, opened 2019-06-03
fixup ewd handling

wizardofoz123:
This patch adjusts a simple bit of code that checks for EWD hit on boot
test. If this happens, then it is suspicious and likely a kernel
problem.
The adjustment just moves the code to make sure it is checked both when
the task fails and (by error) succceeds, even though EWD was hit.


veruu:
@wizardofoz123 the aborted job handler is called from https://github.com/CKI-project/skt/blob/master/skt/runner.py#L523 so making a change in `decide_run_result_by_task` doesn't help :(

PR #479, opened 2019-07-24
Return different code if booting failed

veruu:
The warn/aborted task result is the same as for infrastructure failures.
Instead of having to go through all the pain of differentiating between
failed boot and general aborted tasks in the reporter, we should simply
pass this information clearly from skt as it already does this check.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -101,7 +101,6 @@ def cmd_run(config_set):
 
     config_set['jobs'] = ' '.join(runner.job_to_recipe_set_map.keys())
 
-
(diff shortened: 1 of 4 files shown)


major:
This looks okay but the linter is displeased:

```
./skt/executable.py:105:5: E303 too many blank lines (2)
    save_state(config_set, {'retcode': retcode})
    ^
```

veruu:
Hmm that doesn't look like my change, but I can push a different commit fixing it tomorrow

veruu:
@major fixed!

PR #482, opened 2019-08-09
fix status/result typo

wizardofoz123:
There is no such result as 'Aborted', but there is status 'Aborted'.


wizardofoz123:
Do not merge yet. Because we check first recipe only , this could make things worse.

wizardofoz123:
Fixed  in !486

PR #483, opened 2019-08-12
Cleanup

wizardofoz123:
Move towards shorter, more readable and better documented skt code.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -26,6 +26,7 @@ from defusedxml.ElementTree import tostring
 
 from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR, SKT_BOOT
 from skt.misc import WaivingWrap
+from skt.misc import shell_run
 
 
 class BeakerRunner:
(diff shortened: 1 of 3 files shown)


skt/runner.py
```diff
@@ -596,33 +597,39 @@ def get_kpkginstall_task(cls, recipe_node):
...
-        args = ["bkr", "job-submit"]
-
-        if self.jobowner is not None:
-            args += ["--job-owner=%s" % self.jobowner]
-
-        args += ["-"]
-
-        bkr = subprocess.Popen(args, stdin=subprocess.PIPE,
-                               stdout=subprocess.PIPE)
+        owner = "--job-owner=%s" % self.jobowner if self.jobowner else ''
```
Toaster192: might want to use the 'new style' of formating here

Toaster192: especialy since it's used right below this (so it would be more consistent)

wizardofoz123:
Coveralls in unavailable but the PR is fine. Merging.

PR #486, opened 2019-08-12
improve retcode checking

wizardofoz123:
In order to prepare skt for multirecipe/recipeset handling, we should
improve retcode checking.
This patch should mean no functional changes, but should make things
easier to read and manage.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -25,7 +25,76 @@ from defusedxml.ElementTree import fromstring
 from defusedxml.ElementTree import tostring
 
 from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR, SKT_BOOT
-from skt.misc import WaivingWrap
+from skt.misc import is_task_waived
+
+
+class ConditionCheck:
+    def __init__(self, retval, **kwargs):
+        self.retval = retval
+        self.kwargs = kwargs
+
+    def __str__(self):
+        values = ' '.join([f'{arg}={self.kwargs[arg]}' for arg in self.kwargs])
+
+        return f'retval={self.retval} {values}'
+
+    def __call__(self, task, waiving, is_task_waived_func, prev_task):
+        """ Evaluates the condition and return retval if matched, else None.
+
+            Args:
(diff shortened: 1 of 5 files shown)


skt/runner.py
```diff
@@ -28,6 +28,70 @@
...
+
+result_condition_checks = [
+    # This contains objects that will return <retval> (first parameter), when
+    # all the specified conditions are met. Empty conditions with no keywords
+    # are never met.
+
+    # Previous task was waived and panicked, which causes the next
+    # task to abort. The task is waived for a reason, return
+    # SKT_SUCCESS.
+    ConditionCheck(SKT_SUCCESS, result='Fail', waived=False, status='Aborted',
```
veruu: Shouldn't this be `result=Warn` ?

skt/runner.py
```diff
@@ -291,12 +331,11 @@ def __getresults(self):
...
-                    if recipe_result.attrib.get('result') != 'Pass':
-                        ret = self.decide_run_result_by_task(recipe_result)
+                    ret = self.decide_run_result_by_task(recipe_result)
 
-                        if ret != SKT_SUCCESS:
-                            logging.info('Failure in a recipe detected!')
-                            return ret
+                    if ret != SKT_SUCCESS:
+                        logging.info('Failure in a recipe detected!')
+                        return ret
```
inakimalerba: This wouldn't change the issue we were discussing last friday on IRC about returning the first error that was found on the recipes.

Maybe errors should be saved and return the .. highest value? .

veruu: @inakimalerba it does not -- the PR just tries to simplify retcode checking here so the fix for the issue you mention is easier to manage

skt/runner.py
wizardofoz123: I think so. The original code would actually allow warn/fail/panic in this check...

wizardofoz123:
@cki-bot test

wizardofoz123:
@cki-bot test

wizardofoz123:
The pipelines are OK. Can we merge this @veruu ?

PR #487, opened 2019-08-12
dont add to blacklist when hostname is forced

wizardofoz123:
When force attribute of hostRequires element is used, it makes no sense
to add more hosts or conditions to blacklist.


wizardofoz123:
Discussed on irc.

PR #488, opened 2019-08-16
fix multirecipe handling

wizardofoz123:
SKT currently decides its retcode based on the first recipe it checks.
If there are any failures, it may prefer SKT_FAILURE over SKT_ERROR or
vice-versa.
This patch attempts to fix that by getting all the individual retcodes
from recipes and deciding result by the presence of a priority return
code.
The log message is modified to include recipeid that hit the priority
check.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -321,18 +321,26 @@ class BeakerRunner:
 
     def __getresults(self):
         """
-        Get return code based on the job results.
+        Get return code based on the job results. This processes all recipes.
+        The priority is (from highest to lowest):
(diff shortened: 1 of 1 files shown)


wizardofoz123:
@cki-bot test [skip rhel7][skip upstream][skip brew][test beaker]

veruu:
I think the logic makes sense, but let's wait and check bot results before merging

wizardofoz123:
Since we've discussed this with @veruu  and pipelines are actually OK, I'm merging this.

PR #489, opened 2019-08-21
don't run bkr job-cancel on completed passed jobs

wizardofoz123:
When runner retcode is 0, don't run cleanup handler to cancel job.
This isn't really necessary, but I find the double 'Cancelling pending
jobs!' at the end disconcerting.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -143,6 +143,9 @@ class BeakerRunner:
         # if True, keep waived tests hidden for this run
         self.waiving = None
 
+        # retcode used here only to decided if cleanup handler should be run
+        self.retcode = None
+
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -106,6 +106,8 @@ def cmd_run(config_set):
 
     save_state(config_set, {'retcode': retcode})
 
+    runner.retcode = retcode
+
```
spbnick: Why do we have to assign the runner's retcode in `executable.py`, when only the runner itself influences it and cares about it? Can't we assign it in `runner.run()` and thus avoid making `executable.py` know about runner's internals?

wizardofoz123: yeah, this was rushed and hackish, sorry

PR #490, opened 2019-08-23
make sure save_state() is called on exit

wizardofoz123:
In our pipeline, skt is being run with timeout utility to kill it after a
set timeout. This is used to timeout on infrastructure issues, but we
currently don't write out skt state to rc-file, so we don't get even a
partial report for that given arch.

This patch makes skt always call save_state_on_exit() inside
cleanup_handler(), unless it was already called. If the run isn't finished,
this will result in SKT_ERROR retcode being set and the rest of the jobs
being cancelled.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -18,7 +18,6 @@ import os
 import platform
 import re
 import subprocess
-import sys
 import time
 
 from defusedxml.ElementTree import fromstring
@@ -138,14 +137,11 @@ class BeakerRunner:
         # Set up the default, allowing for overrides with each run
         self.max_aborted = 3
 
-        # determines if termination cleanup was done and all jobs terminated
-        self.cleanup_done = False
-
         # if True, keep waived tests hidden for this run
         self.waiving = None
 
-        # retcode used here only to decided if cleanup handler should be run
-        self.retcode = None
+        # the actual retcode to return is stored here
+        self.retcode = SKT_ERROR
 
         logging.info("runner type: %s", self.TYPE)
         logging.info("beaker template: %s", self.template)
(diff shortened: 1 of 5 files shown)


wizardofoz123:
@cki-bot test

wizardofoz123:
@spbnick I rewrote the PR

spbnick:
Thank you, Jakub. This is a little better, but could you just keep the configuration management simple, not make any wrapper classes, and passing any callbacks, but just move signal handling to the executable, and call runner's cleanup there, followed by all the necessary state saving, while keeping the runner unaware of it?

wizardofoz123:
@spbnick Hey Nick, I think the configuration management is simple and clean enough now. There may be other ways how to make cleanup handler do all it should without creating the wrapper, etc., but I'm not interested in working on this further, because I don't see good reasons for further changes. If you insist on changing this, could you please take over? I'd be curious to see what you had in mind and how would you solve this. :) Thanks.

skt/runner.py
```diff
@@ -137,14 +136,14 @@ def __init__(self, jobtemplate, jobowner=None, blacklist=None):
...
         # if True, keep waived tests hidden for this run
         self.waiving = None
 
-        # retcode used here only to decided if cleanup handler should be run
-        self.retcode = None
+        # the actual retcode to return is stored here
+        self.retcode = SKT_ERROR
+
+        # if True, cleanup handler was run for this runner
+        self.cleanup_done = False
```
spbnick: Since the runner has nothing to do with cleanup now, please remove `cleanup_done` from it, and move it to the executable completely.

wizardofoz123: Any thoughts on where to put that variable so it can be accessed from the test without making it a global variable?

spbnick: I would instead base the test on something else. Like the update of the configuration file.

spbnick:
Looks good! Thank you, Jakub!

PR #495, opened 2019-09-19
retry bkr job-submit on specific stderr data

wizardofoz123:
This patch adds a retry-decorator and uses it to create
retry_safe_popen. Beaker job submission is retried, if stderr stream
contains any of the specificied strings.

See ticket 1127.

### DIFF
--- a/skt/misc.py
+++ b/skt/misc.py
@@ -15,6 +15,7 @@
 import logging
 import subprocess
 
+from skt.retrying import retrying_on_exception
 # SKT Result
 SKT_SUCCESS = 0
 SKT_FAIL = 1
@@ -22,6 +23,40 @@ SKT_ERROR = 2
 SKT_BOOT = 3
 
 LOGGER = logging.getLogger()
+
+
+@retrying_on_exception(RuntimeError)
+def retry_safe_popen(err_exc_strings, *args, stdin_data=None, **kwargs):
+    """ Call safe_popen with *args, stdin_data=None, **kwargs provided,
+        If stderr stream is present and contains any string in err_exc_strings
+        list, then the process call is done again with retry after 3 seconds
+        (see retrying_on_exception decorator). Log commands retry and allow 3
+        retries max. Also log if last command failed and we gave up. The
+        program execution is not terminated / no exception is raised on last
+        failure.
+
+        Args:
+            err_exc_strings: a list of strings; if any is present in stderr,
+                             retry the command
+            args:            arguments to pass to Popen
+            stdin_data:      None or str, use None when you don't want to pass
+                             string data to stdin
+            kwargs:          keyword arguments to pass to Popen
+        Returns:
+            tuple (stdout, stderr, returncode) where
+                stdout is a string
+                stderr is a string
+                returncode is an integer
+    """
+    stdout, stderr, returncode = safe_popen(*args, stdin_data=stdin_data,
+                                            **kwargs)
+
+    for err_str in err_exc_strings:
+        if stderr and err_str in stderr:
+            logging.warning(stderr.strip())
+            raise RuntimeError
+
+    return stdout, stderr, returncode
 
 
 def safe_popen(*args, stdin_data=None, **kwargs):
--- a/skt/misc.py
+++ b/skt/misc.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2018 Red Hat, Inc. All rights reserved. This copyrighted
+# Copyright (c) 2017-2019 Red Hat, Inc. All rights reserved. This copyrighted
 # material is made available to anyone wishing to use, modify, copy, or
 # redistribute it subject to the terms and conditions of the GNU General
 # Public License v.2 or later.
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2017 Red Hat, Inc. All rights reserved. This copyrighted
+# Copyright (c) 2017-2019 Red Hat, Inc. All rights reserved. This copyrighted
 # material is made available to anyone wishing to use, modify, copy, or
 # redistribute it subject to the terms and conditions of the GNU General
 # Public License v.2 or later.
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -25,7 +25,7 @@ from defusedxml.ElementTree import tostring
 
 from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR, SKT_BOOT
 from skt.misc import is_task_waived
-from skt.misc import safe_popen
+from skt.misc import safe_popen, retry_safe_popen
 
 
 class ConditionCheck:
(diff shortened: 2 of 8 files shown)


wizardofoz123:
Hey Nick, this is the initial work for ticket 1127. We may need to tweak delays and error strings that are caught. Initial thoughts?

skt/retry.py
```diff
@@ -0,0 +1,51 @@
...
+                except exception as exc:
+                    wrapped.failed_count += 1
+
+                    err_msg = str(exc) + ", delaying for %ds" % delay
+                    logging.warning(err_msg)
+                    for arg in args:
+                        logging.warning(str(arg))
+
+                    # don't sleep on last attempt, there's no point
+                    if i != retries - 1:
```
emanchado: Nitpick: maybe we can compare `wrapped.failed_count` instead of `i`, and that way I think we'll skip the `- 1`. Also maybe the intention becomes clearer?

skt/retry.py
```diff
@@ -0,0 +1,51 @@
...
+                        time.sleep(delay)
+                    else:
+                        # reraise exc. instead
+                        raise exc
```
emanchado: I don't understand this: why "instead"? Instead of what?

Also, maybe related to my confusion: isn't it better to just use `raise` without parameters, to keep stack information and such?

skt/misc.py
```diff
@@ -24,6 +25,44 @@
...
+                returncode is an integer
+    """
+    stdout, stderr, returncode = safe_popen(*args, stdin_data=stdin_data,
+                                            **kwargs)
+    for err_str in err_exc_strings:
+        if stderr and err_str in stderr:
+            logging.warning(stderr)
+            safe_popen_args = ' '.join(list(*args))
+            logging.warning('RETRY shell command "%s"', safe_popen_args)
+            if not (retry_safe_popen.failed_count + 1 ==
```
emanchado: Isn't this a bit clearer as `if retry_safe_popen.failed_count + 1 < retry_safe_popen.retries`?

And maybe add a comment stating that `.failed_count` doesn't count the current error, hence the need for the `+ 1`. It took me five minutes of back and forth to figure it out, I almost write that there was a bug in the code 😅

skt/runner.py
```diff
@@ -649,9 +649,11 @@ def __jobsubmit(self, xml):
             args += ["--job-owner=%s" % self.jobowner]
 
         args += ["-"]
-
-        stdout, _, _ = safe_popen(args, stdin_data=xml, stdin=subprocess.PIPE,
-                                  stdout=subprocess.PIPE)
+        err_strings = ['connection to beaker.engineering.redhat.com failed',
```
emanchado: Nitpick: Maybe use `"` for this string, too, so it looks like the other one?

skt/retry.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2018 Red Hat, Inc. All rights reserved. This copyrighted
```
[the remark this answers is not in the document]
wizardofoz123: Well, there should be no inconsistencies. :D I was planning to change all old dates with subsequent patch.

skt/retry.py
```diff
@@ -0,0 +1,51 @@
...
+def retry(exception, retries=3, initial_delay=3):
+    """Decorator to retry method/function that raised exception after
+       initial_delay seconds. Do retries based on how many were set."""
+    def wrapper(function):
+        def wrapped(*args, **kwargs):
+            wrapped.failed_count = 0
+            wrapped.retries = retries
+
+            delay = 0
+            for i in range(0, retries):
```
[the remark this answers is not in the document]
wizardofoz123: It doesn't matter and range(0, retries) is actually more explicit.

skt/retry.py
```diff
@@ -0,0 +1,51 @@
...
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Retry decorator to retry method/function several times."""
+
+import time
+import logging
+
+
+def retry(exception, retries=3, initial_delay=3):
+    """Decorator to retry method/function that raised exception after
+       initial_delay seconds. Do retries based on how many were set."""
```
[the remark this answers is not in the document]
wizardofoz123: Will fix this.

skt/retry.py
wizardofoz123: sure

skt/retry.py
wizardofoz123: 1) Look one or two lines above. :D Well, ideally it would be:
`
                    if i != retries - 1:
                        s = str(e) + ", delaying for %ds" % delay
                        logging.warning(s)
                        time.sleep(delay)
                    else:
                        # don't sleep on last attempt, there's no point;
                        # instead reraise exception
                        raise
`
Meaning "retry" is the most common case (print warning + wait), but on last attempt we reraise the exception instead(of waiting). Makes sense?

2) Yes, no doubt.

skt/runner.py
wizardofoz123: sure

skt/retry.py
```diff
@@ -0,0 +1,61 @@
...
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+"""Retry decorator to retry method/function several times."""
+
+import time
+import logging
+
+
+def retry(exception, retries=3, initial_delay=3):
```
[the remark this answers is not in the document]
emanchado: I agree `retried` is probably better than `retry`, but it still sounds a bit weird. Like it has already done it. I wonder if `retrying` would be a better name? As in, it turns a regular function into a "retrying function", a function which retries?

I also thought of `retryable`, but that sounds like you have to do something to make it retry, as opposed to "it behaves like that now".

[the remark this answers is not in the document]
spbnick: Yeah, maybe `retrying_on` could work better. However, I would avoid the complexity of this particular solution and forgo the decorator altogether, if possible.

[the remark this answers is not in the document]
wizardofoz123: I'll use retryable, as in it doesn't have to be retryed always, it can succeed on the 1st attempt.

wizardofoz123:
Hey Nick, I've changed some parts, but also decided to keep some parts the same, even though we've discussed them. Let me know what you think.

skt/misc.py
```diff
@@ -24,6 +25,39 @@
 LOGGER = logging.getLogger()
 
 
+@retryable(RuntimeError)
```
spbnick: I'm still not entirely happy with the name: it implies that this function could be retried, but so could any function. I would rather call it `retrying`, or maybe `retrying_on_exception`.

skt/misc.py
```diff
@@ -24,6 +25,39 @@
...
+def retry_safe_popen(err_exc_strings, *args, stdin_data=None, **kwargs):
+    """ Call safe_popen with *args, stdin_data=None, **kwargs provided,
+        If stderr stream is present and contains any string in err_exc_strings
+        list, then the process call is done again with retry after 3 seconds
+        (see retryable decorator). Log commands retry and allow 3 retries max.
+        Also log if last command failed and we gave up. The program execution
+        is not terminated / no exception is raised on last failure.
```
spbnick: This is a rather lengthy and detailed explanation. Would be great if you could begin with a single sentence giving a short and simple summary. Something like "Execute a program, retrying up to three times if certain error output is detected." Then have the detailed explanation as a separate paragraph or better two.

PR #498, opened 2019-10-02
Save a recipe set list in state file

veruu:
This will be easier to parse and merge than the indices.

### DIFF
--- a/skt/executable.py
+++ b/skt/executable.py
@@ -100,14 +100,17 @@ def cmd_run(config_set):
             return
 
         recipe_set_index = 0
+        recipe_set_list = []
         for index, job in enumerate(runner.job_to_recipe_set_map.keys()):
             save_state(config_set, {'jobid_%s' % (index): job})
             for recipe_set in runner.job_to_recipe_set_map[job]:
+                recipe_set_list.append(recipe_set)
                 save_state(config_set,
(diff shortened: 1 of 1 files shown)


skt/executable.py
```diff
@@ -100,14 +100,17 @@ def cleanup_handler():
...
         for index, job in enumerate(runner.job_to_recipe_set_map.keys()):
             save_state(config_set, {'jobid_%s' % (index): job})
             for recipe_set in runner.job_to_recipe_set_map[job]:
+                recipe_set_list.append(recipe_set)
                 save_state(config_set,
                            {'recipesetid_%s' % (recipe_set_index): recipe_set})
                 recipe_set_index += 1
 
         config_set['jobs'] = ' '.join(runner.job_to_recipe_set_map.keys())
+        config_set['recipesets'] = ' '.join(recipe_set_list)
```
emanchado: I know this is already merged, but: even inside `config_set`, it seems that we're writing `recipe_set` and not `recipeset` (eg. `max_recipe_set_index` right below).

Would it make sense to call this `recipe_sets` instead?

veruu: For the saved value, I wanted to be consistent with the indexed values which are `recipesetid_X` where it's without any underscore

PR #499, opened 2019-10-15
Don't save jobid_X keys

veruu:
On the rare occasion we need the data we should use the "jobs" key
which is easier to write/read.


veruu:
Has to be merged after https://gitlab.com/cki-project/pipeline-definition/merge_requests/233 as we're removing the jobid_X statefile merging there.

veruu:
All the currently running pipelines are already in the test stage so this should be safe to merge as well.

PR #502, opened 2019-11-13
Use cki lib

wizardofoz123:


emanchado:
Not exactly about this change, but I wonder if `safe_popen` and `retry_safe_popen` should be in a module called `misc`. Maybe once inside `cki_lib` it should have a better module name? Maybe `popen` or something similar?

wizardofoz123:
Sure, the rename would make sense there. Though I wouldn't mind waiting until the lib inflates a bit.

PR #504, opened 2019-11-21
blacklist: sanitize input

wizardofoz123:
We've had a very nasty case, where SKT broke a pipeline, because the
blacklist file generated from beaker-metrics was invalid (containing
something xml parser would not parse ).
This patch makes sure that doesn't happen again by catching exceptions
when adding blacklisted hosts.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -22,6 +22,7 @@ import time
 
 from defusedxml.ElementTree import fromstring
 from defusedxml.ElementTree import tostring
+from defusedxml.ElementTree import ParseError
 
 from skt.misc import SKT_SUCCESS, SKT_FAIL, SKT_ERROR, SKT_BOOT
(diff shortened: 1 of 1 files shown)


skt/runner.py
```diff
@@ -383,8 +384,12 @@ def __blacklist_hreq(self, host_requires):
...
 
         for disabled in self.blacklisted:
-            hostname = fromstring(f'<hostname op="!=" value="{disabled}" />')
-            and_node.append(hostname)
+            try:
+                hostname = fromstring(f'<hostname op="!=" value="{disabled}" '
+                                      f'/>')
+                and_node.append(hostname)
+            except ParseError:
+                logging.info('Skipping invalid blacklist entry!')
```
spbnick: Hmm, why don't you quote it properly instead?

PR #505, opened 2019-11-28
Don't forget aborted jobs if the kernel installed

veruu:
If a test aborts, the results of previous tests are still useful.
Forgetting all aborted jobs meant we never reported these jobs. Only
forget the jobs if we didn't even get past kernel installation.


veruu:
@cki-bot test

PR #506, opened 2019-12-02
don't spam log with "Invalid entry" messages

wizardofoz123:


spbnick:
The change is trivial, but what's the context? Why do we get many of these messages, and how does that help?

wizardofoz123:
Recently, an invalid blacklist was generated (because the old non working code is still in beaker-metrics). The code I've added prevented SKT from crashing, but because it processes each line, there's a warning message for each line (because each line is normally a blacklisted host).

PR #507, opened 2020-01-17
use cki-lib datadefinition

wizardofoz123:
SKT has had issue with its separation of config file and output for a
while. Because it's function is now limited to watching Beaker Jobs,
it's not critical. However, it is a nice example of how the data
definition could be defined and shared.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2017-2019 Red Hat, Inc. All rights reserved. This copyrighted
+# Copyright (c) 2017-2020 Red Hat, Inc. All rights reserved. This copyrighted
 # material is made available to anyone wishing to use, modify, copy, or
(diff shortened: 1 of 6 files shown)


veruu:
:+1: for the plan. I didn't look at it too deeply but right now I don't see anything concerning and I like the direction this is going in.

It also looks extensible enough which is awesome - since we want to make skt the only thing interacting with Beaker, we can eventually extend the datastructure to include closer task results. We can already start with the `sum_failed/aborted` and then go even further. I'm getting ahead of myself, I know.

wizardofoz123:
@cki-bot test

wizardofoz123:
@cki-bot test

wizardofoz123:
@cki-bot test

veruu:
Code changes look fine. Let's wait for the bot to verify everything is ok.

wizardofoz123:
As far as I can see, the pipeline failures are testing issues. *holds breath* Merging!

PR #509, opened 2020-02-13
Remove waiving option and type

veruu:


veruu:
Can only be tested and merged after https://gitlab.com/cki-project/pipeline-definition/-/merge_requests/494 is integrated and pipelines with that change are happily running.

wizardofoz123:
I've rebased this on top of pending cki-lib datadefiniton PR: https://github.com/wizardofoz123/skt/tree/runner

skt/executable.py
```diff
@@ -45,7 +45,7 @@ def save_state(config_set, state):
     """
 
     def set_field(key, val, dst):
-        if key in ['type', 'jobtemplate', 'jobowner', 'blacklist']:
+        if key in ['jobtemplate', 'jobowner', 'blacklist']:
```
emanchado: Super nitpick, mostly for the future, not necessarily for this PR: it's more Pythonic to use tuples in this case (throwaway, literal collection) instead of lists.

skt/executable.py
```diff
@@ -164,9 +163,6 @@ def setup_parser():
     parser.add_argument("-v", "--verbose", help="Increase verbosity level",
                         action="count", default=0)
     parser.add_argument("--rc", help="Path to rc file", required=True)
-    parser.add_argument("--waiving", help=("Hide waived tests."),
-                        type=lambda x: (str(x).lower() == 'true'),
-                        default=True)
     # FIXME Storing state in config file can break the whole system in case
```
emanchado: Question: is there any piece of software that could be using this option? If so, maybe it would be a better idea to accept the option, just make it a no-op?

skt/runner.py
```diff
@@ -53,11 +52,12 @@ def __call__(self, task, waiving, is_task_waived_func, prev_task):
         task_results = {
             'result': task.attrib.get('result'),
             'status': task.attrib.get('status'),
-            'waived': True if waiving and is_task_waived_func(task) else False,
+            'waived': True if is_task_waived_func(task) else False,
```
emanchado: Question: do we want to _not_ assume that the function `it_task_waived_func` will always return `True` or `False`? Because otherwise it seems cleaner to say:
```
'waived': is_task_waived_func(task)
```
I have checked the code and the only function currently used as `is_task_waived_func` is `is_task_waived` (defined in `misc.py`), and that function always returns `True` or `False`, so...

tests/test_runner.py
```diff
@@ -443,18 +443,6 @@ def test_run_wait(self, mock_jobsubmit, mock_getresultstree):
...
-    def test_run_wait2(self, mock_logging, mock_logging_err, mock_jobsubmit):
-        """Ensure BeakerRunner.run works."""
-        # pylint: disable=W0613
-
-        # beaker_results.xml has test failures, returncode should be SKT_FAIL
-        result = misc.exec_on(self.myrunner, mock_jobsubmit,
-                              'beaker_results.xml', 1, 'Completed')
-        self.assertEqual(SKT_FAIL, result)
-
     @mock.patch('logging.warning')
```
emanchado: Why was this test removed? Is it about type and/or waiving?

skt/executable.py
veruu: I didn't want to include unrelated changes in the PR :)

skt/executable.py
veruu: https://gitlab.com/cki-project/pipeline-definition/-/commit/63424d0e0caa71c7c0d216d864640118311b4e8b that change was already merged

tests/test_runner.py
veruu: Because it was verifying the functionality with waiving=False, which is what was removed in the PR. Tests for waiving=True are still present

skt/runner.py
veruu: Good point! I'll refresh the PR later (together with the rebase, if Jakub's PR will be merged while I return)

skt/executable.py
veruu: Whole function dropped by jracek

skt/runner.py
veruu: Fixed.

veruu:
@cki-bot please test this again just in case!

wizardofoz123:
The tests looks OK, just 1 infra issue.

PR #512, opened 2020-03-05
improve getresultstree behavior

wizardofoz123:

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -226,12 +226,7 @@ class BeakerRunner:
             logging.warning(stderr)
             raise RuntimeError('failed getting Beaker job-results')
 
-        # Write the Beaker results locally so they could be stored as an
-        # artifact.
-        results_filename = 'beaker-results-{}.xml'.format(taskspec)
-        with open(results_filename, 'w') as fileh:
-            fileh.write(stdout)
-
+        # return Beaker results parsed xml
         return fromstring(stdout)
 
     def __forget_taskspec(self, recipe_set_id):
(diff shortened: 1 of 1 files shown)


skt/runner.py
```diff
@@ -217,9 +217,14 @@ def getresultstree(cls, taskspec):
...
-                                        stderr=subprocess.PIPE,
-                                        stdout=subprocess.PIPE)
+        stdout, stderr, returncode = retry_safe_popen(err_strings, args,
+                                                      stderr=subprocess.PIPE,
+                                                      stdout=subprocess.PIPE)
+
+        if returncode:
+            logging.warning(stdout)
+            logging.warning(stderr)
+            raise RuntimeError('failed getting Beaker job-results')
```
veruu: Can you explain the reasoning for the exception in the commit message please? It only mentions the added logging but not why we should immediately abort if something goes wrong so the functionality changes don't match.

wizardofoz123: I've expanded the description in the commit message. Thanks. :)

veruu: Thanks, that makes more sense now! And creates a followup question: do we also want to remove the beaker-results xml creation form here, since we've 1) removed them as artifacts and 2) even when they were artifacted we redownloaded the results from beaker in the pipeline?

Feel free to merge this PR if you don't want to make that followup cleanup commit.

wizardofoz123: Nice catch, that file isn't used anymore. So let's not keep it.

PR #513, opened 2020-03-09
Fix bugs

wizardofoz123:


wizardofoz123:
@cki-bot test

wizardofoz123:
The issues aren't related to this change, it's the kernel tests failing + infra issues. 
With this patch, SKT is closer to what was intended. All recipes are evaluated....

@veruu: now it's up to you and your review.

tests/test_runner.py
```diff
@@ -497,7 +497,7 @@ def test_run_wait7(self, mock_logging, mock_logging_err, mock_jobsubmit):
         # abort later on, change last recipe to Aborted
         result = misc.exec_on(self.myrunner, mock_jobsubmit,
                               'beaker_aborted_some.xml', 2, 'Aborted')
-        self.assertEqual(SKT_ERROR, result)
+        self.assertEqual(SKT_FAIL, result)
```
veruu: The comment above says `abort later on, change last recipe to Aborted`, can you also adjust that to explain why we report failure here? Since the max_aborted is overrident to 2 and we have 2 aborted recipes there, the error result sounds valid to me so I'm not sure what I'm missing here

wizardofoz123: A small tweak was needed about when to evaluate .has_aborted.

I've debugged this and there's a slightly problem - SKT doesn't care about the original status of XML file, it only detects it when there's a change. The only way this would work would be if `exec_on` make 2 changes to the XML. (First abort, second - termination condition is now True -> abort). However, we can simulate this by setting the amount of attempts to 0.

Thoughts?

veruu: That adjustment sounds reasonable.

However, since the `self.has_aborted` location was also changed (https://github.com/CKI-project/skt/compare/741f4985cd48c3347c5503660327ef6fa63dff12..22526059499e8d930bc85c06e1e6450fa44d9b85) I'd go with another set of testing to ensure this is ok before merging.

veruu:
@cki-bot test please!

veruu:
Behavior looks consistent with the previous runs so let's merge this

PR #516, opened 2020-04-17
use do_xml_replacements from cki-lib

wizardofoz123:
Some code was modified and moved to cki-lib to be reused. This makes use
of it, adjusts unit tests and removes some code.

### DIFF
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -15,6 +15,7 @@
 import copy
 import logging
 import os
+import pathlib
 import platform
 import re
 import subprocess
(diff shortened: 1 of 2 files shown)


wizardofoz123:
@cki-bot test

skt/runner.py
```diff
@@ -732,11 +698,11 @@ def run(self, url, max_aborted, release, wait=False,
         self.max_aborted = max_aborted
 
         try:
-            job_xml_tree = fromstring(self.__getxml(
-                {'KVER': release,
-                 'KPKG_URL': url,
-                 'ARCH': arch}
-            ))
+            lines = open(self.template).read().splitlines()
```
veruu, line 701: Please also close the file after using it (or use a context manager as the removed function did)

mh21:
```suggestion
            lines = pathlib.Path(self.template).read_text().splitlines()
```

(needs a pathlib import)

PR #517, opened 2020-05-06
Don't remove hostname settings in retries

veruu:
kpet-db contains a lot of hostname blacklisting to avoid running on
unsupported machines. We need to keep them for retries, otherwise the
recipes could end up running on those machines and just abort.

### DIFF
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -359,8 +359,10 @@ class TestRunner(unittest.TestCase):
 
         result = self.myrunner._BeakerRunner__recipe_set_to_job(xml_parsed)
 
-        # check that <hostname op="!=" value="hst1"/> was removed
-        self.assertEqual(result.findall('.//hostname'), [])
+        # check that <hostname op="!=" value="hst1"/> wasn't removed
+        self.assertEqual(len(result.findall('.//hostname')), 1)
+        self.assertEqual(tostring(result.find('.//hostname')).decode(),
+                         '<hostname op="!=" value="hst1" />')
 
     @mock.patch('skt.runner.BeakerRunner._BeakerRunner__jobsubmit')
     def test_run(self, mock_jobsubmit):
--- a/skt/runner.py
+++ b/skt/runner.py
@@ -377,10 +377,11 @@ class BeakerRunner:
 
         for recipe in tmp.findall('recipe'):
             hreq = recipe.find("hostRequires")
-            hostname = hreq.find('hostname')
-            if hostname is not None:
-                hreq.remove(hostname)
             if samehost:
+                hostname = hreq.find('hostname')
+                if hostname is not None:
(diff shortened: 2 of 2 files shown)


veruu:
@cki-bot please test

tests/test_runner.py
```diff
@@ -359,8 +359,8 @@ def test_recipe_set_to_job_whst(self):
 
         result = self.myrunner._BeakerRunner__recipe_set_to_job(xml_parsed)
 
-        # check that <hostname op="!=" value="hst1"/> was removed
-        self.assertEqual(result.findall('.//hostname'), [])
+        # check that <hostname op="!=" value="hst1"/> wasn't removed
+        self.assertEqual(len(result.findall('.//hostname')), 1)
```
emanchado, line 363: Why check only the length and not the actual content? Does `findall` return an annoying object that is hard to compare or something?

veruu: It's an element object that needs to be transformed to string in order to be easily compared. I can do that if you think it's useful

veruu:
@emanchado the bot is testing it. I'd be surprised if there wouldn't be a single resubmitted jobs. If that doesn't work, I'll create a pipeline definition PR that overrides skt location a specify a pipeline that crashes a lot.

veruu:
It looks like the only pipelines using the hostname overrides (not the "not hostname" ones, which are not affected) are upstream s390x pipelines. There's a LOT of recipes currently queued for them so I don't want to submit even more :(

However, the normal testing works well with this change. @rasibley do you think that's good enough to merge?

rasibley:
@veruu I think it should be ok, yeah we are down two s390x nodes at the moment :-/ Thanks for working on this!

tests/test_runner.py
emanchado: I know this is merged, but... I think it's ok as-is, I just wondered why it was originally a comparison between the lists, and then just their lengths. However, as a possible follow-up, there are two cases that come to mind that I wonder if they are worth testing:

1. Non-ASCII in the tests? I guess it won't be a problem for hostnames, but I wonder if we have any reason to think that non-ASCII characters could be a problem, in which case we should add a case for that.
1. If there is more than one result, does the order matter? Could the order become a problem? If so, test for it.

veruu: 1. By non-ascii, are you talking about the decodes? Those are normal for python3 as it defaults to bytes. If you mean actual unexpected unicode characters, I think we're fine here as we control the passed values.

2. No, order shouldn't matter.

emanchado: I'm not sure what I meant with the first point, haha. I guess if we're using well-tested libraries for all this, it shouldn't be a problem. The producer of that XML is also using a library, I assume? No `printf`-like things or string concatenating or anything, I hope... So just ignore my comment.

